PR #10, opened 2025-07-15
Use the rest api types from go driver

sfc-gh-mrolbiecki:


sf_core/Cargo.toml
```diff
@@ -22,4 +21,4 @@ uuid = { version = "1.0", features = ["v4"] }
 base64 = "0.22.1"
 arrow-ipc = "55.1.0"
 arrow = { version = "55.1.0", features = ["ffi"] }
-
+tempfile = "3.11.0"
```
sfc-gh-jszczerbinski: Why do we need it?

sf_core/src/api_server/database_driver_v1.rs
```diff
@@ -550,7 +550,23 @@ impl DatabaseDriverSyncHandler for DatabaseDriverV1 {
...
+                        StatusCode::UNKNOWN,
+                        None,
+                        None,
+                        None,
+                    )));
+                }
+
+                // TODO: Branch out to handle PUT / GET commands
+
+                let rowset_base64 = response.data.rowset_base64.unwrap();
```
sfc-gh-jszczerbinski: Don't just unwrap, return an error if it's missing

sf_core/src/rest/snowflake/mod.rs
```diff
@@ -169,26 +172,27 @@ pub async fn snowflake_login(
     let status = response.status();
     tracing::debug!(status = %status, "Received login response");
 
+    let response_text = response.text().await.map_err(|e| {
+        tracing::error!(error = %e, "Failed to get response text");
+        RestError::Internal(format!("Failed to get response text: {e}"))
```
sfc-gh-jszczerbinski: We will return "Failed to get response text: {e}" when request failed, and we won't get status code or information that the request failed as we did previously. I think this is suboptimal

sfc-gh-jszczerbinski (approved):

PR #13, opened 2025-07-21
Branch out on PUT / GET queries

sfc-gh-mrolbiecki:


sf_core/src/handle_manager.rs
```diff
@@ -65,7 +65,7 @@ impl<T> HandleManager<T> {
         let magic = handle_value.magic;
         match handle_value.value.as_ref() {
             Some(val) if magic == handle.magic => {
-                tracing::debug!(target: "handle_manager", "Handle retrieved successfully");
+                //tracing::debug!(target: "handle_manager", "Handle retrieved successfully");
```
sfc-gh-jszczerbinski: Why remove it? Maybe we should move all logs here to TRACE level?

sfc-gh-jszczerbinski (approved):

PR #17, opened 2025-07-28
[SNOW-2226829] File encryption for PUT

sfc-gh-mrolbiecki:
For this PR I implemented compression and encryption to match exactly the behavior of the Python Connector. Later I will explore more elegant approach.


sf_core/src/api_server/encryption.rs
```diff
@@ -0,0 +1,245 @@
+use crate::rest::error::RestError;
+use base64::{Engine, engine::general_purpose};
+use openssl::rand::rand_bytes;
+use openssl::symm::{Cipher, encrypt};
+use serde::Serialize;
+
+// Cryptographic constants
+const AES_256_KEY_SIZE: usize = 32; // 256 bits ÷ 8 = 32 bytes
```
sfc-gh-jszczerbinski: Comment is very informative when you read the definition, but I think it would more readable in place of use if we included unit in the name
```suggestion
const AES_256_KEY_SIZE_IN_BYTES: usize = 32; // 256 bits ÷ 8 = 32 bytes
```

sfc-gh-jszczerbinski: Same for other constants

sf_core/src/api_server/encryption.rs
```diff
@@ -0,0 +1,245 @@
...
+/// 4. Encrypt the file data using AES-CBC with PKCS#7 padding
+/// 5. Encrypt the data encryption key using the master key with PKCS#7 padding
+pub fn encrypt_file_data(
+    file_data: &[u8],
+    encryption_material: &EncryptionMaterial,
+) -> Result<EncryptionResult, RestError> {
+    // Step 1: Decode master key to determine key size and algorithms
+    let master_key = general_purpose::STANDARD
+        .decode(&encryption_material.query_stage_master_key)
+        .map_err(|e| RestError::Internal(format!("Failed to decode master key: {e}")))?;
```
sfc-gh-jszczerbinski: I think we shouldn't use RestError here. Instead maybe we should introduce encryption specific error type

sf_core/src/api_server/encryption.rs
```diff
@@ -0,0 +1,245 @@
...
+        key_size: key_size_bits.to_string(),
+    };
+    let mat_desc = serde_json::to_string(&material_desc).map_err(|e| {
+        RestError::Internal(format!("Failed to serialize material description: {e}"))
+    })?;
+
+    let metadata = EncryptedFileMetadata {
+        encryption_material: encryption_material.clone(),
+        encrypted_key: general_purpose::STANDARD.encode(&encrypted_file_key),
+        iv: general_purpose::STANDARD.encode(&iv),
```
sfc-gh-jszczerbinski: It's not very readable what general_purpose::STANDARD is. Maybe we should do sth like that
```rust
let base64_engine = base64::engine::general_purpose::STANDARD
...
base64_engine.encode(&iv)
```

sf_core/src/api_server/encryption.rs
```diff
@@ -0,0 +1,245 @@
...
+        mat_desc,
```
sfc-gh-jszczerbinski: Can this field be a struct and we move out the serialization to the function where we submit a request?

sf_core/src/api_server/file_transfer.rs
```diff
@@ -134,52 +197,56 @@ async fn upload_to_s3_simple(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use flate2::read::GzDecoder;
-    use std::collections::HashSet;
-    use std::io::{Read, Write};
-    use tempfile::NamedTempFile;
+
+    // Helper functions for hex conversion
+    fn hex_to_bytes(hex: &str) -> Result<Vec<u8>, std::num::ParseIntError> {
```
sfc-gh-jszczerbinski: Create utils.rs file and move the function there?

sf_core/src/api_server/file_transfer.rs
```diff
@@ -110,13 +152,34 @@ async fn upload_to_s3_simple(
...
+    let mut put_object_request = s3_client
         .put_object()
         .bucket(bucket)
         .key(&s3_key)
         .body(ByteStream::from(data))
-        .content_type("application/gzip")
+        .content_type("application/octet-stream");
+
+    // Add encryption metadata headers if available
+    if let Some(metadata) = encryption_metadata {
```
sfc-gh-jszczerbinski: Is the encryption optional?

sfc-gh-jszczerbinski (changes requested):

sf_core/src/api_server/file_transfer.rs
sfc-gh-mrolbiecki: No! Thanks for noticing

sf_core/src/rest/snowflake/query_response.rs
```diff
@@ -0,0 +1,372 @@
...
+    #[serde(rename = "queryId")]
+    query_id: String,
+    #[serde(rename = "smkId")]
+    smk_id: i64,
+}
+
+impl Data {
+    /// Extracts the fields necessary for file transfer, leaving `None` in their place.
+    /// This consumes the inner data of the fields but not the struct itself.
+    pub fn extract_file_transfer_data(&mut self) -> Result<file_manager::Data, RestError> {
```
sfc-gh-jszczerbinski: To we really need to mutate the object?

sfc-gh-jszczerbinski (approved):

PR #18, opened 2025-07-31
SNOW-2231412: Implement GET command support

sfc-gh-mrolbiecki:


sf_core/src/api_server/database_driver_v1.rs
```diff
@@ -573,23 +573,28 @@ impl DatabaseDriverSyncHandler for DatabaseDriverV1 {
...
-                                Error::from(RestError::Internal(format!(
-                                    "Failed to transfer file: {e}"
-                                )))
-                            })?;
-                        stmt.state = StatementState::Executed;
-                        return Ok(ExecuteResult::new(
-                            Box::new(ArrowArrayStreamPtr::new(Vec::new())),
-                            0,
-                        ));
+                        let file_upload_data = response.data.to_file_upload_data()?;
```
sfc-gh-jszczerbinski, line 576: Having this implementation here is fine for now, but I think it's best if this class was an adapter and most of the implementation lived separate classes / functions

sfc-gh-jszczerbinski (approved):

PR #23, opened 2025-08-04
SNOW-2239736: Implement support for wildcard characters in PUT

sfc-gh-mrolbiecki:


sf_core/src/file_manager/types.rs
```diff
@@ -16,6 +17,13 @@ pub struct UploadData {
     pub auto_compress: bool,
 }
 
+pub struct SingleUploadData {
+    pub src_location: String,
```
sfc-gh-jszczerbinski, line 21: Path instead of String?

sf_core/src/file_manager/mod.rs
```diff
@@ -118,3 +139,35 @@ fn validate_src_location_and_encryption_materials(
         encryption_materials.pop().unwrap(),
     ))
 }
+
+/// Expands file names using glob patterns, returning a list of valid file paths
+fn expand_file_names(pattern: &str) -> Result<Vec<String>, FileManagerError> {
+    let mut expanded_files = Vec::new();
+    let paths = glob::glob(pattern)?;
```
sfc-gh-jszczerbinski, line 146: It's ok to use it for now, but are you sure globs work **exactly** the same in Java, Python and Rust?

sfc-gh-jszczerbinski (approved):
Minor comments, but in general LGTM

PR #31, opened 2025-08-14
SNOW-2259555: Key pair authentication in ODBC

sfc-gh-jszczerbinski:


sf_core/src/auth.rs
```diff
@@ -0,0 +1,118 @@
...
+    exp: i64,
+}
+
+fn generate_jwt_token(
+    account: &str,
+    username: &str,
+    private_key: &str,
+    passphrase: Option<&str>,
+) -> Result<String, AuthError> {
+    use base64::{Engine as _, engine::general_purpose::STANDARD as BASE64};
```
sfc-gh-mrolbiecki, line 36: Inconsistent with what I've used earlier, but I think we can change my code cause BASE64 is shorter than BASE64_ENGINE

sfc-gh-mrolbiecki (approved):

PR #45, opened 2025-08-22
Snow 2272196 create poc reference tests for python wrapper universal driver

sfc-gh-fpawlowski:


.github/workflows/test-pep249.yml
```diff
@@ -0,0 +1,196 @@
...
+  build_rust_core:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y unixodbc-dev
```
sfc-gh-jszczerbinski: We don't need this, I think

pep249_dbapi/Makefile
```diff
@@ -0,0 +1,194 @@
...
+
+# Absolute paths to avoid relative path issues across different working directories
+DEFAULT_CORE_PATH_MACOS := $(shell cd .. && pwd)/target/debug/libsf_core.dylib
+DEFAULT_CORE_PATH_LINUX := $(shell cd .. && pwd)/target/debug/libsf_core.so
+DEFAULT_PARAMETER_PATH := $(shell cd .. && pwd)/parameters.json
+
+SHELL := bash
+.ONESHELL:  # Execute each recipe in a single shell (needed for multi-line commands)
+
+.PHONY: check-env-or-build check-env-or-fail setup build-wheel test test-local test-local-sequential test-local-tox test-local-tox-sequential test-integ-local-tox test-reference-local-tox test-all test-integ-reference compare compare-local compare-artifacts clean clean-reports run-with-setup run-with-uv run-with-tox
```
sfc-gh-jszczerbinski, line 18: Why are we using make if all our targets are PHONY? I would KISS nad use shell

pep249_dbapi/pyproject.toml
```diff
@@ -1,58 +1,63 @@
 [build-system]
```
sfc-gh-jszczerbinski, line 1: I like that you are updating this file

sfc-gh-jszczerbinski (approved):
In general LGTM, but I am not sure if we need make

pep249_dbapi/Makefile
sfc-gh-fpawlowski: It is quite common for Python stack to include Makefile - for common tasks automation. Some of them require certain setup (as it can be seen in this file) and it's easier to enclose it within a descriptive make recipe.
I get the argument, that we can do the same with shell files - but I am not sure if this way we wouldn't end up with bigger mess and less developer friendly work environment.

Instead of _make test_, we would probably call ./scripts/test.sh or ./scripts/testing/test_local_tox.sh. 
It is not horrible (tho I can imagine how big would directory with scripts grow and also potential namespace collisions issues), but it seems more verbose for those very common and quick tasks like running linting / tests / build.

There are some articles (sorry for the medium link) in favour of this approach - if You need more context: https://medium.com/@petrica.leuca/what-does-a-makefile-do-5e987d537235

.github/workflows/test-pep249.yml
sfc-gh-fpawlowski: Good catch! Leftover, removed 👍

pep249_dbapi/Makefile
sfc-gh-fpawlowski: Some alternative I can see is having a separate ./scripts directory and calling them from Makefile - if we want to achieve better encapsulations or if our Makefile itself grows to some maintenance-painful size. Then Makefile would be just a nice interface / wrapper for those scripts.

PR #49, opened 2025-08-27
NO-SNOW: Add uv.lock file

sfc-gh-fpawlowski:


sfc-gh-mrolbiecki (approved):
Python formatting changes should be in another PR but otherwise looks good

PR #61, opened 2025-09-12
NO-SNOW: fix pep249 put/get tests to work with Python Connector

sfc-gh-mrolbiecki:
Tests should now work with Python Connector, but will fail for universal driver. This is for a reason and I will fix this the universal driver in the next PR.

### DIFF
--- a/pep249_dbapi/tests/integ/test_put_get_simple.py
+++ b/pep249_dbapi/tests/integ/test_put_get_simple.py
@@ -22,6 +22,7 @@ from .utils_put_get import (
 )
 
 from ..connector_types import ConnectorType
+from ..utils import NEW_DRIVER_ONLY, OLD_DRIVER_ONLY, select_by_driver
 
 
 def test_put_select(cursor):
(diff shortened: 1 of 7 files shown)


pep249_dbapi/tests/integ/test_put_get_simple.py
```diff
@@ -116,10 +117,10 @@ def test_put_get_rowset(cursor, connector_type):
         assert row[PUT_ROW_TARGET_IDX] == "test_put_get_rowset.csv.gz"
         assert row[PUT_ROW_SOURCE_SIZE_IDX] == 6
 
-        # BREAKING CHANGE: changing the compression behavior reduces the size of a compressed file
-        if connector_type == ConnectorType.REFERENCE:
+        if OLD_DRIVER_ONLY("BC#1: header of a gzip-compressed file does not contain filename"):
```
sfc-gh-akolodziejczyk: please change to the current format with only numbers: OLD_DRIVER_ONLY("BC#1") in all places

sfc-gh-akolodziejczyk (approved):

PR #67, opened 2025-09-13
Added X.509 certificate and CRL parsing capabilities to support TLS certificate validation.


sfc-gh-snoonan:
- Added new dependencies for X.509 certificate handling and TLS functionality
- Created a new `x509_utils.rs` module with functions to extract Subject Key Identifier (SKID) from certificates and Authority Key Identifier (AKID) from CRLs
- Added a new binary target `tls_client` for testing TLS functionality
- Updated the CRL validator to prepare for proper certificate revocation checking
- Modified dependency configurations to better support TLS and certificate operations

PR #66, opened 2025-09-12
Initial CRL stubs with tests and fail open only

sfc-gh-snoonan:
- Created a new `crl` module with the following components:
    - `CrlConfig`: Configuration struct with an `enabled` flag
    - `CrlValidator`: Core validator that checks certificate revocation status
    - `RevocationOutcome` enum with three states: `NotRevoked`, `Revoked`, and `Unknown`
    - Error handling via `CrlError` enum
- The current implementation is a placeholder that returns `Unknown` when enabled and an error when disabled
- Added X.509 certificate utilities in `tls/x509_utils.rs` for extracting Subject Key Identifier (SKID) and Authority Key Identifier (AKID)
- Added necessary dependencies for X.509 certificate and CRL parsing
- Included a TLS client test that verifies connection setup with TLS configuration
- Fixes for ODBC tests; we can split these out if needed.

PR #65, opened 2025-09-12
Add support for initial tls plumbing

sfc-gh-snoonan:
- Implements a Rust TLS client with custom certificate store and verification options
- Adds bindings for ODBC and Python drivers to configure TLS settings
- Creates a standalone TLS client tool for testing and debugging connections

The implementation allows clients to:
- Specify a custom root certificate store path
- Control hostname verification
- Control certificate verification
- Configure these settings through connection parameters

PR #68, opened 2025-09-14
initial crl cache implementation

sfc-gh-snoonan:
This PR adds Certificate Revocation List (CRL) validation capabilities

- Implements a complete CRL validation system with configurable behavior modes (Disabled, Enabled, Advisory)
- Adds a `CrlCache` for efficient CRL retrieval and storage with both memory and disk caching
- Provides certificate parsing utilities to extract CRL distribution points and check revocation status
- Integrates with the TLS stack via a custom `CrlServerCertVerifier` implementation
- Exposes configuration options through connection string parameters:
    - `CRL_ENABLED` - Enable/disable CRL checking
    - `CRL_MODE` - Set validation mode (DISABLED, ENABLED, ADVISORY)

### DIFF
--- a/sf_core/src/crl/cache.rs
+++ b/sf_core/src/crl/cache.rs
@@ -1,469 +1,242 @@
-use crate::crl::validator::RevocationOutcome;
+use crate::crl::config::CrlConfig;
+use crate::crl::error::CrlError;
+use chrono::{DateTime, Utc};
 use lru::LruCache;
-use std::collections::{HashMap, hash_map::DefaultHasher};
-use std::hash::{Hash, Hasher};
+use once_cell::sync::OnceCell;
+use opentelemetry::metrics::{Counter, Histogram, Meter};
+use opentelemetry::{KeyValue, global};
+use sha2::{Digest, Sha256};
+use std::collections::HashMap;
 use std::num::NonZeroUsize;
 use std::sync::{Arc, Mutex};
-use std::time::{Duration, Instant};
-use x509_parser::prelude::FromDer;
 
-/// Configuration for CRL cache behavior.
 #[derive(Debug, Clone)]
-pub struct CrlCacheConfig {
-    /// Enable CRL cache logic. When disabled, operations return Unknown (fail-open).
-    pub enabled: bool,
-    /// Maximum number of outcome entries to keep in-memory.
-    pub outcome_capacity: usize,
-    /// Maximum number of raw CRL blobs to keep in-memory.
-    pub raw_capacity: usize,
-    /// Fallback TTL to apply to CRLs when no validity info is available.
-    pub ttl_seconds: u64,
-    /// Base backoff in milliseconds after a failed fetch.
-    pub backoff_base_ms: u64,
-    /// Maximum backoff cap in milliseconds.
-    pub backoff_cap_ms: u64,
-    /// Maximum jitter in milliseconds added to backoff.
-    pub backoff_jitter_ms: u64,
+pub struct CachedCrl {
+    pub crl: Vec<u8>,
+    pub download_time: DateTime<Utc>,
+    pub url: String,
 }
 
-impl Default for CrlCacheConfig {
-    fn default() -> Self {
+#[derive(Debug)]
+pub struct CrlCache {
+    config: CrlConfig,
+    memory_cache: Option<Arc<Mutex<LruCache<String, CachedCrl>>>>,
+    url_locks: Arc<Mutex<HashMap<String, Arc<Mutex<()>>>>>,
+    backoff: Arc<Mutex<HashMap<String, (u32, std::time::Instant)>>>,
+}
+
+// Outcome memoization removed for now; can be reintroduced if needed
+
+#[derive(Debug, Clone)]
+struct CrlMetrics {
+    get_total: Counter<u64>,
+    get_ms: Histogram<u64>,
+    fetch_total: Counter<u64>,
+    fetch_error_total: Counter<u64>,
+    fetch_ms: Histogram<u64>,
+}
+
+impl CrlMetrics {
+    fn init(meter: &Meter) -> Self {
         Self {
-            enabled: false,
-            outcome_capacity: 10_000,
-            raw_capacity: 1_000,
-            ttl_seconds: 60 * 10, // 10 minutes
-            backoff_base_ms: 100,
-            backoff_cap_ms: 5_000,
-            backoff_jitter_ms: 100,
+            get_total: meter.u64_counter("crl_get_total").build(),
+            get_ms: meter.u64_histogram("crl_get_ms").build(),
+            fetch_total: meter.u64_counter("crl_fetch_total").build(),
+            fetch_error_total: meter.u64_counter("crl_fetch_error_total").build(),
+            fetch_ms: meter.u64_histogram("crl_fetch_ms").build(),
         }
     }
 }
 
-/// Abstraction for fetching CRLs from URLs. Synchronous for simplicity.
-pub trait CrlFetcher: Send + Sync + 'static {
-    fn fetch(&self, url: &str) -> Result<Vec<u8>, String>;
+fn metrics() -> &'static CrlMetrics {
+    static METRICS: OnceCell<CrlMetrics> = OnceCell::new();
+    METRICS.get_or_init(|| {
+        let meter = global::meter("sf_core.crl");
+        CrlMetrics::init(&meter)
+    })
 }
 
-struct CachedCrl {
-    bytes: Vec<u8>,
-    expires_at: Instant,
-}
-
-fn default_expires(ttl_seconds: u64) -> Instant {
-    Instant::now() + Duration::from_secs(ttl_seconds)
-}
-
-fn hash_bytes(bytes: &[u8]) -> u64 {
-    let mut hasher = DefaultHasher::new();
-    bytes.hash(&mut hasher);
-    hasher.finish()
-}
-
-/// In-memory CRL cache with raw CRL and outcome memoization.
-pub struct CrlCache<F: CrlFetcher> {
-    config: CrlCacheConfig,
-    raw_crls: Arc<Mutex<LruCache<String, CachedCrl>>>,
-    outcomes: Arc<Mutex<LruCache<(u64, String), RevocationOutcome>>>, // (serial_hash, url)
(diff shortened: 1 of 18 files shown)


sfc-gh-snoonan:
* **#68** <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/68?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#69](https://github.com/snowflakedb/universal-driver/pull/69) <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/69?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#71](https://github.com/snowflakedb/universal-driver/pull/71) <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/71?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/68?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#66** <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/66?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#67](https://github.com/snowflakedb/universal-driver/pull/67) <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/67?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#65** <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/65?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sf_core/src/crl/cache.rs
```diff
@@ -0,0 +1,328 @@
+use crate::crl::config::CrlConfig;
+use crate::crl::error::CrlError;
+use chrono::{DateTime, Utc};
+use lru::LruCache;
+use once_cell::sync::OnceCell;
+use opentelemetry::metrics::{Counter, Histogram, Meter};
```
sfc-gh-jszczerbinski, line 5: Opentelemetry was added to the project in order to test our debugging options - emmiting OTEL traces and using jaeger to debug. We can keep CrlMetrics  - but FYI we don't have long term strategy around telemetry yet (especially metrics) - CrlMetrics might be dead code soon

sf_core/src/tls/crl_verifier.rs
```diff
@@ -0,0 +1,117 @@
...
+        self.webpki_verifier
+            .verify_tls13_signature(message, cert, dss)
+    }
+
+    fn supported_verify_schemes(&self) -> Vec<SignatureScheme> {
+        self.webpki_verifier.supported_verify_schemes()
+    }
+}
+
+fn shared_runtime() -> &'static tokio::runtime::Runtime {
```
sfc-gh-jszczerbinski, line 109: Let's use lazy static here instead

sf_core/src/crl/validator.rs
```diff
@@ -1,54 +1,72 @@
...
+            let issuer_der = chain.get(idx + 1).map(|v| v.as_slice());
+            match self.cache.check_revocation(cert_der, issuer_der).await {
+                Ok(outcome) => {
+                    use crate::tls::revocation::RevocationOutcome;
+                    if let RevocationOutcome::Revoked { .. } = outcome {
+                        return Ok(false);
+                    }
+                }
+                Err(_) => {
+                    // Non-deterministic -> treat as not determined; continue to next cert
```
sfc-gh-jszczerbinski: Shouldn't we return error here? Or at least remember that we encountered error and return it later unless we encounter revoked certificate:
```
        chainResult = CHAIN_UNREVOKED

 for cert in certChain except root {  // (1), check: all CRL URLs should pass 

            if (isShortLived(cert)) {
                continue; // cert loop
            }
            
            if (cert.crlUrls is empty list or has blank values) {
                // Internal flag to cover scenarios where CRL URL is not present.
                // By design we return CHAIN_UNREVOKED. This is only valid for a
                // leaf (Subscriber) certificate, but we do not implement that check.
                if (AllowCertificatesWithoutCrlUrl) {
                    log.warn("Certificate has missing CRL Distribution Point URLs")
                    continue; // cert loop
                }                
 chainResult = CHAIN_ERROR;
                continue; // cert loop
            }
            certStatus = validateCert(cert)

            if (certStatus == CERT_REVOKED) {
                chainResult = CHAIN_REVOKED;
                break; // cert loop
            }

            if (certStatus == CERT_ERROR) {
                chainResult = CHAIN_ERROR;
                continue; // cert loop
            }
        } // cer
```

sf_core/src/crl/cache.rs
```diff
@@ -0,0 +1,328 @@
...
+        Ok(Self {
+            config,
+            memory_cache,
+            url_locks: Arc::new(Mutex::new(HashMap::new())),
+            backoff: Arc::new(Mutex::new(HashMap::new())),
+            http_client,
+        })
+    }
+
+    pub fn global(config: CrlConfig, memory_capacity: usize) -> &'static Arc<CrlCache> {
```
sfc-gh-jszczerbinski: I don't think the api makes sense here. We will set capacity once when we create the cache for the first time. Can we make the cache resizable and adjust the size when we change configuration? It's propably overkill, but this feels wrong.

sf_core/src/crl/cache.rs
```diff
@@ -0,0 +1,328 @@
...
+        let guard = lock.lock().map_err(|e| CrlError::MutexPoisoned {
+            message: format!("lock poisoned: {e}"),
+            location: snafu::Location::new(file!(), line!(), 0),
+        })?;
+        if let Some(mem) = self.get_cached(url)? {
+            return Ok(mem.crl);
+        }
+
+        // Fetch and optionally persist
+        drop(guard);
```
sfc-gh-jszczerbinski: Should we really drop the guard here? I think if we would create multiple connections to the same host this would result in duplicate downloads of the CRLs - due to race condition.
If we drop the guard after putting downloaded CRL to cache we can avoid duplicate downloads and probably some memory issues

sf_core/src/crl/cache.rs
```diff
@@ -0,0 +1,328 @@
...
+        self.record_backoff_success(url);
+        let ms = start.elapsed().as_millis() as u64;
+        metrics().fetch_ms.record(ms, &[]);
+        metrics().fetch_total.add(1, &[]);
+        Ok(bytes.to_vec())
+    }
+
+    async fn maybe_sleep_backoff(&self, url: &str) -> Result<(), CrlError> {
+        let (failures, last) = {
+            let guard = self.backoff.lock().map_err(|e| CrlError::MutexPoisoned {
```
sfc-gh-jszczerbinski: Please use MutexPoisonedSnafu here instead. If it's not visible in this scope use `#[snafu(visibility(pub))]` on Error enum definitions to make Snafu public
```
MutexPoisonedSnafu { message: format!("backoff map poisoned: {e}") }.build()
```

sf_core/src/crl/cache.rs
```diff
@@ -0,0 +1,328 @@
...
+                url: url.to_string(),
+                source: e,
+                location: snafu::Location::new(file!(), line!(), 0),
+            })?;
+        let resp = resp.error_for_status().map_err(|e| CrlError::CrlDownload {
+            url: url.to_string(),
+            source: e,
+            location: snafu::Location::new(file!(), line!(), 0),
+        })?;
+        let bytes = resp.bytes().await.map_err(|e| CrlError::CrlDownload {
```
sfc-gh-jszczerbinski: Use Snafu here (this is a similar issue to the one below), but here you can also use context (to fill in source error) instead of map_err:
```
.context(CrlDownloadSnafu { url: url.to_string() })
```

sfc-gh-jszczerbinski (changes requested):
In general looks good, but I have some reservations. I left only two comments about snafu usage, but please improve the usage in the whole PR

sf_core/src/crl/cache.rs
sfc-gh-snoonan: Ack. I think it would be pretty nice if we kept otel.

sf_core/src/crl/cache.rs
sfc-gh-snoonan: Yeah, you're right. I was trying to limit this because we only connect to ~dozens of endpoints in a given client, but then copilot yelled at me about magic numbers. I'm just going to leave it unbounded rather than being LRU with capacity now. If there's a reasonable upper bound we want to keep or if we want it there just in case, I can add it back?

sfc-gh-snoonan: The outcome cache is gone for now; it gets added later in the stack. Right now we just fetch, write to disk, put the CRL in memory, and now that's basically unbounded. We don't ever read the CRLs from disk in this PR.

I think this is fine for the next few days; we don't default to enabling CRLs. If you'd rather have a larger portion of the implementation in this PR, I can do that.

sfc-gh-jszczerbinski (approved):
LGTM

sf_core/src/crl/cache.rs
```diff
@@ -0,0 +1,334 @@
...
+    pub async fn check_revocation(
+        &self,
+        cert_der: &[u8],
+        issuer_der: Option<&[u8]>,
+    ) -> Result<crate::tls::revocation::RevocationOutcome, crate::tls::revocation::RevocationError>
+    {
+        use crate::tls::revocation::{RevocationError, RevocationOutcome};
+        // Extract CRL URLs
+        let crl_urls = crate::crl::certificate_parser::extract_crl_distribution_points(cert_der)
+            .map_err(|e| RevocationError::Crl {
```
sfc-gh-jszczerbinski: Still snafuing wrong :(

sfc-gh-jszczerbinski (changes requested):
Still snafu usage is not there

sf_core/src/crl/cache.rs
sfc-gh-jszczerbinski: .context(CrlSnafu{});

sfc-gh-jszczerbinski: We need an error type that wraps error returned by: extract_crl_distribution_points

sfc-gh-jszczerbinski (approved):
LGTM

PR #71, opened 2025-09-15
TLS CRL validation plumbing: parsing, caching, verifier integration, and safer defaults

sfc-gh-snoonan:
Introduces CRL-based certificate revocation checking into the TLS stack with best-effort signature verification, in-memory and optional disk caching, and a custom rustls server cert verifier. Adds robust error types and consistent logging.

PR #72, opened 2025-09-15
Improve and Simplify CRL Implementation

sfc-gh-snoonan:
Removed the openssl fallback for CRL signature verification to rely exclusively on aws-lc-rs. Refactored the CRL cache, validator, and parser functions by breaking down large methods into smaller, more focused helpers. CRL refetch logic is now only triggers on expired CRLs, improving efficiency.


sfc-gh-snoonan:
* **#73** <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/73?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#72** <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/72?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/72?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#71** <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/71?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#68** <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/68?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#69](https://github.com/snowflakedb/universal-driver/pull/69) <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/69?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#66** <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/66?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#67](https://github.com/snowflakedb/universal-driver/pull/67) <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/67?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#65** <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/65?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sf_core/src/c_api.rs
```diff
@@ -130,26 +130,33 @@ pub unsafe extern "C" fn sf_core_api_call_proto(
     response: *mut *const u8,
     response_len: *mut usize,
 ) -> usize {
-    unsafe {
+    // Prevent unwinding across the FFI boundary. Any panic will be converted to a transport error.
+    let result = std::panic::catch_unwind(|| unsafe {
```
sfc-gh-jszczerbinski, line 134: Great addition :)

tests/definitions/tls/crl_enabled.feature
```diff
@@ -0,0 +1,9 @@
+@odbc_e2e @python_e2e
```
sfc-gh-akolodziejczyk:
```suggestion
@core @odbc @python
Feature: CRL enabled

  @core_e2e @odbc_e2e @python_e2e
  Scenario: should connect and execute simple query with CRL enabled
    Given Snowflake client is logged in
    When Query "SELECT 1" is executed with CRL enabled
    Then The query is successful
```

tests/definitions/tls/handshake.feature
```diff
@@ -0,0 +1,20 @@
+Feature: TLS handshake
```
sfc-gh-akolodziejczyk, line 2: Annotations are missing for this test, please add as described in: `tests/definitions/README.md`.
I.e. @core at the feature level, and @core_e2e at the test levels

tests/definitions/tls/handshake.feature
```diff
@@ -0,0 +1,20 @@
...
+  As a client developer
+  I want to verify TLS handshakes with default and custom roots
+  So that connections can be established under different trust settings
+
+  Background:
+    Given an HTTPS server URL from env E2E_TLS_SERVER or default "https://example.com"
+
+  Scenario: Handshake with default roots
```
sfc-gh-akolodziejczyk: Please use the test definition format starting with should to be consistent with other tests

tests/definitions/tls/handshake.feature
```diff
@@ -0,0 +1,20 @@
...
+    Given a TLS client configured with default roots
+    When I send a GET request to the server URL
+    Then the request attempt should complete (success or error acceptable in CI)
+
+  Scenario: Handshake with custom PEM roots
+    Given E2E_TLS_ROOTS_PEM is set to a PEM bundle path
+    And a TLS client configured with that custom root store
+    When I send a GET request to the server URL
+    Then the request attempt should complete
```
sfc-gh-akolodziejczyk: should it also accept error in CI ? if not maybe we can change to `Then The request is successful` ?

sf_core/tests/e2e/tls/handshake.rs
```diff
@@ -0,0 +1,34 @@
...
+    // Then the request attempt should complete (success or error acceptable in CI)
+    // Only assert we can attempt; don't fail CI when network blocked
+    assert!(resp.is_ok() || resp.is_err());
+}
+
+#[tokio::test]
+async fn handshake_with_custom_roots_env() {
+    // Scenario: Handshake with custom PEM roots
+    if let Ok(pem_path) = std::env::var("E2E_TLS_ROOTS_PEM") {
+        // Given E2E_TLS_ROOTS_PEM is set to a PEM bundle path
```
sfc-gh-akolodziejczyk: Given step should be above line 20

sf_core/tests/e2e/tls/handshake.rs
```diff
@@ -0,0 +1,34 @@
+use sf_core::tls::config::TlsConfig;
+use sf_core::tls::create_tls_client_with_config;
+
+#[tokio::test]
+async fn handshake_with_default_roots() {
+    // Scenario: Handshake with default roots
```
sfc-gh-akolodziejczyk: I think we don't need those comments with Scenario name, please remove in all tests

sf_core/tests/e2e/tls/handshake.rs
```diff
@@ -0,0 +1,34 @@
...
+    // Given a TLS client configured with default roots
+    let server_url = std::env::var("E2E_TLS_SERVER").unwrap_or("https://example.com".to_string());
+    // When I send a GET request to the server URL
```
sfc-gh-akolodziejczyk: Please add a space before each Gherkin comment for better readability

sf_core/tests/e2e/tls/handshake.rs
```diff
@@ -0,0 +1,34 @@
...
+        let cfg = TlsConfig {
+            custom_root_store_path: Some(pem_path.into()),
+            ..Default::default()
+        };
+        let client = create_tls_client_with_config(cfg).expect("client");
+        let server_url =
+            std::env::var("E2E_TLS_SERVER").unwrap_or("https://example.com".to_string());
+        // When I send a GET request to the server URL
+        let _ = client.get(server_url).send().await;
+        // Then the request attempt should complete
```
sfc-gh-akolodziejczyk: missing assertions

pep249_dbapi/tests/e2e/tls/test_crl_enabled.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+
+
+@pytest.mark.skip_reference(reason="CRL e2e applies to universal driver")
+def test_connect_and_select_with_crl_enabled(connection_factory):
```
sfc-gh-akolodziejczyk: Relevant actions should be after the Gherkin step descriptions, i.e:
```
    with connection_factory(crl_check_mode="ENABLED") as conn:
        # Given Snowflake client is logged in
        cur = conn.cursor()

        # When Query "SELECT 1" is executed
        cur.execute("SELECT 1")
 
        # Then the request attempt should complete
        row = cur.fetchone()
        assert row is not None
        assert row[0] in (1, "1")

```

tests/definitions/tls/handshake.feature
```diff
@@ -0,0 +1,20 @@
...
+  As a client developer
+  I want to verify TLS handshakes with default and custom roots
+  So that connections can be established under different trust settings
+
+  Background:
+    Given an HTTPS server URL from env E2E_TLS_SERVER or default "https://example.com"
+
+  Scenario: Handshake with default roots
+    Given a TLS client configured with default roots
+    When I send a GET request to the server URL
```
sfc-gh-akolodziejczyk: let's use passive voice, i.e: `WHEN GET request is sent to the server URL` to have consistency in all tests

sfc-gh-akolodziejczyk (changes requested):
Some changes are needed for proper e2e tests format

tests/definitions/tls/crl_enabled.feature
```diff
@@ -0,0 +1,8 @@
+@core @odbc_e2e @python_e2e
```
sfc-gh-akolodziejczyk: On the feature level, there should be: @core @odbc @python
And on the test level: @core_e2e @odbc_e2e @python_e2e
I can see it's not clearly described in the Readme. I will improve it next week

sfc-gh-akolodziejczyk (approved):
LGTM! Just one comment in the feature file

tests/definitions/tls/crl_enabled.feature
sfc-gh-snoonan: Thanks! Fixed.

sf_core/src/crl/certificate_parser.rs
```diff
@@ -9,57 +10,43 @@ pub fn extract_crl_distribution_points(cert_der: &[u8]) -> Result<Vec<String>, C
...
+        .filter_map(|name| match name {
+            x509_parser::extensions::DistributionPointName::FullName(names) => Some(names.iter()),
+            _ => None,
+        })
+        .flatten()
+        .filter_map(|general_name| match general_name {
+            GeneralName::URI(uri) => Some(uri.to_string()),
+            _ => None,
+        })
+        .filter(|url| url.starts_with("http://") || url.starts_with("https://"))
```
sfc-gh-jkasten, line 32: no action required: HTTPS isn't actually allowed for CAs. (CRLs are signed and you create weird circular dependencies if validating your HTTPS connection requires additional HTTPS connections).

https://github.com/cabforum/servercert/blob/main/docs/BR.md#:~:text=A%20fullName%20MUST%20contain%20at%20least%20one%20GeneralName%3B%20it%20MAY%20contain%20more%20than%20one.%20All%20GeneralNames%20MUST%20be%20of%20type%20uniformResourceIdentifier%2C%20and%20the%20scheme%20of%20each%20MUST%20be%20%22http%22.%20The%20first%20GeneralName%20must%20contain%20the%20HTTP%20URL%20of%20the%20Issuing%20CA%27s%20CRL%20service%20for%20this%20certificate.

Occassionally, a smaller CA stills messes this up, but they are called out for it pretty quickly (CT) and have to perform a postmortem. Since linting is now a requirement for all certificates, I imagine this will drop even more.

sf_core/src/crl/certificate_parser.rs
```diff
@@ -76,9 +63,9 @@ pub fn is_short_lived_certificate(cert_der: &[u8]) -> Result<bool, CrlError> {
         })?;
 
     let validity_period_seconds = (not_after_dt - not_before_dt).num_seconds();
```
sfc-gh-jkasten: nit / outside the scope of this PR - notAfter is inclusive so states that the certificates are one second shorter than they actually are.
https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.5

This falls on the safe side of the check down below, so it isn't a big deal. I mention it mostly to prevent copying of the code and creating errors for the future.

(The web PKI generally tries to stay away from the edges of the requirements so that they don't have to worry about it. Avoid issues with libraries trying to handle leap seconds etc)

sf_core/src/crl/certificate_parser.rs
```diff
@@ -88,13 +75,34 @@ pub fn is_short_lived_certificate(cert_der: &[u8]) -> Result<bool, CrlError> {
     Ok(is_short_lived)
 }
 
+/// Use CA/B BR short-lived threshold: 10 days until 2026-03-15, then 7 days
+pub fn is_short_lived_certificate(cert_der: &[u8]) -> Result<bool, CrlError> {
+    let now = chrono::Utc::now();
+    let cutoff = chrono::Utc.with_ymd_and_hms(2026, 3, 15, 0, 0, 0).unwrap();
+    let threshold_days = if now < cutoff { 10 } else { 7 };
```
sfc-gh-jkasten: I think this should use the date of the notBefore rather than "now"?

"For Certificates issued on or after 15 March 2026, " (https://github.com/cabforum/servercert/blob/main/docs/BR.md#161-definitions)

FWIW, notBefore is also not a perfect substitute for "issued on" as CAs backdate certificates. Browsers typically treat the issued at time as the timestamp of the earliest issued SCT for the certificate. (Way too complicated for our usecase). I would honestly prefer to just give an extra day of buffer, but this is not something that we have consistently done through all of the drivers.

The nice thing is that we only really interact with the large CAs and they won't be playing close to the edges.

sf_core/src/crl/certificate_parser.rs
```diff
@@ -88,13 +75,34 @@ pub fn is_short_lived_certificate(cert_der: &[u8]) -> Result<bool, CrlError> {
...
+/// Determine if a certificate is a CA certificate (BasicConstraints CA=true)
+pub fn is_ca_certificate(cert_der: &[u8]) -> Result<bool, CrlError> {
+    let (_, cert) =
+        x509_parser::certificate::X509Certificate::from_der(cert_der).context(CrlParsingSnafu)?;
+    for ext in cert.extensions() {
+        if let ParsedExtension::BasicConstraints(bc) = ext.parsed_extension() {
+            return Ok(bc.ca);
+        }
+    }
+    // If BasicConstraints missing, treat as end-entity by default per common practice
```
sfc-gh-jkasten:
```suggestion
    // If BasicConstraints missing, treat as end-entity by RFC 5280 Section 4.2.1.9
```
This is defined.
https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.9

sf_core/src/crl/cache.rs
```diff
@@ -70,34 +138,107 @@ impl CrlCache {
...
+                    // Not revoked in this CRL, continue
+                    any_verified = true;
+                    let full_coverage = match crate::tls::x509_utils::extract_crl_idp_scope(&bytes)
+                    {
+                        Ok(Some(scope)) => !scope.has_only_some_reasons && !scope.only_attribute,
+                        _ => true,
+                    };
+                    if full_coverage {
+                        any_full_coverage = true;
+                    }
```
sfc-gh-jkasten, line 355: Awesome! Very impressed you actually implemented the reason codes of Section 6.3 of 5280 :)

sf_core/src/tls/x509_utils.rs
```diff
@@ -125,6 +167,28 @@ pub fn verify_crl_signature_best_effort(
                 return InvalidCrlSignatureSnafu {}.fail();
             }
         }
+
+        // Enforce basic IssuingDistributionPoint scope if present
+        if let Ok((_, parsed_crl)) =
+            x509_parser::revocation_list::CertificateRevocationList::from_der(crl_der)
+        {
+            for ext in parsed_crl.tbs_cert_list.extensions() {
+                if let ParsedExtension::IssuingDistributionPoint(idp) = ext.parsed_extension() {
```
sfc-gh-jkasten: Sorry, I am sure I just missed it, but where are we checking IssuingDistributionPoint.distributionPoint matches the CRL distribution point?

sf_core/src/crl/cache.rs
```diff
@@ -170,81 +317,76 @@ impl CrlCache {
...
 
+    fn should_replace_cached_crl(&self, prev: &CachedCrl, new: &CachedCrl) -> bool {
+        if let (Some(prev_num), Some(new_num)) = (prev.crl_number, new.crl_number) {
+            return new_num > prev_num;
+        }
+        if let Ok(Some(prev_dt)) = crate::tls::x509_utils::extract_crl_next_update(&prev.crl)
+            && let Ok(Some(new_dt)) = crate::tls::x509_utils::extract_crl_next_update(&new.crl)
+        {
+            return new_dt > prev_dt;
+        }
```
sfc-gh-jkasten: nit/optional: A CA that shortens the lifetime of their CRL might cause you not to update even when the CRL is "newer". 

I think the thisUpdate would be a better gauge, but all of the CAs are required to use CRL Numbers anyway (right now at least), so it is a very very minor point
https://github.com/cabforum/servercert/blob/main/docs/BR.md#722-crl-and-crl-entry-extensions

sfc-gh-jkasten (commented):
Looks great, Sean!
I wanted to make sure that I found the IssuingDistributionPoint.distributionPoint comparison before approving. I apologize if I just missed it.

sf_core/src/crl/certificate_parser.rs
sfc-gh-snoonan: Yeah, I know the logic. The RFC allows it though (arbitrary protocols, but I don't think we'll want to support ldap :-)) and it wouldn't surprise me if someone proposes that we can easily break the cycle using short lived certs so everything is HTTPS at some point...

sf_core/src/crl/certificate_parser.rs
sfc-gh-snoonan: Fixed anyway.

sf_core/src/crl/certificate_parser.rs
sfc-gh-snoonan: Fixed.

sf_core/src/tls/x509_utils.rs
sfc-gh-snoonan: Nope, I just missed implementing it. PTAL.

sf_core/src/crl/cache.rs
```diff
@@ -70,34 +160,117 @@ impl CrlCache {
...
+                if crate::tls::x509_utils::verify_crl_signature(crl_bytes, Some(cand)).is_ok() {
+                    verified = true;
+                    break;
+                }
+            }
+        }
+
+        if !verified {
+            return Ok(None);
+        }
```
sfc-gh-jkasten, line 405: I would assume that if we can't verify the CRL we should return a different answer than if we have confirmed the certificate is not revoked? (Might have missed something)

sf_core/src/crl/cache.rs
```diff
@@ -277,12 +453,48 @@ impl CrlCache {
...
+            metrics()
+                .get_total
+                .add(1, &[KeyValue::new("source", "disk")]);
+            return Ok(disk);
+        }
+
+        let fetched = self.fetch_from_network_and_cache(url).await?;
         let ms = start.elapsed().as_millis() as u64;
         metrics()
             .get_ms
```
sfc-gh-jkasten, line 674: <p class='graphite__hidden'><i>[Re: lines +468 to +506]</i></p>
Non-blocking: I know at some point we discussed trying to fetch the new CRL before the old CRL entirely expired [[design doc thread](https://docs.google.com/document/d/1Dy4eb8dDV75XIUKBvOMfeZjnxYca6ICKsCtgWwfU9Fs/edit?disco=AAABk6j3w_w)]. I don't see that here yet, but I could have also missed it. Non-blocking regardless.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/72?utm_source=unchanged-line-comment">Graphite</a>.</p>

sf_core/src/tls/x509_utils.rs
```diff
@@ -328,10 +327,243 @@ pub fn extract_issuer_skid(issuer_der: &[u8]) -> Option<Vec<u8>> {
...
+                return Some(kid.0.to_vec());
+            }
+        }
+    }
+    None
+}
+
+// Build candidate chains from an end-entity and a list of intermediates.
+// Each chain is a vector of cert DER bytes from EE up to last found parent.
+pub fn build_candidate_chains(end_entity: &[u8], intermediates: &[Vec<u8>]) -> Vec<Vec<Vec<u8>>> {
```
sfc-gh-jkasten: I would expect to see roots/trust anchors be included in these chains so that the highest level intermediate can be validated? (Maybe this is being handled in a different way?)

sfc-gh-jkasten (approved):
Thanks, Sean! I apologize for the slow review. This is my first time looking at this code, so I am still trying to get my bearings. I think I gave it a reasonably close review.

sf_core/src/tls/x509_utils.rs
sfc-gh-snoonan: Good catch! I think there was a bug here where we might choose a different path than the one that was used for the handshake, which might not be anchored properly.

Rust doesn't give us a nice interface that provides chains anchored all the way to the root. The expectation seems to be that all you need is subject/SPKI, which is true, but it makes building chains all the way to the root somewhat annoying type-wise. (https://docs.rs/webpki/latest/webpki/struct.TrustAnchor.html)

So now we build the candidate chains, then check them against the trust store to ensure they have a valid anchor. For each chain with a valid anchor, we do CRL checks.

For the CRLs, we resolve the CRL issuer to that same trust anchor by matching issuer Name DER to anchor.subject (resolve_anchor_issuer_key), then verify the CRL signature with anchor.spki.

I *think* this is correct and RFC compliant. It does, however, make me a bit annoyed with Rust.

sf_core/src/tls/x509_utils.rs
```diff
@@ -328,10 +376,251 @@ pub fn extract_issuer_skid(issuer_der: &[u8]) -> Option<Vec<u8>> {
...
+            {
+                return Some(kid.0.to_vec());
+            }
+        }
+    }
+    None
+}
+
+// Build candidate chains from an end-entity and a list of intermediates.
+// Each chain is a vector of cert DER bytes from EE up to last found parent.
```
sfc-gh-jkasten, line 444:
```suggestion
// Each chain is a vector of cert DER bytes from EE up to last found parent.
// Chains are not verified.
```

sf_core/src/tls/x509_utils.rs
```diff
@@ -127,17 +183,50 @@ pub fn verify_crl_signature_best_effort(
...
         .as_bytes()
         .context(InvalidCrlSignatureSnafu)?;
-    // First, try verification using aws-lc-rs (ring-compatible API)
-    let try_verify = |alg: &'static dyn aws_lc_rs::signature::VerificationAlgorithm| {
-        aws_lc_rs::signature::UnparsedPublicKey::new(alg, spk_bytes).verify(&tbs, sig)
-    };
+
     use aws_lc_rs::signature::{
         ECDSA_P256_SHA256_ASN1, ECDSA_P384_SHA384_ASN1, ED25519, RSA_PKCS1_2048_8192_SHA256,
         RSA_PKCS1_2048_8192_SHA384, RSA_PKCS1_2048_8192_SHA512, RSA_PSS_2048_8192_SHA256,
```
sfc-gh-jkasten: This is the first driver that I have seen have any support for PSS. Thanks!

nit/optional: I think we should also add
`ring::signature::RSA_PSS_2048_8192_SHA384` and
`ring::signature::RSA_PSS_2048_8192_SHA512`
if it is easy.
https://github.com/cabforum/servercert/blob/main/docs/BR.md#71321-rsa

I also know that P-521 with SHA-512 is allowed, but I don't see that algorithm supported by the package we are using :(
https://github.com/cabforum/servercert/blob/main/docs/BR.md#71312-ecdsa
Mozilla just recently (last few years) started to allow P-521 so I don't think any CAs have really even started to seed the root programs with it. P-384 is already believed to be much stronger than every other algorithm/key size being used (192 bits of security vs. 140 with RSA 4096). There would be compatibility issues with P-521.

I didn't write a quick program to see if anyone is using it, https://www.ccadb.org/resources, but my guess is that either P-384 will be completely broken or we will move to PQ before the ecosystem needs P-521.

Finally, DJB's ED25519 is not allowed. It probably doesn't hurt too much to have it - it is a little attack surface. Occassionally someone pops onto the forums to argue for its inclusion, but I think the likelyhood is very low it ever gets adopted. (HSM support, client support, etc)

sf_core/src/tls/x509_utils.rs
sfc-gh-jkasten: Thanks for the additional pointers, Sean! Yes, I see that now and that makes sense.

sfc-gh-jkasten (approved):
Thanks, Sean!

sf_core/src/tls/crl_verifier.rs
```diff
@@ -46,38 +50,69 @@ impl ServerCertVerifier for CrlServerCertVerifier {
...
-        self.webpki_verifier.verify_server_cert(
-            end_entity,
-            intermediates,
-            server_name,
-            ocsp_response,
-            now,
-        )?;
+        // Helper closure to re-validate a path with fixed verifier inputs
+        let verify_path = |inters: &[rustls::pki_types::CertificateDer<'_>]| {
+            self.webpki_verifier.verify_server_cert(
```
sfc-gh-jkasten, line 55: Note: It isn't in the documentation, but I am assuming that this is checking that each of the interemdiates is actually an intermediate. (basicConstraints). (I think that would be a bug in the library rather than this code)
https://docs.rs/rustls/latest/rustls/client/struct.WebPkiServerVerifier.html#method.verify_server_cert

sf_core/src/tls/x509_utils.rs
sfc-gh-jkasten: Actually, I was trying to take one more look at the code as it got pretty late here. I will just be open with the concern before I fully vet it.

I think that we are creating chains with potentially more certificates than we need. `build_candiate_chains`'s will keep on adding issuers even if a path to a root exists.

We verify each candidate chain we generate (which will work really well for branches in the hierarchy), but that function will ignore additional intermediates at the end of the chain that it doesn't need. I am thinking about chains like the "green" box in this [GTS hierarchy](https://docs.google.com/document/d/1cgkHlAqFBmF0qrnSgdwvGp1RhJdx0w2nayIl_iJIknE/edit?tab=t.l6soesqfmoiq#bookmark=id.93z0hklhv4l5). I believe that the GTS R4 intemedate (signed by GlobalSign) will be included in the candidate chain, but not actually be used by the verifier. It is extraneous.

I don't think this is a problem as long as we short-circuit the CRL verification as soon as we hit the first root. I didn't see that in the crl_validator though. Very possible I missed it. (I didn't check tests and I probably need to dive into the crl_validator code very closely)

sfc-gh-snoonan: We do not short circuit. I've added that.
The alternatives I considered:
1. Pass our TLS verifier into `build_candidate_chains` so we return only anchored chains in the first place.
2. Return all possible short chains from `build_candidate_chains` e.g. [ee, i1], [ee, i1, i2], [ee, i1, i2, i3] etc. We'd then filter these to anchored chains and have no need to short circuit.

Embarrassing as this is basically the same error we had in other drivers 🤦

sfc-gh-jkasten: I think the first alternative is the safest. The construct that it outputs has fewer rough edges and requires the least amount of documentation to explain. 

My comment was just coming from a correctness standpoint. I defer to you on how best to handle it.

sfc-gh-jkasten: FWIW, we "skipped" over this chain building complexity in the design doc [[comment](https://docs.google.com/document/d/1Dy4eb8dDV75XIUKBvOMfeZjnxYca6ICKsCtgWwfU9Fs/edit?disco=AAABk5HwcwU)] as we used Golang's library (the easiest implementation) as the pseudocode.

Chain building is the enduring aspect of any revocation implementation, but we didn't really have any good implementations (Starfield). It is complex.

sfc-gh-snoonan: Okay, I've taken alternative 1; we now build only anchored chains, and we stop building them as soon as we have reach a trustanchor.

sf_core/src/tls/x509_utils.rs
sfc-gh-snoonan: Why ring versus aws_lc_rs? (It's a ~compatible library that has FIPS certification, which is why we chose it).
RSA_PSS_2048_8192_SHA384/512 are on the line below. They're attempted on L256-260 below.
It does support P-521, so I've added it 🎉 .

For ED25519, 'not allowed' is BR, not RFC, right? So I guess we have to choose.

sfc-gh-jkasten: > Why ring versus aws_lc_rs?

I just tried to Google the identifier and didn't pay attention to the package. Glad to see there is support! I wasn't trying to make a statement about packages :)

> RSA_PSS_2048_8192_SHA384/512 are on the line below. They're attempted on L256-260 below.

Oops, Github/Granite cut the line off. Thanks!

> For ED25519, 'not allowed' is BR, not RFC, right? So I guess we have to choose.

Good point ([RFC 8410](https://datatracker.ietf.org/doc/html/rfc8410)). So much of the rest of the code is built to handle more than the Web PKI. It makes sense to keep it.

sf_core/src/tls/x509_utils.rs
sfc-gh-jkasten: Awesome! I think `build_candidate_chains_with_filter` works. I think you might be able to remove `build_candidate_chains`?

sfc-gh-jkasten (approved):
Thanks, Sean!

sf_core/src/tls/x509_utils.rs
sfc-gh-snoonan: Yup. Removed now.

PR #97, opened 2025-09-24
Provide rpath so loading built dependencies works

sfc-gh-snoonan:
We now depend on the aws-lc library, which is built as part of the core. We thus need to add an rpath so the library can be found. We should consider whether we want this to be dynamic or static long term, but fixing the paths at least unblocks testing.

### DIFF
--- a/pep249_dbapi/Makefile
+++ b/pep249_dbapi/Makefile
@@ -7,17 +7,18 @@ REFERENCE_DRIVER_VERSION ?= 3.17.2
 PYTEST_ARGS ?=
 FAIL_ON_REGRESSIONS ?= 0
 
+# Project root (absolute)
+PROJECT_ROOT := $(abspath ..)
+
 # Library search paths (for dependent native libs)
-# Default search locations
-DEFAULT_LIB_DEBUG := $(shell cd .. && pwd)/target/debug
-DEFAULT_LIB_DEPS := $(shell cd .. && pwd)/target/debug/deps
+# Default search locations derived from PROJECT_ROOT
+DEFAULT_LIB_DEBUG := $(PROJECT_ROOT)/target/debug
+DEFAULT_LIB_DEPS := $(DEFAULT_LIB_DEBUG)/deps
 
 # Absolute paths to avoid relative path issues across different working directories
-# Derive all core-related paths from DEFAULT_LIB_DEBUG (../target/debug)
+# Derive all core-related paths from DEFAULT_LIB_DEBUG
 DEFAULT_CORE_PATH_MACOS := $(DEFAULT_LIB_DEBUG)/libsf_core.dylib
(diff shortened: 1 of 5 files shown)


pep249_dbapi/Makefile
```diff
@@ -7,10 +7,18 @@ REFERENCE_DRIVER_VERSION ?= 3.17.2
...
+
 # Absolute paths to avoid relative path issues across different working directories
-DEFAULT_CORE_PATH_MACOS := $(shell cd .. && pwd)/target/debug/libsf_core.dylib
-DEFAULT_CORE_PATH_LINUX := $(shell cd .. && pwd)/target/debug/libsf_core.so
-DEFAULT_PARAMETER_PATH := $(shell cd .. && pwd)/parameters.json
+# Derive all core-related paths from DEFAULT_LIB_DEBUG (../target/debug)
+DEFAULT_CORE_PATH_MACOS := $(DEFAULT_LIB_DEBUG)/libsf_core.dylib
+DEFAULT_CORE_PATH_LINUX := $(DEFAULT_LIB_DEBUG)/libsf_core.so
+# Project root is two levels up from DEFAULT_LIB_DEBUG (../target/debug -> ../..)
+PROJECT_ROOT := $(abspath $(DEFAULT_LIB_DEBUG)/../..)
```
sfc-gh-jszczerbinski: Let's define PROJECT_ROOT first and then define DEFAULT_LIB_DEBUG using project root not the other way around

sf_core/build.rs
```diff
@@ -0,0 +1,40 @@
+use std::env;
+use std::path::PathBuf;
+
+fn main() {
+    // Always include an rpath relative to the loaded library location so colocated deps work
+    if cfg!(target_os = "macos") {
```
sfc-gh-jszczerbinski: This for release?

sf_core/build.rs
```diff
@@ -0,0 +1,40 @@
...
+        println!("cargo:rustc-link-arg=-Wl,-rpath,@loader_path");
+    } else if cfg!(target_os = "linux") {
+        println!("cargo:rustc-link-arg=-Wl,-rpath,$ORIGIN");
+    }
+
+    // Best-effort: add rpath to AWS-LC artifacts if present in target dir layout
+    // target/debug/build/aws-lc-fips-sys-*/out/build/artifacts
+    if let Ok(out_dir) = env::var("OUT_DIR") {
```
sfc-gh-jszczerbinski: This is for development?

sfc-gh-jszczerbinski (commented):
Have you tried running odbc tests too?

sfc-gh-jszczerbinski (approved):

sf_core/build.rs
sfc-gh-snoonan: No, these are cumulative. So both get printed and used by the linker.

pep249_dbapi/Makefile
sfc-gh-snoonan: Done.

PR #100, opened 2025-09-25
Migrate put_get auto-compress tests to e2e format in Core

sfc-gh-akolodziejczyk:


sfc-gh-akolodziejczyk:
* **#100** <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/100?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/snowflakedb/universal-driver/100?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tests/definitions/put_get/put_get_auto_compress.feature
```diff
@@ -0,0 +1,16 @@
...
+    When File is downloaded using GET command
+    Then Only compressed file should be downloaded
+    And Have correct content
+
+  @core_e2e
+  Scenario: should get file uploaded to stage with AUTO_COMPRESS set to false
+    Given File is uploaded to stage with AUTO_COMPRESS set to false
+    When File is downloaded using GET command
+    Then Only uncompressed file should be downloaded
+    And Have correct content
```
sfc-gh-mrolbiecki, line 16: I think those are not communicating the purpose of this test clearly - this test is mostly about testing if we compressed the file before upload or not, and if the compressed file has the correct content. So I think something like this would be more appropriate:

  Scenario: should compress the file before uploading to stage when AUTO_COMPRESS set to true
    Given File is uploaded to stage with AUTO_COMPRESS set to true
    When File is downloaded using GET command
    Then Downloaded file is compressed
    And Has correct content

sfc-gh-mrolbiecki (changes requested):

tests/definitions/put_get/put_get_auto_compress.feature
sfc-gh-akolodziejczyk: I changed the steps, now WHEN block is: 
`When File is uploaded to stage with AUTO_COMPRESS set to false`

sfc-gh-mrolbiecki (approved):

PR #113, opened 2025-10-03
SNOW-2391352: linting Python code fix

sfc-gh-pcyrek:
SNOW-2391352: linting Python code fix

This PR applies black rules to the code


sfc-gh-pcyrek:
Black is to be configured against the project structure. Leave it for now.

PR #115, opened 2025-10-06
SNOW-2371279: Add Python e2e tests for PUT/GET - wildcards and source compression

sfc-gh-pcyrek:
SNOW-2371279: Add Python e2e tests for PUT/GET - wildcards and source compression


pep249_dbapi/tests/e2e/put_get/put_get_helper.py
```diff
@@ -99,6 +100,54 @@ def get_file_from_stage(cursor, stage_name: str, filename: str, download_dir: Pa
     return cursor.fetchone()
 
 
+def upload_files_with_wildcard(
```
sfc-gh-akolodziejczyk: Is this function used in more than one class? If not, we can store it as a helper method in that class. The same for other functions

pep249_dbapi/tests/e2e/put_get/test_put_get_source_compression.py
```diff
@@ -0,0 +1,214 @@
...
+        # When File is uploaded with SOURCE_COMPRESSION set to explicit type
+        put_command = f"PUT 'file://{as_file_uri(test_file_path)}' @{stage_name} SOURCE_COMPRESSION={compression}"
+
+        if compression == "BROTLI" and OLD_DRIVER_ONLY("BC#3"):
+            # BC#3: BROTLI compression type option is now supported
+            with pytest.raises(Exception) as exc_info:
+                cursor.execute(put_command)
+            assert "253007" in str(exc_info.value)
+            assert "Feature is not supported" in str(exc_info.value)
+        else:
```
sfc-gh-akolodziejczyk, line 103: Please change to `elif NEW_DRIVER_ONLY("BC#3")` so it will be clear that this is new behaviour

pep249_dbapi/tests/e2e/put_get/test_put_get_source_compression.py
```diff
@@ -0,0 +1,214 @@
...
+        if expected_compression == "DEFLATE" and OLD_DRIVER_ONLY("BC#2"):
+            expected_target = f"{filename}.gz"
+            assert_put_compression_result(
+                result,
+                filename,
+                "NONE",
+                expected_target,
+                "GZIP",
+            )
+        else:
```
sfc-gh-akolodziejczyk, line 57: Please change to `elif NEW_DRIVER_ONLY("BC#2")` so it will be clear that this is new behaviour

pep249_dbapi/tests/e2e/put_get/test_put_get_source_compression.py
```diff
@@ -0,0 +1,214 @@
...
+            # Old driver supports BROTLI with AUTO_DETECT
```
sfc-gh-akolodziejczyk: Is this comment correct? It is an assertion for a new driver; why is an old driver described here?

pep249_dbapi/tests/e2e/put_get/test_put_get_source_compression.py
```diff
@@ -0,0 +1,214 @@
...
+        # And File with standard type (GZIP, BZIP2, BROTLI, ZSTD, DEFLATE, RAW_DEFLATE)
+        stage_name, test_file_path = create_stage_and_get_compression_file(
+            cursor, f"TEST_STAGE_{compression}", compression
+        )
+
+        # When File is uploaded with SOURCE_COMPRESSION set to explicit type
+        put_command = f"PUT 'file://{as_file_uri(test_file_path)}' @{stage_name} SOURCE_COMPRESSION={compression}"
+
+        if compression == "BROTLI" and OLD_DRIVER_ONLY("BC#3"):
+            # BC#3: BROTLI compression type option is now supported
```
sfc-gh-akolodziejczyk: Please remove the comment; the description in BreakingChanges.md is enough

sfc-gh-akolodziejczyk (changes requested):

pep249_dbapi/tests/e2e/put_get/put_get_helper.py
sfc-gh-pcyrek: This function is encapsulated in create_temporary_stage_and_upload_multiple_files, which is used several times. I would like to divide test abstraction from helper functions. Those functions are well documented, they are used now, and will be reused for further tests.

pep249_dbapi/tests/e2e/put_get/test_put_get_source_compression.py
sfc-gh-pcyrek: alternative approach is to have:

```
if compression == "BROTLI" and OLD_DRIVER_ONLY("BC#3"):
   pass
elif NEW_DRIVER_ONLY("BC#3") or  (OLD_DRIVER_ONLY("BC#3) and compression != "BROTLI" ):
   pass

```   
  
   vs. orginal:
   
```
   if compression == "BROTLI" and OLD_DRIVER_ONLY("BC#3"):
   pass
else:
   pass
```
   I think if it's not required by framework current version is better

pep249_dbapi/tests/e2e/put_get/test_put_get_source_compression.py
sfc-gh-pcyrek: Same as above

pep249_dbapi/tests/e2e/put_get/test_put_get_source_compression.py
sfc-gh-pcyrek: It's not, my fault, I forgot about cucumber requirements trying to keep code commented

pep249_dbapi/tests/e2e/put_get/test_put_get_source_compression.py
sfc-gh-pcyrek: true

pep249_dbapi/tests/e2e/put_get/put_get_helper.py
sfc-gh-akolodziejczyk: I think the good flow is:
- Create a helper function in the same file when more than one test needs it
- If you need it in one more test file, extract it to a helper class

That way, we instantly know if the function is test-specific or more broadly used.
For helper functions in test classes, you can also leave the documentation.
If this function is needed in multiple files, you can leave it in a helper, but please check if that applies to all functions

pep249_dbapi/tests/e2e/put_get/test_put_get_source_compression.py
sfc-gh-pcyrek: proposed

pep249_dbapi/tests/e2e/put_get/test_put_get_source_compression.py
sfc-gh-pcyrek: propsed solution

pep249_dbapi/tests/e2e/put_get/test_put_get_source_compression.py
```diff
@@ -0,0 +1,228 @@
...
+        if compression == "BROTLI":
+            if OLD_DRIVER_ONLY("BC#3"):
+                with pytest.raises(Exception) as exc_info:
+                    cursor.execute(put_command)
+                assert "253007" in str(exc_info.value)
+                assert "Feature is not supported" in str(exc_info.value)
+            elif NEW_DRIVER_ONLY("BC#3"):
+                cursor.execute(put_command)
+                result = cursor.fetchone()
+                assert_put_compression_result(
```
sfc-gh-akolodziejczyk: why additional assert here?

pep249_dbapi/tests/e2e/put_get/test_put_get_source_compression.py
sfc-gh-akolodziejczyk: Looks ok, I think it's the most readable option

pep249_dbapi/tests/e2e/put_get/test_put_get_source_compression.py
sfc-gh-pcyrek: wdym? 

if compression BROTLI and new driver there is one.
if compression not BROTLI and any type of driver there is second

This is required overhead as a side effect of using:

```
        if compression == "BROTLI":
            if OLD_DRIVER_ONLY("BC#3"):
            elif NEW_DRIVER_ONLY("BC#3"):
       else:
       
```

vs.                
```
if compression == "BROTLI" and OLD_DRIVER_ONLY("BC#3"):
else:
```

pep249_dbapi/tests/e2e/put_get/put_get_helper.py
sfc-gh-pczajka: I kinda agree with Adam here - some of these helpers makes sense only in very specific test setup (for example `create_matching_files` only makes sense in wildcard context - I had to think a little what does it do).
What Adam proposed is a good heuristic, however if something sounds useful / does not need additional context to understand, or even better - is used across multiple tests - I'd leave it here (`create_temp_stage_and_upload_file` makes sense for me).

pep249_dbapi/tests/e2e/put_get/test_put_get_wildcards.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def test_should_download_files_that_are_matching_wildcard_pattern(connection):
+    base_file_name = "test_get"
+
+    with tempfile.TemporaryDirectory() as temp_dir:
+        temp_path = Path(temp_dir)
+
+        with connection.cursor() as cursor:
+
```
sfc-gh-pczajka:
```suggestion
    with (tempfile.TemporaryDirectory() as temp_dir, connection.cursor() as cursor):
        temp_path = Path(temp_dir)

```
will reduce indentation level :)
`download_dir` could be moved here as well, or do:
```
temp_path = Path(temp_dir) / "upload"
download_dir = Path(temp_dir) / "download"
```

pep249_dbapi/tests/e2e/put_get/test_put_get_source_compression.py
sfc-gh-pcyrek: discussed offline, fixed with return

pep249_dbapi/tests/e2e/put_get/put_get_helper.py
sfc-gh-pcyrek: Thanks for your review. You're right. I moved specific functions to the test files, leaving only generic ones.

pep249_dbapi/tests/e2e/put_get/test_put_get_wildcards.py
sfc-gh-pcyrek: sure

sfc-gh-akolodziejczyk (approved):
LGTM

PR #133, opened 2025-11-10
SNOW-2371565: Basic HTTP retry policy with exponential backoff

sfc-gh-snoonan:

PR #136, opened 2025-11-10
SNOW-2371565: Provide async query execution and retry policies

sfc-gh-snoonan:
Adds a Snowflake async execution path backed by new async_exec and Snafu SfError, exposes the request option as per-statement async_execution and plumbs it through statement_execute_query, surfaces the necessary response fields for async polling, and adds matching Gherkin/E2E coverage plus helper support in SnowflakeTestClient.

### DIFF
--- a/sf_core/src/rest/snowflake/async_exec.rs
+++ b/sf_core/src/rest/snowflake/async_exec.rs
@@ -1,9 +1,11 @@
 use crate::chunks::ChunkDownloadData;
+use crate::config::rest_parameters::{ClientInfo, QueryParameters};
 use crate::config::retry::{BackoffConfig, RetryPolicy};
 use crate::http::retry::{HttpContext, HttpError, execute_with_retry};
 use crate::rest::snowflake::error::SfError;
-use crate::rest::snowflake::{query_request, query_response};
-use reqwest::header;
+use crate::rest::snowflake::{
+    QUERY_REQUEST_PATH, apply_json_content_type, apply_query_headers, query_request, query_response,
+};
 use reqwest::{Method, StatusCode};
 use snafu::Location;
 use std::collections::HashMap;
@@ -18,8 +20,6 @@ const INLINE_SHORT_POLL_DELAYS: &[Duration] = &[
     Duration::from_millis(20),
     Duration::from_millis(40),
 ];
-const QUERY_REQUEST_PATH: &str = "/queries/v1/query-request";
-const JSON_MIME: &str = "application/json";
 const QUERY_SEQUENCE_ID: u64 = 1;
 
 fn join_server_path(server_url: &str, path: &str) -> Result<String, SfError> {
@@ -59,26 +59,15 @@ fn build_async_query_request(
 fn build_submit_request(
     client: &reqwest::Client,
     endpoint: &str,
+    client_info: &ClientInfo,
     session_token: &str,
     request_id: uuid::Uuid,
     payload: &query_request::Request,
 ) -> reqwest::RequestBuilder {
-    let auth_header = authorization_header(session_token);
-    client
-        .post(endpoint)
-        .header(header::AUTHORIZATION, auth_header)
-        .header(header::ACCEPT, header::HeaderValue::from_static(JSON_MIME))
-        .header(
-            header::CONTENT_TYPE,
-            header::HeaderValue::from_static(JSON_MIME),
-        )
+    let builder = client.post(endpoint);
+    apply_json_content_type(apply_query_headers(builder, client_info, session_token))
         .query(&[("requestId", request_id.to_string())])
         .json(payload)
-}
-
-fn authorization_header(session_token: &str) -> header::HeaderValue {
-    let value = format!("Snowflake Token=\"{session_token}\"");
-    header::HeaderValue::from_str(&value).expect("authorization header construction must succeed")
 }
 
 async fn parse_submit_response(
@@ -133,17 +122,27 @@ fn current_epoch_millis() -> i64 {
 
 pub async fn submit_statement_async(
     client: &reqwest::Client,
-    server_url: &str,
+    params: &QueryParameters,
     session_token: &str,
     sql: String,
     parameter_bindings: Option<&HashMap<String, query_request::BindParameter>>,
     request_id: uuid::Uuid,
     policy: &RetryPolicy,
 ) -> Result<SubmitOk, SfError> {
+    let server_url = &params.server_url;
+    let client_info = &params.client_info;
     let endpoint = join_server_path(server_url, QUERY_REQUEST_PATH)?;
     let request_body = build_async_query_request(sql, parameter_bindings);
-    let submit_request =
-        || build_submit_request(client, &endpoint, session_token, request_id, &request_body);
+    let submit_request = || {
+        build_submit_request(
+            client,
+            &endpoint,
+            client_info,
+            session_token,
+            request_id,
+            &request_body,
+        )
+    };
 
     let ctx = HttpContext::new(Method::POST, QUERY_REQUEST_PATH).allow_post_retry();
     let response = execute_with_retry(submit_request, &ctx, policy, |r| async move { Ok(r) })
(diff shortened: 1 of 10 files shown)


sfc-gh-snoonan:
* **#138** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/138?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#137** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/137?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#136** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/136?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/136?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#133** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/133?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#134](https://github.com/snowflakedb/universal-driver/pull/134) <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/134?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

sf_core/src/rest/snowflake/async_exec.rs
```diff
@@ -0,0 +1,544 @@
...
+use tracing::debug;
+use url::Url;
+
+const INLINE_SHORT_POLL_DELAYS: &[Duration] = &[
+    Duration::from_millis(5),
+    Duration::from_millis(10),
+    Duration::from_millis(20),
+    Duration::from_millis(40),
+];
+const QUERY_REQUEST_PATH: &str = "/queries/v1/query-request";
```
sfc-gh-turbaszek: This path is in use in `mod.rs`. Would it make sense to reuse this constant there or extract common code?

sf_core/src/rest/snowflake/async_exec.rs
```diff
@@ -0,0 +1,544 @@
...
+        is_internal: false,
+        describe_only: None,
+        parameters: None,
+        bindings: parameter_bindings.cloned(),
+        bind_stage: None,
+        query_context: query_request::QueryContext { entries: None },
+    }
+}
+
+fn build_submit_request(
```
sfc-gh-turbaszek, line 59: What is a "submit request"?

sf_core/src/rest/snowflake/async_exec.rs
```diff
@@ -0,0 +1,544 @@
...
+        .header(header::ACCEPT, header::HeaderValue::from_static(JSON_MIME))
+        .header(
+            header::CONTENT_TYPE,
+            header::HeaderValue::from_static(JSON_MIME),
+        )
+        .query(&[("requestId", request_id.to_string())])
+        .json(payload)
+}
+
+fn authorization_header(session_token: &str) -> header::HeaderValue {
```
sfc-gh-turbaszek:
```suggestion
fn build_authorization_header(session_token: &str) -> header::HeaderValue {
```
Or something suggesting that this method does not return constant. WDYt?

sf_core/src/rest/snowflake/async_exec.rs
```diff
@@ -0,0 +1,544 @@
...
+    let value = format!("Snowflake Token=\"{session_token}\"");
```
sfc-gh-turbaszek: Not sure if we have more custom headers, but if yes would it make sense to collect them in single place for reuse?

sf_core/src/rest/snowflake/async_exec.rs
```diff
@@ -0,0 +1,544 @@
...
+pub async fn submit_statement_async(
+    client: &reqwest::Client,
+    server_url: &str,
+    session_token: &str,
+    sql: String,
+    parameter_bindings: Option<&HashMap<String, query_request::BindParameter>>,
+    request_id: uuid::Uuid,
+    policy: &RetryPolicy,
+) -> Result<SubmitOk, SfError> {
+    let endpoint = join_server_path(server_url, QUERY_REQUEST_PATH)?;
```
sfc-gh-turbaszek, line 134: May be worth to reuse in `mod.rs` instead of
```
let query_url = format!("{server_url}/queries/v1/query-request");
```

sf_core/src/rest/snowflake/async_exec.rs
```diff
@@ -0,0 +1,544 @@
...
+            location: current_location(),
+        })
+}
+pub struct SubmitOk {
+    pub query_id: Option<String>,
+    pub get_result_url: Option<String>,
+    pub response: query_response::Response,
+}
+
+fn build_async_query_request(
```
sfc-gh-turbaszek, line 41:
```suggestion
fn build_async_query_request_body(
```
WDYT? Otherwise there's naming inconsistency between `build_async_query_request` and `build_submit_request`. Both in theory build request but that's not true as far as I understand.

sf_core/src/rest/snowflake/async_exec.rs
```diff
@@ -0,0 +1,544 @@
...
+}
+
+#[track_caller]
+fn current_location() -> Location {
+    let caller = StdLocation::caller();
+    Location::new(caller.file(), caller.line(), caller.column())
+}
+
+#[track_caller]
+fn map_http_error(err: HttpError) -> SfError {
```
sfc-gh-turbaszek, line 259:
```suggestion
fn map_http_error_to_sf_error(err: HttpError) -> SfError {
```
WDYT?

sf_core/src/rest/snowflake/async_exec.rs
```diff
@@ -0,0 +1,544 @@
...
+        } => SfError::RetryBudgetExceeded {
+            retry_after,
+            remaining,
+            location,
+        },
+    }
+}
+
+#[track_caller]
+fn transport_error(source: reqwest::Error) -> SfError {
```
sfc-gh-turbaszek, line 294:
```suggestion
fn build_transport_error(source: reqwest::Error) -> SfError {
```
Similar to previous comment and would apply to few other methods later in the code. But I may be unaware of rust naming conventions

sf_core/src/rest/snowflake/async_exec.rs
sfc-gh-snoonan: Rust naming conventions are to describe what a function returns more than how it computes it. Appending build_ is usually for builders that consume inputs in a notable way. Since this is just a helper returning the header value, I think the current name is idiomatic.

sf_core/src/rest/snowflake/async_exec.rs
sfc-gh-snoonan: It's the POST taht we'll eventually do to the async endpoint; it builds the request, but doesn't send it. I'm open to a better name, but it isn't the query itself...

sf_core/src/rest/snowflake/async_exec.rs
sfc-gh-snoonan: One builds a request, the other builds a request builder. This isn't really just a 'body'

A lot of this comes from the types rather than the naming tbh...

sf_core/src/rest/snowflake/async_exec.rs
sfc-gh-snoonan: Done.

sf_core/src/rest/snowflake/async_exec.rs
sfc-gh-snoonan: Good catch, thanks. Done.

sf_core/src/rest/snowflake/async_exec.rs
sfc-gh-snoonan: I don't think we map to anything else yet, so convention would be to leave it until we do. I can't think of what else we'd map to...
(Ideally this wouldn't even be necessary, as they'd return different types, but rust doesn't support function overloading based only on return types :-().

Even then , I think we'd switch to doing traits instead.

sf_core/src/rest/snowflake/async_exec.rs
sfc-gh-snoonan: Same response - this is a simple construction rather than a builder.

sfc-gh-turbaszek (approved):
Looks reasonable as far as I can tell 👌

PR #137, opened 2025-11-10
Switch driver API to shared TLS client + retry-aware async path

sfc-gh-snoonan:
Connections now build one TLS client up front, stash it with a default retry policy, and statements reuse it via snowflake_query_with_client, so async/blocking calls share headers and timeouts consistently.
Chunk downloads run through a lazy global HTTP client with the retry helper.

PR #138, opened 2025-11-10
SNOW-2371565: Enable async execution by default with auto-detection for blocking statements

sfc-gh-snoonan:
Queries now use async mode unless explicitly disabled or the statement is PUT/GET (which requires blocking)
Add metrics for submit, inline poll, and wait phase timings for performance analysis.

Default behavior is submit, get query id, poll for results. Remaining TODO is to handle session refresh. This results in at least one additional round trip. For a user running in the same cloud region, this should be negligible (~5 ms). From my laptop hitting us-west-2, it adds another 180ms. I think this is the right tradeoff for us.

On the snowflake backend, this will cause all queries will be persisted. This is probably too expensive to keep this as the default long term due to the effect on hybrid tables latency and the load on FDB.

### DIFF
--- a/sf_core/src/apis/database_driver_v1/statement.rs
+++ b/sf_core/src/apis/database_driver_v1/statement.rs
@@ -154,7 +154,7 @@ pub fn statement_execute_query(stmt_handle: Handle) -> Result<ExecuteResult, Api
     let mut stmt = stmt_ptr
         .lock()
         .map_err(|_| StatementLockingSnafu {}.build())?;
-    let query = stmt.query.take().ok_or_else(|| {
+    let query_str = stmt.query.as_deref().ok_or_else(|| {
         InvalidArgumentSnafu {
             argument: "Query not found".to_string(),
         }
@@ -184,6 +184,9 @@ pub fn statement_execute_query(stmt_handle: Handle) -> Result<ExecuteResult, Api
         )
     };
 
+    let execution_mode = stmt.execution_mode(Some(query_str));
+    let query = stmt.query.take().expect("query must be present");
+
     let response = rt
         .block_on(snowflake_query_with_client(
             &http_client,
@@ -197,7 +200,7 @@ pub fn statement_execute_query(stmt_handle: Handle) -> Result<ExecuteResult, Api
                 .build()
             })?,
             &retry_policy,
-            stmt.execution_mode(),
+            execution_mode,
         ))
         .context(LoginSnafu)?;
 
@@ -316,38 +319,52 @@ impl Statement {
         }
     }
 
-    fn execution_mode(&self) -> QueryExecutionMode {
+    fn execution_mode(&self, query: Option<&str>) -> QueryExecutionMode {
         match self
             .settings
             .get(snowflake::STATEMENT_ASYNC_EXECUTION_OPTION)
         {
             Some(Setting::String(value)) => {
-                if value.eq_ignore_ascii_case("true")
-                    || value.eq_ignore_ascii_case("yes")
-                    || value == "1"
+                let normalized = value.trim();
+                if normalized.eq_ignore_ascii_case("false")
+                    || normalized.eq_ignore_ascii_case("no")
+                    || normalized == "0"
                 {
-                    QueryExecutionMode::Async
-                } else {
                     QueryExecutionMode::Blocking
+                } else {
+                    QueryExecutionMode::Async
                 }
             }
             Some(Setting::Int(value)) => {
-                if *value != 0 {
-                    QueryExecutionMode::Async
-                } else {
+                if *value == 0 {
                     QueryExecutionMode::Blocking
+                } else {
+                    QueryExecutionMode::Async
                 }
             }
             Some(Setting::Double(value)) => {
-                if *value != 0.0 {
-                    QueryExecutionMode::Async
-                } else {
+                if *value == 0.0 {
                     QueryExecutionMode::Blocking
+                } else {
+                    QueryExecutionMode::Async
                 }
             }
-            Some(Setting::Bytes(_)) | None => QueryExecutionMode::Blocking,
+            Some(Setting::Bytes(_)) => QueryExecutionMode::Async,
+            None => match query {
+                Some(sql) if requires_blocking(sql) => QueryExecutionMode::Blocking,
+                _ => QueryExecutionMode::Async,
+            },
         }
     }
+}
+
+fn requires_blocking(sql: &str) -> bool {
+    let trimmed = sql.trim_start();
+    if trimmed.is_empty() {
+        return false;
+    }
+    let upper = trimmed.to_ascii_uppercase();
+    upper.starts_with("PUT ") || upper == "PUT" || upper.starts_with("GET ") || upper == "GET"
 }
 
 #[derive(Snafu, Debug)]
--- a/sf_core/src/apis/database_driver_v1/statement.rs
+++ b/sf_core/src/apis/database_driver_v1/statement.rs
@@ -324,32 +324,11 @@ impl Statement {
             .settings
             .get(snowflake::STATEMENT_ASYNC_EXECUTION_OPTION)
         {
-            Some(Setting::String(value)) => {
-                let normalized = value.trim();
-                if normalized.eq_ignore_ascii_case("false")
-                    || normalized.eq_ignore_ascii_case("no")
-                    || normalized == "0"
-                {
-                    QueryExecutionMode::Blocking
-                } else {
-                    QueryExecutionMode::Async
-                }
-            }
-            Some(Setting::Int(value)) => {
-                if *value == 0 {
-                    QueryExecutionMode::Blocking
-                } else {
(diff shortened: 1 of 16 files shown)


sfc-gh-snoonan:
* **#138** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/138?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/138?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#137** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/137?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#136** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/136?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#133** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/133?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#134](https://github.com/snowflakedb/universal-driver/pull/134) <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/134?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

sf_core/src/apis/database_driver_v1/statement.rs
```diff
@@ -316,40 +319,51 @@ impl Statement {
...
             }
-            Some(Setting::Bytes(_)) | None => QueryExecutionMode::Blocking,
         }
+        Setting::Int(v) => Some(*v != 0),
+        _ => None,
     }
 }
 
+fn requires_blocking(sql: &str) -> bool {
+    let s = sql.trim_start();
```
sfc-gh-pfus: I think we should also trim comments - we had some bugs about that

sf_core/src/apis/database_driver_v1/statement.rs
```diff
@@ -316,40 +319,51 @@ impl Statement {
...
+    if s.len() < 3 {
+        return false;
+    }
+    let prefix = &s[..3];
+    let is_put_or_get = prefix.eq_ignore_ascii_case("PUT") || prefix.eq_ignore_ascii_case("GET");
+    is_put_or_get && (s.len() == 3 || s.as_bytes()[3] == b' ')
```
sfc-gh-pfus: I think that if the command size is 3, it is not a valid PUT :) This line also doesn't support newlines after PUT or comments.

sf_core/src/apis/database_driver_v1/statement.rs
```diff
@@ -316,40 +319,51 @@ impl Statement {
...
     }
 }
 
+fn requires_blocking(sql: &str) -> bool {
+    let s = sql.trim_start();
+    if s.len() < 3 {
+        return false;
+    }
+    let prefix = &s[..3];
+    let is_put_or_get = prefix.eq_ignore_ascii_case("PUT") || prefix.eq_ignore_ascii_case("GET");
```
sfc-gh-pfus, line 372: General comment - why do we need to handle this? I may be wrong, but I think our drivers "guess" that the query is PUT/GET for some historical reasons, or it was just incorrect. I looked at go driver code and there's no point in analyzing the query on the driver side. We use those regexes for two cases:
1. Adding `Accept: application/json` header, which, I believe, can be added in all cases, not file transfer only.
2. Trigger file transfer code instead of regular response parsing. But it can be safely replaced with checking that the response (which we already have) contains file transfer metadata.

sf_core/src/apis/database_driver_v1/statement.rs
```diff
@@ -316,40 +319,51 @@ impl Statement {
         }
     }
 
-    fn execution_mode(&self) -> QueryExecutionMode {
+    fn execution_mode(&self, query: Option<&str>) -> QueryExecutionMode {
         match self
             .settings
             .get(snowflake::STATEMENT_ASYNC_EXECUTION_OPTION)
```
sfc-gh-pfus, line 325: Can you explain what is checked here? Is it a client decision to trigger a query in async mode?

sf_core/src/rest/snowflake/async_exec.rs
```diff
@@ -308,6 +390,19 @@ fn body_parse_error(source: serde_json::Error) -> SfError {
 
 #[track_caller]
 fn http_status_error(status: StatusCode) -> SfError {
+    // TODO(session-refresh): Implement automatic session renewal on 401.
```
sfc-gh-pfus: Shouldn't we add a Jira ID here?

tests/performance/drivers/core/app/src/config.rs
```diff
@@ -56,14 +57,51 @@ impl TestConfig {
...
+
+/// Parse async execution override from string value.
+/// Returns `Some(true)` for enabled, `Some(false)` for disabled, `None` for auto/unset.
+fn parse_async_override(value: &str) -> Result<Option<bool>> {
+    let trimmed = value.trim();
+    if trimmed.is_empty() {
+        return Ok(None);
+    }
+    match trimmed.to_ascii_lowercase().as_str() {
+        "true" | "1" | "yes" => Ok(Some(true)),
```
sfc-gh-pfus: We can add also `y` here :) Jokes aside - I think such a logic may be helpful in other cases too, shouldn't we extract this?

sf_core/src/apis/database_driver_v1/statement.rs
sfc-gh-snoonan: So if I run an async=true PUT:
```
{
  "data" : {
    "queryId" : "01c13818-0813-3dc6-000c-a90e5ef44ad3",
    "getResultUrl" : "/queries/01c13818-0813-3dc6-000c-a90e5ef44ad3/result",
    "queryAbortsAfterSecs" : 300,
    "progressDesc" : null
  },
  "code" : "333334",
  "message" : "Asynchronous execution in progress. Use provided query id to perform query monitoring and management.",
  "success" : true
}
```
and then poll:
```
{
  "data" : {
    "errorCode" : "000612",
    "age" : 0,
    "sqlState" : "02000",
    "queryId" : null
  },
  "code" : "000612",
  "message" : "Result not found",
  "success" : false,
  "headers" : null
}
```
GET shows the same behavior.

So we don't actually get file transfer metadata with async submission. I've made this a bit more robust with the initial detection (thanks for the edge case feedback :-() and now we retry in sync if we immediately get back a 612.

sfc-gh-snoonan: Rather than try to reverse engineer this, I asked claude to look at the snowflake source code and trace the async behavior versus the changes I made:

Based on my investigation of the Snowflake server code, here's what I found:

## Key Findings

**Error 612 (`RESULT_NOT_FOUND`)** is defined in `ErrorCode000000.java`:
```java
RESULT_NOT_FOUND("000612", SqlState.NO_DATA)
```

**PUT/GET behavior with async** - From `QueriesInternal.java` line 4241:
> "This is commonly seen on PUT/GET queries: QC immediately unregisters and exits after execution, because it has no knowledge of client states."

**The flow:**
1. Submit query with `asyncExec=true`
2. Server returns `QueryInProgressResponse` with `queryId` and `getResultUrl`
3. Client polls `getResultUrl`
4. For PUT/GET, the Query Coordinator exits immediately → results aren't persisted → poll returns 612

**No explicit rejection** - The server doesn't block async mode for file transfers. It accepts the request, gives you a poll URL, but that URL returns 612 because results were never stored.

## Our Implementation is Correct

1. **Detect 612 on first poll** → retry with sync mode ✓
2. **612 after prior polls** → error, don't retry ✓
3. **SQL parsing for early detection** → optimization to avoid wasted round-trip ✓

The only alternative would be for Snowflake to either:
- Reject `asyncExec=true` for PUT/GET upfront (they don't)
- Return file transfer metadata in the async submission response (they don't - only `queryId` and poll URL)

Our approach matches how the problem needs to be solved client-side.

tests/performance/drivers/core/app/src/config.rs
sfc-gh-snoonan: Sure, done.

sf_core/src/apis/database_driver_v1/statement.rs
sfc-gh-pfus: Ok, so it's sad that we have to do parsing on our own. There are plenty of corner cases that we would have to take care of, like various whitespaces, comments and who knows what more...

sf_core/src/apis/database_driver_v1/statement.rs
sfc-gh-snoonan: It won't usually be set, but yes it can be a client decision. In the rust client it would be statement_set_option - which can get used from any of the wrappers.

For example, I used the environment variable to do perf tests to compare the async+poll vs sync.

sfc-gh-pfus (approved):

PR #145, opened 2025-11-25
NO-SNOW: Load core module from python package data

sfc-gh-turbaszek:
Load core module from Python package sources instead of environment variable path. Adds `make dev-build` for easy development setup to populate `snowflake.ud_connector._core` with core binaries.

### DIFF
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -31,14 +31,15 @@ jobs:
           sudo apt-get update
 
       - name: Build Rust core library
+        working-directory: ./python
         run: |
-          RUSTFLAGS="" cargo build --package sf_core
+          RUSTFLAGS="" make dev-build
 
       - name: Upload Rust artifacts
         uses: actions/upload-artifact@v4
         with:
           name: rust-core-library
-          path: target/debug/libsf_core.so
+          path: python/src/snowflake/ud_connector/_core/libsf_core.so
 
   # Combined unit + integration tests with universal connector
   python_tests:
@@ -76,7 +77,7 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           name: rust-core-library
-          path: target/debug/
+          path: python/src/snowflake/ud_connector/_core/
 
       - name: Decode secrets
         run: ./scripts/decode_secrets.sh
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -17,7 +17,7 @@ env:
   REFERENCE_PYTHON_VERSION: "3.13"
   # OS version for consistent naming across jobs
   REFERENCE_OS_VERSION: "ubuntu-latest"
-  PYTHON_CORE_DOWNLOAD_PATH: "python/src/snowflake/ud_connector/_core"
+  PYTHON_CORE_DOWNLOAD_PATH: "${{ github.workspace }}/python/src/snowflake/ud_connector/_core"
 
 jobs:
   # Build Rust core library once and share across Python jobs
@@ -167,7 +167,7 @@ jobs:
 
       - name: Run reference integration tests
         env:
-          CORE_PATH: ${{ github.workspace }}/target/debug/libsf_core.so
+          CORE_PATH: ${{ env.PYTHON_CORE_DOWNLOAD_PATH }}
           PARAMETER_PATH: ${{ github.workspace }}/parameters.json
           REFERENCE_DRIVER_VERSION: ${{ env.PYTHON_REFERENCE_DRIVER_VERSION }}
         run: make ci-test-reference PYTHON_VERSION=${{ env.REFERENCE_PYTHON_VERSION }} OS_VERSION=${{ env.REFERENCE_OS_VERSION }}
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -17,7 +17,7 @@ env:
   REFERENCE_PYTHON_VERSION: "3.13"
   # OS version for consistent naming across jobs
   REFERENCE_OS_VERSION: "ubuntu-latest"
-  PYTHON_CORE_DOWNLOAD_PATH: "${{ github.workspace }}/python/src/snowflake/ud_connector/_core"
+  PYTHON_CORE_MODULE_PATH: "${{ github.workspace }}/python/src/snowflake/ud_connector/_core"
 
 jobs:
   # Build Rust core library once and share across Python jobs
@@ -40,7 +40,7 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: rust-core-library
-          path: "${{ env.PYTHON_CORE_DOWNLOAD_PATH }}/libsf_core.so"
+          path: "${{ env.PYTHON_CORE_MODULE_PATH }}/libsf_core.so"
 
   # Combined unit + integration tests with universal connector
   python_tests:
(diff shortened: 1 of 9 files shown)


sfc-gh-snoonan (approved):

python/Makefile
```diff
@@ -217,6 +217,10 @@ clean-reports:
 
 clean: clean-reports # Short alias
 
+dev-build:
+	cd .. && cargo build --package sf_core --target-dir=python/src/snowflake/ud_connector/_core \
+	&& cd python/src/snowflake/ud_connector/_core && cp -R debug/ . &&  rm -rf debug/
```
sfc-gh-snoonan: Needs a fix for the cp -R debug/ .

python/Makefile
```diff
@@ -217,6 +217,10 @@ clean-reports:
 
 clean: clean-reports # Short alias
 
+dev-build:
+	cd .. && cargo build --package sf_core --target-dir=python/src/snowflake/ud_connector/_core \
```
sfc-gh-fpawlowski: do we want to update the check-env-or-build target in this Makefile as well?

python/src/snowflake/ud_connector/_internal/api_client/c_api.py
```diff
@@ -1,22 +1,48 @@
...
+        lib_name = f"{_CORE_LIB_NAME}.dll"
+    elif sys.platform.startswith("darwin"):
+        lib_name = f"{_CORE_LIB_NAME}.dylib"
+    else:
+        lib_name = f"{_CORE_LIB_NAME}.so"
+
+    files = resources.files("snowflake.ud_connector._core")
+    return files.joinpath(lib_name)
+
+def _load_minicore() -> ctypes.CDLL:
```
sfc-gh-fpawlowski: Is it minicore or just core?

sfc-gh-fpawlowski (approved):
Left 2 nits

python/src/snowflake/ud_connector/_internal/api_client/c_api.py
sfc-gh-turbaszek: Good catch 👌

python/Makefile
sfc-gh-turbaszek: Done

python/Makefile
sfc-gh-turbaszek: Good point, I figured out how all of this works, updated

.github/workflows/test-python.yml
```diff
@@ -17,6 +17,7 @@ env:
   REFERENCE_PYTHON_VERSION: "3.13"
   # OS version for consistent naming across jobs
   REFERENCE_OS_VERSION: "ubuntu-latest"
+  PYTHON_CORE_DOWNLOAD_PATH: "${{ github.workspace }}/python/src/snowflake/ud_connector/_core"
```
sfc-gh-fpawlowski: nit: upload as well uses it so better name might be PYTHON_CORE_ARTIFACT_PATH

python/src/snowflake/ud_connector/_internal/api_client/c_api.py
```diff
@@ -1,22 +1,48 @@
...
+def _get_core_path():
+    # Define the file name for each platform
+    if sys.platform.startswith("win"):
+        lib_name = f"{_CORE_LIB_NAME}.dll"
+    elif sys.platform.startswith("darwin"):
+        lib_name = f"{_CORE_LIB_NAME}.dylib"
+    else:
+        lib_name = f"{_CORE_LIB_NAME}.so"
+
+    files = resources.files("snowflake.ud_connector._core")
```
sfc-gh-fpawlowski, line 28: Should we extract this into some constant?

python/Makefile
```diff
@@ -13,8 +13,7 @@ PROJECT_ROOT := $(abspath ..)
 
 # Library search paths (for dependent native libs)
 # Default search locations derived from PROJECT_ROOT
-DEFAULT_LIB_DEBUG := $(PROJECT_ROOT)/target/debug
-DEFAULT_LIB_DEPS := $(DEFAULT_LIB_DEBUG)/deps
+DEFAULT_LIB_DEBUG := $(PROJECT_ROOT)/python/src/snowflake/ud_connector/_core
```
sfc-gh-fpawlowski: is it just DEFAULT_LIB_PATH or DEFAULT_CORE_DIR_PATH or is the name not outdated?

python/Makefile
```diff
@@ -32,9 +31,7 @@ SHELL := bash
 
 # Platform-aware path detection with environment variable override
 DETECTED_CORE_PATH := $(shell \
-	if [ -n "$$CORE_PATH" ]; then \
```
sfc-gh-fpawlowski, line 35: Not being able to override the CORE_PATH for the test run or other make commands may be limiting.
In JDBC i think such approach is utilised as well.

If we needed to test using specific artifact, it may be easier to change the env var rather then move them and rename each time.

python/Makefile
```diff
@@ -71,16 +68,17 @@ TOX_E2E_ENV = py$(PYTHON_ID)-e2e
...
 	@echo "Using CORE_PATH: $(DETECTED_CORE_PATH)"
 	@echo "Using PARAMETER_PATH: $(DETECTED_PARAMETER_PATH)"
-	cd .. && cargo build --package sf_core || { \
+	cd .. && cargo build --package sf_core --target-dir=$(DEFAULT_LIB_DEBUG) && cd $(DEFAULT_LIB_DEBUG) && mv debug/* . &&  rm -rf debug/ || { \
 		echo "Failed to build Rust core library"; \
 		echo "Make sure Rust toolchain is installed: curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh"; \
 		exit 1; \
 	}
+
+check-env-or-build: build-core
```
sfc-gh-fpawlowski, line 81: nit: was it happening in wrong order in the original target already? Shouldn't check for env var be happening first? Nothing to change in this PR - just out of curiosity

python/src/snowflake/ud_connector/_internal/api_client/c_api.py
sfc-gh-turbaszek: I'm not sure, this value is ready in this single place only

python/Makefile
sfc-gh-turbaszek: The `CORE_PATH` variable wasn't used only for testing. The whole implementation of Python wrapper was based on it. Imo such a approach in production code is no go because of being a potential vector of attack. 

Other aspect is that, we may consider building wheel for tests. That would give us reproducible artifact.

python/Makefile
sfc-gh-turbaszek: Yes, the original implementation was more of `build-and-check-env`. Given our discussion about hatch I would like to move most of those steps to `pyproject.toml` and abandon make.

python/Makefile
sfc-gh-turbaszek: Good point, fixed

.github/workflows/test-python.yml
sfc-gh-turbaszek: Good point, fixed

sfc-gh-fpawlowski (approved):
lgtm

python/Makefile
sfc-gh-fpawlowski: thats true, convinced : )

PR #150, opened 2025-12-02
SNOW-2681494-creating PUT-GET ODBC tests for Universal Driver

sfc-gh-pcyrek:
SNOW-2681494-creating PUT-GET tests for Universal Driver in Gherkin format


sfc-gh-akolodziejczyk (approved):
From the test framework perspective, it looks okay. Please get a review for ODBC code and discuss if we should extract and reuse some of the methods

PR #154, opened 2025-12-04
SNOW-2681494: odbc PUT-GET new tests

sfc-gh-pcyrek:
SNOW-2681494: odbc PUT-GET new tests


sfc-gh-akolodziejczyk (approved):
From the test framework perspective, it looks okay. Please get a review for ODBC code and discuss if we should extract and reuse some of the methods

odbc_tests/tests/e2e/put_get/put_get_auto_compress.cpp
```diff
@@ -0,0 +1,99 @@
...
+  REQUIRE(!fs::exists(download_dir / filename));
+
+  // And Have correct content
+  std::ifstream dl(download_dir / compressed, std::ios::binary);
+  std::string downloaded_bytes((std::istreambuf_iterator<char>(dl)), std::istreambuf_iterator<char>());
+  std::ifstream ref(file_gz, std::ios::binary);
+  std::string reference_bytes((std::istreambuf_iterator<char>(ref)), std::istreambuf_iterator<char>());
+
+  OLD_DRIVER_ONLY("BD#1") { CHECK(downloaded_bytes != reference_bytes); }
+  NEW_DRIVER_ONLY("BD#1") { CHECK(downloaded_bytes == reference_bytes); }
```
sfc-gh-pbulawa: Why there's a different behavior of old driver and new driver? Shouldn't it return the same file content? Does the old driver automatically decompress the file?

odbc_tests/tests/e2e/put_get/put_get_basic_operations.cpp
```diff
@@ -0,0 +1,159 @@
...
+  // Given File is uploaded to stage
+  Connection conn;
+  const std::string stage = pg_utils::create_stage(conn, "ODBCTST_GET_ROWSET_GET");
+  auto [filename, file] = basic_test_file();
+
+  std::string put_sql = "PUT 'file://" + as_file_uri(file) + "' @" + stage;
+  conn.execute(put_sql);
+
+  // When File is downloaded using GET command
+  fs::path download_dir = fs::temp_directory_path() / (std::string("odbc_put_get_") + random_hex());
```
sfc-gh-pbulawa: Do we have a mechanism in place to clean up the temporary folders? What about using the RAII pattern to create the temporary folder and then clean it up when the object is released? It could be reusable between tests, E.g. `TempTestFile / TempTestDir` class based on the code already available in ODBC: https://github.com/snowflakedb/libsnowflakeclient/blob/b3dfbb87e2ea21b6b60a935f5912eaadae0fb3a9/tests/utils/FileCleanup.hpp

Example usage:
```c++
TempTestFile tempFile(fs::temp_directory_path() / (std::string("odbc_put_get_") + random_hex()));
```

odbc_tests/tests/e2e/put_get/put_get_source_compression.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+  // Then Target compression has GZIP type and all PUT results are correct
+  CHECK(get_data<SQL_C_CHAR>(stmt, PUT_ROW_SOURCE_IDX) == filename);
+  CHECK(get_data<SQL_C_CHAR>(stmt, PUT_ROW_TARGET_IDX) == filename + ".gz");
+  compare_compression_type(get_data<SQL_C_CHAR>(stmt, PUT_ROW_SOURCE_COMPRESSION_IDX), "NONE");
+  compare_compression_type(get_data<SQL_C_CHAR>(stmt, PUT_ROW_TARGET_COMPRESSION_IDX), "GZIP");
+  CHECK(get_data<SQL_C_CHAR>(stmt, PUT_ROW_STATUS_IDX) == std::string("UPLOADED"));
+}
+
+TEST_CASE("should compress uncompressed file when SOURCE_COMPRESSION set to NONE and AUTO_COMPRESS set to TRUE",
```
sfc-gh-pbulawa, line 155: Should we check what happens if SOURCE_COMPRESSION is different than the algorithm used for already compressed file?

odbc_tests/tests/e2e/put_get/put_get_wildcards.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include "get_data.hpp"
+#include "put_get_utils.hpp"
+#include "utils.hpp"
+
+namespace fs = std::filesystem;
+using namespace pg_utils;
+
+static fs::path wildcard_tests_dir() { return test_utils::shared_test_data_dir() / "wildcard"; }
+
+static std::string unique_stage_name(const std::string& prefix) { return prefix + "_" + random_hex(4); }
```
sfc-gh-pbulawa: Do we clean up the databases/stages that we use afterwards?

sfc-gh-pbulawa (changes requested):

odbc_tests/tests/e2e/put_get/put_get_auto_compress.cpp
sfc-gh-pcyrek: Behavioral differences between old and new driver, defined in the BehavioralDifferences.yaml file

sfc-gh-pbulawa: Please update the BD# as it seems that the code does not correspond to the real cause of the difference.

odbc_tests/tests/e2e/put_get/put_get_source_compression.cpp
sfc-gh-pcyrek: this is a good observation, and such test should be added to all driver, tho not part of this PR, as it's only test parity

odbc_tests/tests/e2e/put_get/put_get_wildcards.cpp
sfc-gh-pcyrek: Yes - we use CREATE OR REPLACE TEMPORARY STAGE, which is automatically cleaned up by Snowflake when the connection ends

sfc-gh-pbulawa (approved):

PR #156, opened 2025-12-05
NO-SNOW: Add license file and notice

sfc-gh-turbaszek:


sfc-gh-jszczerbinski (approved):
I like the copilot to make the link to the license prettier, otherwise LGTM

PR #157, opened 2025-12-09
Add Python nanoarrow converter to Python wrapper

sfc-gh-pczajka:
This PR moves nanoarrow conversion implementation from old connector and tests it on basic numeric types.
Changes:
 * move .cpp files from old connector: #166 
 * adjust C++ code to consume arrow batches from rust stream
 * add setup.py required for complicated compilation logic (flags dependent on OS etc), based on old connector implementation
 * add `CArrowStreamIterator`, which consumes batches from rust stream and pass them to nanoarrow converters, returning row iterator
 * generate `arrow_stream_iterator.pyx` using cursor - simple wrapper of C++ code fed to Python

Follow-up PR (reenable clang) - https://github.com/snowflakedb/universal-driver/pull/189


.pre-commit-config.yaml
```diff
@@ -22,13 +22,13 @@ repos:
...
         types: [file, rust]
         language: system
-    -   id: clang-format
-        name: clang-format
-        description: Run clang-format on C/C++ files included in the commit. clang-format should be installed before-hand.
-        entry: clang-format -i
-        pass_filenames: true
-        types_or: [c, c++]
-        language: system
+    # -   id: clang-format
```
sfc-gh-pczajka: Temporary, should be reverted

python/src/snowflake/ud_connector/_internal/arrow_context.py
```diff
@@ -0,0 +1,226 @@
+#!/usr/bin/env python
```
sfc-gh-pczajka: copy of arrow_context.py from old connector. Class called by C++ iterators when no appropriate converter was found (for example large values of Decimal are passed here)

python/src/snowflake/ud_connector/_internal/nanoarrow_cpp/ArrowIterator/ConverterUtil.cpp
```diff
@@ -0,0 +1,587 @@
...
+    }
+
+    PyDict_SetItemString(pydict, colName, val);
+    Py_DECREF(val);  // PyDict_SetItemString increments reference
+  }
+
+  m_latestReturnedRow.reset(pydict);
+}
+
+std::shared_ptr<sf::IColumnConverter> getConverterFromSchema(ArrowSchema* schema,
```
sfc-gh-pczajka, line 27: Function moved from CArrowChunkIterator.cpp

python/tests/e2e/types/test_numeric.py
```diff
@@ -0,0 +1,106 @@
+from decimal import Decimal
+
+
+def test_should_cast_number_and_its_synonyms_to_appropriate_type_and_preserve_values_when_selecting_literals(
+    cursor,
+):
+    # Given Snowflake client is logged in
+    assert not cursor.connection.is_closed(), "Connection should be open"
```
sfc-gh-turbaszek: Probably redundant or should be moved to cursor fixture

python/tests/e2e/types/test_numeric.py
```diff
@@ -0,0 +1,106 @@
...
+            int("123456789"),
+            Decimal("123456.789"),
+            Decimal("123.0"),
+            Decimal("0.123456789"),
+        )
+        assert values == expected_values, f"Expected {expected_values}, got {values}"
+    finally:
+        try:
+            cursor.execute(f"DROP TABLE IF EXISTS {table_name}")
+        except Exception:
```
sfc-gh-turbaszek: Instead of that I would suggest providing a schema or opening a cursor with schema and the deleting the schema after tests. That will be easier to reuse and will clean up everything

tests/definitions/types/numeric.feature
```diff
@@ -0,0 +1,26 @@
+@python
+Feature: Numeric Types Support
+
+  @python
+  Scenario: should cast NUMBER and its synonyms to appropriate type and preserve values when selecting literals
+    Given Snowflake client is logged in
+    When Query selecting literal values of NUMBER, DECIMAL, DEC, NUMERIC types is executed
+    Then All returned values should be of appropriate type
```
sfc-gh-turbaszek: Imo we should split this to case by case for example:
> when query selecting NUMBER then it should be casted...

This would give more information. On the other hand it must be kept with sync with code...

sfc-gh-pczajka (commented):
Comments for reviewers

python/tests/e2e/types/test_numeric.py
sfc-gh-pczajka: It was more "something should be under this comment" case, I changed to less cryptic "assert cursor"

sfc-gh-fpawlowski:
We have conflicts with the main branch - are they a reason for jobs lock?

python/src/snowflake/ud_connector/nanoarrow_cpp/ArrowIterator/arrow_batch_iterator.pyx
```diff
@@ -0,0 +1,197 @@
...
+    Converts Arrow RecordBatch to Python tuples/dicts row-by-row.
+    """
+    cdef CArrowBatchIterator* iterator
+    cdef DictCArrowBatchIterator* dict_iterator
+    cdef cpp_bool use_dict_result
+    cdef object arrow_context
+    
+    def __cinit__(
+        self,
+        object record_batch,
```
sfc-gh-jszczerbinski: I think we shouldn't rely on pyarrow here. From Core we get a pointer to https://arrow.apache.org/docs/format/CStreamInterface.html which is very good access to c_array and c_schema so we can skip pyarrow altogether

sfc-gh-jszczerbinski (changes requested):
In general looks good to me. Seems like nanoarrow is up and running in UD, but I think the goal of this PR should be to drop dependency to pyarrow in code altogether

tests/definitions/types/numeric.feature
sfc-gh-pczajka: Moved to a different PR

python/tests/e2e/types/test_numeric.py
sfc-gh-pczajka: done

python/src/snowflake/ud_connector/nanoarrow_cpp/ArrowIterator/arrow_batch_iterator.pyx
sfc-gh-pczajka: Good point! Changed

sfc-gh-turbaszek:
@sfc-gh-pczajka should we move whole logic of type parsing under `snowflake._internal`? I don't think this needs to be exposed as public API. WDYT?

python/setup.py
```diff
@@ -0,0 +1,173 @@
...
+        ],
+    )
+
+    class MyBuildExt(build_ext):
+        def build_extension(self, ext):
+            if options["debug"]:
+                ext.extra_compile_args.append("-g")
+                ext.extra_link_args.append("-g")
+                ext.extra_compile_args.append("-O0")
+                ext.extra_link_args.append("-O0")
```
sfc-gh-turbaszek: Can we convert this to `.extend(["-g" , "-O0"])` wdyt?

python/setup.py
```diff
@@ -0,0 +1,173 @@
...
+                if sys.platform == "win32":
+                    if not any("/std" in s for s in ext.extra_compile_args):
+                        ext.extra_compile_args.append("/std:c++17")
+                elif sys.platform == "linux" or sys.platform == "darwin":
+                    if "std=" not in os.environ.get("CXXFLAGS", ""):
+                        ext.extra_compile_args.append("-std=c++11")
+                        ext.extra_compile_args.append("-D_GLIBCXX_USE_CXX11_ABI=0")
+                    # Define endianness for flatcc
+                    ext.extra_compile_args.append("-DFLATBUFFERS_LITTLEENDIAN=1")
+                    ext.extra_compile_args.append("-DFLATBUFFERS_PROTOCOL_IS_LE=1")
```
sfc-gh-turbaszek: Similar comment, using extend may be cleaner

python/src/snowflake/ud_connector/connection.py
```diff
@@ -38,18 +39,36 @@ def __init__(self, **kwargs):
...
+                self.db_api.connection_set_option_double(
+                    ConnectionSetOptionDoubleRequest(
+                        conn_handle=self.conn_handle, key=key, value=value
+                    )
+                )
+
+        self.db_api.connection_init(
+            ConnectionInitRequest(
+                conn_handle=self.conn_handle, db_handle=self.db_handle
+            )
```
sfc-gh-turbaszek: Is this formatting-only change?

python/src/snowflake/ud_connector/connection.py
```diff
@@ -78,16 +97,20 @@ def rollback(self):
         """
         raise NotSupportedError("rollback is not implemented")
 
-    def cursor(self):
+    def cursor(self, use_dict_result=False, use_numpy=False):
```
sfc-gh-turbaszek: Please add type hints

python/src/snowflake/ud_connector/cursor.py
```diff
@@ -3,287 +3,295 @@
...
-    @rowcount.setter  
-    def rowcount(self, value):
-        self._rowcount = value
-    
+            int: Number of rows to fetch
+        """
+        return self._arraysize
+
+    @arraysize.setter
+    def arraysize(self, value) -> None:
```
sfc-gh-turbaszek: Should we set a type hint for `value`? What is expected here?

python/src/snowflake/ud_connector/cursor.py
```diff
@@ -3,287 +3,295 @@
...
 
     def fetchmany(self, size=None):
         """
         Fetch the next set of rows of a query result.
-        
+
         Args:
             size (int): Number of rows to fetch (defaults to arraysize)
-            
+
```
sfc-gh-turbaszek, line 173: There's a plenty of unrelated changes - should those be first applied on main?

sfc-gh-pczajka: Removed from this PR

python/src/snowflake/ud_connector/cursor.py
sfc-gh-pczajka: removed

python/src/snowflake/ud_connector/connection.py
sfc-gh-pczajka: Removed - out of scope

python/src/snowflake/ud_connector/connection.py
sfc-gh-pczajka: reverted for readability

python/setup.py
sfc-gh-pczajka: changed

python/setup.py
sfc-gh-pczajka: changed

python/src/snowflake/ud_connector/nanoarrow_cpp/ArrowIterator/CArrowStreamIterator.cpp
```diff
@@ -0,0 +1,272 @@
...
+  m_columnCount = m_schema->n_children;
+  logger->debug(__FILE__, __func__, __LINE__,
+                "CArrowStreamIterator initialized with %lld columns", m_columnCount);
+
+  m_initialized = true;
+}
+
+CArrowStreamIterator::~CArrowStreamIterator() {
+  // Release the stream if we own it
+  if (m_ownsStream && m_stream != nullptr && m_stream->release != nullptr) {
```
sfc-gh-jszczerbinski: Doesn't it always own stream? Why do we need this variable

python/src/snowflake/ud_connector/nanoarrow_cpp/ArrowIterator/CArrowStreamIterator.cpp
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  m_columnCount = m_schema->n_children;
+  logger->debug(__FILE__, __func__, __LINE__,
+                "CArrowStreamIterator initialized with %lld columns", m_columnCount);
+
+  m_initialized = true;
+}
+
+CArrowStreamIterator::~CArrowStreamIterator() {
```
sfc-gh-jszczerbinski: I think you are breaking rule of 5 here.

python/src/snowflake/ud_connector/nanoarrow_cpp/ArrowIterator/CArrowStreamIterator.cpp
```diff
@@ -0,0 +1,272 @@
...
+#include <string>
+#include <vector>
+
+namespace sf {
+
+Logger* CArrowStreamIterator::logger = new Logger("snowflake.connector.CArrowStreamIterator");
+
+CArrowStreamIterator::CArrowStreamIterator(int64_t stream_ptr, PyObject* context,
+                                           PyObject* use_numpy, PyObject* use_dict_result)
+    : m_stream(reinterpret_cast<ArrowArrayStream*>(stream_ptr)),
```
sfc-gh-jszczerbinski: Can it be unique_ptr with custom deleter? This way we could avoid writing a destructor

python/src/snowflake/ud_connector/nanoarrow_cpp/ArrowIterator/CArrowStreamIterator.cpp
```diff
@@ -0,0 +1,272 @@
...
+      m_ownsStream(true),
+      m_currentRowIndex(0),
+      m_rowCount(0),
+      m_columnCount(0),
+      m_context(context),
+      m_useNumpy(use_numpy == Py_True),
+      m_useDictResult(use_dict_result == Py_True),
+      m_initialized(false),
```
sfc-gh-jszczerbinski: Is this needed? Let's create quasi-constuctor - static member function that returns CArrowStreamIterator and does the validation and initialization then passes validated values to the constructor

sfc-gh-jszczerbinski (approved):

sfc-gh-jszczerbinski (changes requested):
Let's fix value semantics so we are sure that we are not leaking memory

python/src/snowflake/ud_connector/nanoarrow_cpp/ArrowIterator/CArrowStreamIterator.cpp
sfc-gh-pczajka: Removed

python/src/snowflake/ud_connector/nanoarrow_cpp/ArrowIterator/CArrowStreamIterator.cpp
sfc-gh-pczajka: Changes as discussed - now iterator holds unique pointer with custom deleter on the stream

python/src/snowflake/ud_connector/nanoarrow_cpp/ArrowIterator/CArrowStreamIterator.cpp
sfc-gh-pczajka: done

python/src/snowflake/ud_connector/nanoarrow_cpp/ArrowIterator/CArrowStreamIterator.cpp
sfc-gh-pczajka: nice! done

python/src/snowflake/ud_connector/_internal/nanoarrow_cpp/ArrowIterator/CArrowStreamIterator.cpp
```diff
@@ -6,57 +6,66 @@
...
+  if (stream == nullptr || stream->release == nullptr) {
     std::string errorInfo = "[Snowflake Exception] Invalid ArrowArrayStream pointer";
     logger->error(__FILE__, __func__, __LINE__, errorInfo.c_str());
     PyErr_SetString(PyExc_Exception, errorInfo.c_str());
-    return;
+    return nullptr;
   }
 
+  // Create the iterator (constructor only sets attributes)
+  auto iterator = std::unique_ptr<CArrowStreamIterator>(
```
sfc-gh-jszczerbinski: Can we move the constructor to the end (line 47) and pass schema to it, instead of setting it and m_columnCount from the outside?

sfc-gh-jszczerbinski (approved):

python/src/snowflake/ud_connector/_internal/nanoarrow_cpp/ArrowIterator/CArrowStreamIterator.cpp
sfc-gh-jszczerbinski: Also nit: use make_unique

sfc-gh-pczajka: done

python/hatch_build_nanoarrow.py
```diff
@@ -0,0 +1,178 @@
...
+            "nanoarrow.c",
+            "nanoarrow_ipc.c",
+        ]
+
+        # Add subdirectory sources
+        subdirectory_sources = [
+            ("Python", "Common.cpp"),
+            ("Python", "Helpers.cpp"),
+            ("Util", "time.cpp"),
+        ]
```
sfc-gh-turbaszek: Are those lists constant? If yes, would it make sens to move them to module or class level? sucha move would improve readability of _build_extensions. WDYT?

python/pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["hatchling"]
+requires = ["hatchling", "Cython", "setuptools"]
```
sfc-gh-turbaszek, line 2:
```suggestion
requires = ["hatchling", "Cython"]
```
setuptools should no longer be required

python/src/snowflake/ud_connector/_internal/arrow_context.py
sfc-gh-turbaszek: Lets remove the shebangs

python/tests/integ/test_cursor.py
```diff
@@ -105,9 +103,229 @@ def test_simple_select(self, cursor):
...
+    """Test dict result mode.
+
+    Note: DictCursor is not yet implemented. These tests use ArrowStreamIterator
+    directly to verify dict result functionality works correctly.
+    """
+
+    @pytest.mark.skip_reference
+    def test_next_returns_dict(self, cursor):
+        """Test next() returns dict with column names as keys."""
+        # TODO: Replace with DictCursor when implemented
```
sfc-gh-turbaszek, line 297: Why?

python/hatch_build_nanoarrow.py
```diff
@@ -0,0 +1,178 @@
+"""Custom Hatch build hook for building Cython extensions with nanoarrow C++ code."""
```
sfc-gh-turbaszek:
```suggestion
"""Custom Hatch build hooks."""
```

To be future proof. Also, let's change the file name to `hatch_build_hooks.py`

sfc-gh-pczajka: done

python/pyproject.toml
sfc-gh-pczajka: Unfortunately, this is the easiest way to get complicated logic around compiler flags - it is used by the hook:
```
    from Cython.Build import cythonize
    from setuptools import Distribution, Extension
    from setuptools.command.build_ext import build_ext
```

python/hatch_build_nanoarrow.py
sfc-gh-pczajka: Nice, moved to class

python/src/snowflake/ud_connector/_internal/arrow_context.py
sfc-gh-pczajka: Whoops, removed

python/tests/integ/test_cursor.py
sfc-gh-pczajka: This is a placeholder for DictCursor tests (which is out of scope of this PR) - two simple tests ensuring that the functionality works on Stream object. If we see value of testing stream itself they should be moved to another module; IMO writing these would be hard/hacky (you'd need to mock the stream or get them from cursor - like here), and the functionality is already well-covered in cursor tests

sfc-gh-turbaszek (approved):

PR #168, opened 2026-01-07
Initial compatibility with snowflake.core

sfc-gh-turbaszek:


python/src/snowflake/ud_connector/__init__.py
```diff
@@ -21,6 +24,12 @@
 threadsafety = 1  # Threads may share the module, but not connections
 paramstyle = "format"  # Python extended format codes, e.g. ...WHERE name=%s
 
+# Optional module aliasing: make this module available as snowflake.connector
+# Set SNOWFLAKE_ENABLE_CONNECTOR_ALIAS=1 (or true/yes) to enable
+_alias_enabled = os.environ.get("SNOWFLAKE_ENABLE_CONNECTOR_ALIAS", "").lower()
+if _alias_enabled in ("1", "true"):
+    sys.modules["snowflake.connector"] = sys.modules[__name__]
```
sfc-gh-fpawlowski: Open question - i recall finding some articles stating that changing manually things in sys.modules may not work as expected / cause issues later on.
I am not 100% certain here, just leaving it here not to forget about it.

sfc-gh-turbaszek: Definitely this is something that should not be used for anything but testing. I think I will move the logic from main to internal module. From my local tests it works as expected.

PR #174, opened 2026-01-09
Fix empty result handling when query returns zero rows

sfc-gh-pczajka:
When Snowflake returns zero rows, it sends an empty rowsetBase64 string. This made wrappers finish with an error.

This fix checks for empty decoded bytes and creates an empty Arrow reader with the schema from rowtype instead of attempting to parse empty IPC data.

### DIFF
--- a/sf_core/src/apis/database_driver_v1/query.rs
+++ b/sf_core/src/apis/database_driver_v1/query.rs
@@ -67,6 +67,11 @@ async fn read_batches(
     if let Some(rowset_base64) = &data.rowset_base64 {
         let rowset_bytes = BASE64.decode(rowset_base64).context(Base64DecodingSnafu)?;
 
+        // Handle empty result - return empty reader with schema from rowtype
+        if rowset_bytes.is_empty() {
+            return create_empty_reader_from_rowtype(data);
+        }
+
         let reader_result = if let Some(chunk_download_data) = data.to_chunk_download_data() {
             ChunkReader::multi_chunk(
                 rowset_bytes,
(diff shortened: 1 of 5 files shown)


tests/definitions/shared/query/empty_result.feature
```diff
@@ -0,0 +1,13 @@
+@odbc @python @jdbc
```
sfc-gh-jszczerbinski, line 1: Please add e2e test to core too

sf_core/src/apis/database_driver_v1/query.rs
```diff
@@ -188,6 +193,30 @@ fn build_generic_fixed_rowtype(name: &str) -> RowType {
     RowType::fixed_with_scale_zero(name, false, PUT_GET_ROWSET_FIXED_LENGTH)
 }
 
+/// Creates an empty Arrow reader with schema from rowtype metadata.
+/// Used when query returns 0 rows (empty rowsetBase64).
+fn create_empty_reader_from_rowtype(
+    data: &query_response::Data,
+) -> Result<Box<dyn RecordBatchReader + Send>, ReadBatchesError> {
+    if let Some(rowtype) = &data.row_type {
+        let row_types: Vec<RowType> = rowtype
```
sfc-gh-jszczerbinski, line 202: Let's add a guideline to include "empty rowset" test for each datatype in the doc, since right now `RowType` parsing is incomplete and we should make sure that it works for all types

sfc-gh-jszczerbinski (changes requested):
In general looks good to me, please add the missing test

PR #180, opened 2026-01-13
NO-SNOW: Use hatch for python wrapper

sfc-gh-turbaszek:


python/pyproject.toml
```diff
@@ -151,4 +150,68 @@ strict_equality = true
...
+# All tests
+all = "pytest tests/ --connector=universal {args}"
+# All tests with coverage
+all-cov = "pytest tests/ --connector=universal --cov=src/snowflake --cov-report=xml --cov-report=html {args}"
+
+# Reference connector environment (installs local package + reference connector)
+[tool.hatch.envs.reference]
+features = ["test"]
+dependencies = [
+    "snowflake-connector-python=={env:REFERENCE_DRIVER_VERSION:3.17.2}",
```
sfc-gh-pczajka, line 203: Why 3.17.2, not 4.2.0? Or just newest?

python/README.md
```diff
@@ -3,60 +3,94 @@
...
 ```
 
+You can verify the detected path with:
+```bash
+hatch run show-paths
+```
+
+**Note:** Unlike the old Makefile which auto-detected paths, you must explicitly set `PARAMETER_PATH` before running tests. If not set, it will be an empty string, which will cause test failures.
+
 ## Testing
```
sfc-gh-pczajka, line 52:
```suggestion
## Testing

See README in `tests/` directory
```

python/pyproject.toml
sfc-gh-turbaszek: Good point, this mimics the `REFERENCE_DRIVER_VERSION ?= 3.17.2` value. Imo this should be changed in separate PR to avoid unexpected test faliures caused by different version

sfc-gh-pczajka (approved):

sfc-gh-pczajka:
LGTM

PR #189, opened 2026-01-14
reenable clang

sfc-gh-pczajka:


sfc-gh-jszczerbinski (changes requested):
Are we sure we want to do that? Is there a world when we want to upgrade nanoarrow? If yes this will make the upgrade harder due more VC conflicts

sfc-gh-pczajka:
Good point, I've excluded nanoarrow directory from clang

sfc-gh-jszczerbinski (approved):

PR #195, opened 2026-01-16
int conversion support in Python

sfc-gh-pczajka:
Adds tests for int conversion (and fixes 1 bug)

Tests according to https://docs.google.com/document/d/1FCioVbI1rGEQSsDRJQE-I_ob_8IjpU9-hyIZzkDM42w/edit?tab=t.0, omiting generative ones for now.

**Stack:**
- #194 ← base PR (merged)
- #196 ← parent PR (merge first)
- This PR

### DIFF
--- a/python/src/snowflake/ud_connector/_internal/arrow_context.py
+++ b/python/src/snowflake/ud_connector/_internal/arrow_context.py
@@ -140,7 +140,7 @@ class ArrowConverterContext:
         nanoseconds = int(decimal.Decimal(epoch).scaleb(9) + decimal.Decimal(fraction))
         return numpy.datetime64(nanoseconds, "ns")
 
-    def DECIMAL128_to_decimal(self, int128_bytes: bytes, scale: int) -> decimal.Decimal | int:
+    def DECIMAL128_to_decimal_or_int(self, int128_bytes: bytes, scale: int) -> decimal.Decimal | int:
         """When scale=0 (integer), returns Python int. When scale>0 (decimal), returns Python Decimal.
 
         Large scale integers (int128) fall back here as they're not supported by native C++ conversion.
--- a/python/src/snowflake/ud_connector/_internal/arrow_context.py
+++ b/python/src/snowflake/ud_connector/_internal/arrow_context.py
@@ -7,7 +7,6 @@ and provides Python helper functions for Arrow conversions in the C++ converters
 from __future__ import annotations
 
 import decimal
-import time
 
 from datetime import datetime, timedelta, timezone, tzinfo
 from logging import getLogger
@@ -15,8 +14,6 @@ from sys import byteorder
 from typing import TYPE_CHECKING
 
 import pytz
-
-from pytz import UTC
 
 
 if TYPE_CHECKING:
@@ -143,10 +140,10 @@ class ArrowConverterContext:
         nanoseconds = int(decimal.Decimal(epoch).scaleb(9) + decimal.Decimal(fraction))
         return numpy.datetime64(nanoseconds, "ns")
 
-    def DECIMAL128_to_decimal(self, int128_bytes: bytes, scale: int) -> decimal.Decimal:
+    def DECIMAL128_to_decimal(self, int128_bytes: bytes, scale: int) -> decimal.Decimal | int:
         int128 = int.from_bytes(int128_bytes, byteorder=byteorder, signed=True)
         if scale == 0:
-            return decimal.Decimal(int128)
+            return int128
         digits = [int(digit) for digit in str(int128) if digit != "-"]
         sign = int128 < 0
         return decimal.Decimal((sign, digits, -scale))
--- a/python/src/snowflake/ud_connector/_internal/arrow_context.py
+++ b/python/src/snowflake/ud_connector/_internal/arrow_context.py
@@ -141,6 +141,10 @@ class ArrowConverterContext:
         return numpy.datetime64(nanoseconds, "ns")
 
     def DECIMAL128_to_decimal(self, int128_bytes: bytes, scale: int) -> decimal.Decimal | int:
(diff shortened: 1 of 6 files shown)


sfc-gh-jszczerbinski (approved):
Gherkins look good, we need python team member to review too

python/src/snowflake/ud_connector/_internal/arrow_context.py
```diff
@@ -7,7 +7,6 @@
 from __future__ import annotations
 
 import decimal
-import time
```
sfc-gh-fpawlowski, line 10: Was this import not used? If yes, why it was not detected in our precommit? Can we make PR to change that?

python/src/snowflake/ud_connector/_internal/arrow_context.py
```diff
@@ -143,10 +140,10 @@ def TIMESTAMP_NTZ_TWO_FIELD_to_numpy_datetime64(self, epoch: int, fraction: int)
         nanoseconds = int(decimal.Decimal(epoch).scaleb(9) + decimal.Decimal(fraction))
         return numpy.datetime64(nanoseconds, "ns")
 
-    def DECIMAL128_to_decimal(self, int128_bytes: bytes, scale: int) -> decimal.Decimal:
+    def DECIMAL128_to_decimal(self, int128_bytes: bytes, scale: int) -> decimal.Decimal | int:
```
sfc-gh-fpawlowski: I see that this code snippet was the same way in our original code, but are we are of a reason why method ...to_decimal does not return decimal (apart from the fact that scale is 0)?

Is this convention across python, or was this decision in Snowflake Python driver early development? Maybe worth leaving a comment in code to avoid future confusion.

python/tests/e2e/types/test_int.py
```diff
@@ -0,0 +1,360 @@
...
+# https://docs.snowflake.com/en/sql-reference/data-types-numeric
+INT_TYPE_SYNONYMS = [
+    "INT",
+    "INTEGER",
+    "BIGINT",
+    "SMALLINT",
+    "TINYINT",
+    "BYTEINT",
+]
+int_type_parametrize = pytest.mark.parametrize("int_type", INT_TYPE_SYNONYMS)
```
sfc-gh-fpawlowski, line 27: Simple but cool idea!

python/tests/e2e/types/test_int.py
```diff
@@ -0,0 +1,360 @@
...
+    @pytest.mark.skip("SNOW-3006013 - parameter binding is not yet implemented")
+    @int_type_parametrize
+    def test_should_insert_and_select_integers_from_table_using_parameter_binding_for_int_and_synonyms(
+        self, execute_query, tmp_schema, int_type
+    ):
+        # Given Snowflake client is logged in
+
+        # And Table with <type> column exists
+        table_name = f"{tmp_schema}.int_bind_table_{int_type.lower()}"
+        execute_query(f"CREATE TABLE {table_name} (col {int_type})")
```
sfc-gh-fpawlowski, line 347: nit: We have quite some create table calls like
`execute_query(f"CREATE TABLE {table_name} (col {int_type})")`.
Do we want to extract some common helper to create table? Given args like name, mapping of col names to col types etc? Similar could be done for inserts probably.

Side note: For more fancy approach we could introduce the whole TableTesting class that would allow us to call .create, .insert etc on it, but we don't have resources to reinvent snowpark here so just leaving it as a loose observation : )

python/tests/unit/test_arrow_context.py
```diff
@@ -315,12 +321,17 @@ class TestDecimal128ToDecimal:
             "max_38_digit",
         ],
     )
-    def test_decimal128_to_decimal(self, int_value, scale, expected):
-        """Test DECIMAL128 conversion with various values and scales."""
+    def test_decimal128_to_decimal(self, int_value, scale, expected, expected_type):
+        """Test DECIMAL128 conversion with various values and scales.
+
+        When scale=0 (integer), returns Python int to match old driver behavior.
```
sfc-gh-fpawlowski: Here we might have answer for my previous comment, but still would be nice to have it next to the actual implementation.

sfc-gh-turbaszek (approved):

python/tests/e2e/types/test_int.py
sfc-gh-pczajka: IMO this is a good moment to cut helpers, CREATE do not take too much "space" and are easy to understand (you don't need to look up how to use another helper class :) )

python/src/snowflake/ud_connector/_internal/arrow_context.py
sfc-gh-pczajka: The actual reason is NUMBER(any, 0) are synonymous to integers, which should be treated as integers.
However, Snowflake does not use all 128 bit space if it does not need to (if column needs only 32bit, it returns 32bit). Up to 64 are natively converted in C++, 128 falls back to Python.
I agree that naming could be better, however that would trigger a bigger refactor of a code we are not likely to revisit during UD development - IMO not worth it for now.

I've added comments trying to reflect the actual reason

sfc-gh-pczajka: ...and slightly rename the function

python/tests/unit/test_arrow_context.py
sfc-gh-pczajka: This one is misleading - fixed

python/tests/e2e/types/test_int.py
sfc-gh-fpawlowski: Convinced : )

sfc-gh-fpawlowski (approved):
Looks good to me!
I would just like to have a jira or PR to fix missing ruff rules for unused imports.

PR #203, opened 2026-01-20
python - add support for DECFLOAT

sfc-gh-pczajka:
Add tests for decfloat type according to https://docs.google.com/document/d/1FCioVbI1rGEQSsDRJQE-I_ob_8IjpU9-hyIZzkDM42w/edit?tab=t.0.
Tests are explicitly separated to float range and extreme values - wrapper which does not support them will have place to document its behavior
Base PR: #195

### DIFF
--- a/python/tests/e2e/types/conftest.py
+++ b/python/tests/e2e/types/conftest.py
@@ -1,0 +1,28 @@
+"""Shared fixtures for type tests."""
+
+from __future__ import annotations
+
+from decimal import getcontext
+
+import pytest
+
+
+# =============================================================================
+# DECIMAL CONTEXT CONFIGURATION
+# =============================================================================
+# DECFLOAT and NUMBER types require Python's Decimal context to have 38-digit
+# precision to properly represent all values without rounding
+DECIMAL_PRECISION_38 = 38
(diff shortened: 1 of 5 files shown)


python/tests/e2e/types/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+def setup_decimal_precision():
+    """Set decimal context precision to 38 for all type tests.
+
+    This ensures Decimal operations don't lose precision for DECFLOAT
+    and NUMBER types which support up to 38 significant digits.
+    """
+    old_prec = getcontext().prec
+    getcontext().prec = DECIMAL_PRECISION_38
+    yield
+    getcontext().prec = old_prec
```
sfc-gh-turbaszek: Should there be try/except?

sfc-gh-turbaszek (approved):

python/tests/e2e/types/conftest.py
sfc-gh-pczajka: Nice spot, fixed!

sfc-gh-jszczerbinski (approved):

PR #207, opened 2026-01-21
python - logical types support

sfc-gh-pczajka:
Add tests for BOOLEAN convertion according to [test plan](https://docs.google.com/document/d/1FCioVbI1rGEQSsDRJQE-I_ob_8IjpU9-hyIZzkDM42w/edit?tab=t.0.) and [documentation](https://docs.snowflake.com/en/sql-reference/data-types-logical)

### DIFF
--- a/python/tests/e2e/types/test_boolean.py
+++ b/python/tests/e2e/types/test_boolean.py
@@ -10,8 +10,7 @@ from .utils import assert_type
 # =============================================================================
 # LARGE RESULT SET SIZE
 # =============================================================================
-LARGE_RESULT_SET_SIZE_GENERATOR = 1_000_000
-LARGE_RESULT_SET_SIZE_TABLE = 1_000_000
+LARGE_RESULT_SET_SIZE = 1_000_000
 
 
 class TestBooleanTypeCasting:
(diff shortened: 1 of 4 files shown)


tests/definitions/shared/types/boolean.feature
```diff
@@ -0,0 +1,81 @@
...
+  #                             Table operations                                #
+  # =========================================================================== #
+
+  @python_e2e
+  Scenario: should select boolean values from table
+    Given Snowflake client is logged in
+    And Table with columns (BOOLEAN, BOOLEAN, BOOLEAN) exists
+    And Row (TRUE, FALSE, TRUE) is inserted
+    When Query "SELECT * FROM <table>" is executed
+    Then Result should contain [TRUE, FALSE, TRUE]
```
sfc-gh-turbaszek, line 47:
```suggestion
    Then Result should contain [TRUE, FALSE, TRUE] in any order
```
To be consistent with next scenario or let's add `ORDER BY 1`

python/tests/e2e/types/test_boolean.py
```diff
@@ -0,0 +1,177 @@
...
+
+import pytest
+
+from .utils import assert_sequential_values, assert_type
+
+
+# =============================================================================
+# LARGE RESULT SET SIZE
+# =============================================================================
+LARGE_RESULT_SET_SIZE = 1_000_000
```
sfc-gh-turbaszek, line 13: Should we consider moving to a global constant instead of defining this per type?

tests/definitions/shared/types/boolean.feature
sfc-gh-jszczerbinski: We insert 3 columns so i assume the order is deterministic here

sfc-gh-jszczerbinski (approved):
Gherkins look good to me

tests/definitions/shared/types/boolean.feature
sfc-gh-pczajka: This is a single row, should be fine :)

python/tests/e2e/types/test_boolean.py
sfc-gh-pczajka: I've changed that in #232 to make the tests faster - now it differs between types

sfc-gh-turbaszek (approved):

PR #208, opened 2026-01-21
Add arrow converters to JDBC

sfc-gh-rkowalski:
Arrow converter layer is now integrated into `universal-driver/jdbc` (Arrow-only, no JSON/common). Migrated converters with their unit tests include:
- VarCharConverter
- VarBinaryToBinaryConverter
- BitToBooleanConverter
- DoubleToRealConverter
- TinyIntToFixedConverter + TinyIntToScaledFixedConverter
- SmallIntToFixedConverter + SmallIntToScaledFixedConverter
- IntToFixedConverter + IntToScaledFixedConverter
- BigIntToFixedConverter + BigIntToScaledFixedConverter
- DecimalToScaledFixedConverter
- ArrowVectorConverterUtil and minimal ArrowResultUtil
(no time, date and timestamp)

Other:
- Added integration tests in `SnowflakeResultSetGettersTest` to validate `getInt, getFloat, getDouble, getString, getObject, getBytes, getBigDecimal`, and float special values (inf, -inf, nan).
- `SnowflakeResultSet` now routes getters through Arrow converters (still using row index 0, no batch/row rework).
- Arrow schema metadata is read to populate columnNames/columnTypes using logicalType metadata.

### DIFF
--- a/jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
+++ b/jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
@@ -27,7 +27,6 @@ import java.sql.Time;
 import java.sql.Timestamp;
 import java.sql.Types;
 import java.util.Calendar;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import net.snowflake.client.internal.core.arrow.ArrowVectorConverter;
@@ -63,7 +62,7 @@ public class SnowflakeResultSet implements ResultSet {
   private ArrowArrayStream stream;
   private ArrowReader reader;
   private BufferAllocator allocator;
-  private final Map<Integer, ArrowVectorConverter> converterCache = new HashMap<>();
+  private ArrowVectorConverter[] converterCache;
   private static final DataConversionContext EMPTY_CONTEXT = new DataConversionContext() {};
 
   public SnowflakeResultSet(SnowflakeStatement statement, ExecuteResult result) {
@@ -1233,14 +1232,20 @@ public class SnowflakeResultSet implements ResultSet {
   }
 
   private ArrowVectorConverter getConverter(int columnIndex) throws SQLException {
-    if (converterCache.containsKey(columnIndex)) {
-      return converterCache.get(columnIndex);
+    ensureSchemaInitialized();
+    int index = columnIndex - 1;
+    if (index < 0 || index >= converterCache.length) {
+      throw new SQLException("Invalid column index: " + columnIndex);
+    }
+    ArrowVectorConverter cached = converterCache[index];
+    if (cached != null) {
+      return cached;
     }
     try {
-      FieldVector vector = reader.getVectorSchemaRoot().getVector(columnIndex - 1);
+      FieldVector vector = reader.getVectorSchemaRoot().getVector(index);
       ArrowVectorConverter converter =
-          ArrowVectorConverterUtil.initConverter(vector, EMPTY_CONTEXT, columnIndex - 1);
-      converterCache.put(columnIndex, converter);
+          ArrowVectorConverterUtil.initConverter(vector, EMPTY_CONTEXT, index);
+      converterCache[index] = converter;
       return converter;
     } catch (SnowflakeSQLException e) {
       throw new SQLException("Unable to create converter for column " + columnIndex, e);
@@ -1250,7 +1255,7 @@ public class SnowflakeResultSet implements ResultSet {
   }
 
   private void ensureSchemaInitialized() throws SQLException {
-    if (columnNames != null && columnTypes != null) {
+    if (columnNames != null && columnTypes != null && converterCache != null) {
       return;
     }
     List<Field> fields;
@@ -1261,6 +1266,7 @@ public class SnowflakeResultSet implements ResultSet {
     }
     columnNames = new String[fields.size()];
     columnTypes = new int[fields.size()];
+    converterCache = new ArrowVectorConverter[fields.size()];
     for (int i = 0; i < fields.size(); i++) {
       Field field = fields.get(i);
       columnNames[i] = field.getName();
--- a/jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
+++ b/jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
@@ -27,13 +27,22 @@ import java.sql.Time;
 import java.sql.Timestamp;
 import java.sql.Types;
 import java.util.Calendar;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import net.snowflake.client.internal.core.arrow.ArrowVectorConverter;
+import net.snowflake.client.internal.core.arrow.ArrowVectorConverterUtil;
+import net.snowflake.client.internal.core.arrow.DataConversionContext;
+import net.snowflake.client.jdbc.SFException;
+import net.snowflake.client.jdbc.SnowflakeSQLException;
+import net.snowflake.client.jdbc.SnowflakeType;
 import org.apache.arrow.c.ArrowArrayStream;
 import org.apache.arrow.c.Data;
 import org.apache.arrow.memory.BufferAllocator;
 import org.apache.arrow.memory.RootAllocator;
-import org.apache.arrow.vector.TinyIntVector;
+import org.apache.arrow.vector.FieldVector;
 import org.apache.arrow.vector.ipc.ArrowReader;
+import org.apache.arrow.vector.types.pojo.Field;
 
 /**
  * Snowflake JDBC ResultSet implementation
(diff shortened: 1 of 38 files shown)


jdbc/src/main/java/net/snowflake/client/internal/core/arrow/DoubleToRealConverter.java
```diff
@@ -0,0 +1,120 @@
...
+    return isNull(index) ? null : String.valueOf(toDouble(index));
+  }
+
+  @Override
+  public boolean toBoolean(int index) throws SFException {
+    if (isNull(index)) {
+      return false;
+    }
+    double val = toDouble(index);
+    throw new SFException(
```
sfc-gh-pfus, line 61: Is it correct?

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
```diff
@@ -426,7 +443,7 @@ public boolean isBeforeFirst() throws SQLException {
   @Override
   public boolean isAfterLast() throws SQLException {
     checkClosed();
-    return currentRow >= data.length;
+    return currentRow > 0;
```
sfc-gh-pfus, line 445: Can you explain?

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
```diff
@@ -1201,17 +1218,79 @@ private void checkRowPosition() throws SQLException {
     if (currentRow < 0) {
       throw new SQLException("Before first row");
     }
-    if (currentRow >= data.length) {
+    if (columnNames != null && currentRow >= 1) {
```
sfc-gh-pfus, line 1220: Why do we need column names?

jdbc/src/test/java/com/snowflake/jdbc/SnowflakeDriverTest.java
```diff
@@ -19,8 +20,8 @@ public class SnowflakeDriverTest {
   public void testDriverRegistration() throws SQLException {
     // Test that the driver is properly registered
     Driver driver = DriverManager.getDriver("jdbc:snowflake://test.snowflakecomputing.com");
-    assertNotNull("Driver should be registered", driver);
-    assertTrue("Driver should be instance of SnowflakeDriver", driver instanceof SnowflakeDriver);
+    assertNotNull(driver, "Driver should be registered");
```
sfc-gh-pfus, line 23: Maybe we should include a better assertion library (Fest/Hamcrest), to not bump into such problems and better expressiveness in general?

jdbc/src/test/java/com/snowflake/jdbc/SnowflakeResultSetGettersTest.java
```diff
@@ -0,0 +1,169 @@
...
+import java.sql.ResultSet;
+import java.sql.Statement;
+import java.util.Properties;
+import org.json.JSONObject;
+import org.json.JSONTokener;
+import org.junit.jupiter.api.Test;
+
+public class SnowflakeResultSetGettersTest {
+
+  private Properties loadConnectionProperties() throws Exception {
```
sfc-gh-pfus, line 22: It seems it needs some external class to handle this?

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/DoubleToRealConverter.java
sfc-gh-rkowalski: Thats how its in `snowflake-jdbc` but I agree that it looks sus. I think the intent is to always throw exception when converting to boolean

jdbc/src/test/java/com/snowflake/jdbc/SnowflakeResultSetGettersTest.java
sfc-gh-rkowalski: Yes, I have that in next PR

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-rkowalski: Because I removed the data[][] backing array but still didn’t implement real row/batch indexing. This will be done in subsequent PRs

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-rkowalski: Without that, currentRow >= 1 could trigger before any batch is loaded. This will be improved soon when I implement real batch processing

jdbc/src/test/java/com/snowflake/jdbc/SnowflakeDriverTest.java
sfc-gh-rkowalski: I don't have strong opinion about it. I feel like for most tests JUnit assertions are enough but I'm open to introducing this in the future. This change was dictated by upgrading from junit 4 to 5 in this project

sfc-gh-rkowalski:
* **#212** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/212?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#211** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/211?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#210** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/210?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#208** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/208?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/208?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
```diff
@@ -1201,17 +1218,79 @@ private void checkRowPosition() throws SQLException {
...
+    if (columnNames == null) {
+      ensureSchemaInitialized();
+    }
     if (columnIndex < 1 || columnIndex > columnNames.length) {
       throw new SQLException("Invalid column index: " + columnIndex);
     }
   }
 
+  private ArrowVectorConverter getConverter(int columnIndex) throws SQLException {
+    if (converterCache.containsKey(columnIndex)) {
```
sfc-gh-jszczerbinski: Maybe this is a nit:  
Can we use Converter[] instead Map here? I know those are both O(1), but the constant is much higher for Map. I think it matters since it is a hot path

sfc-gh-jszczerbinski (approved):
Looks good to me. Small performance nitpick

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-rkowalski: Good idea, changed

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/DoubleToRealConverter.java
sfc-gh-pfus: So what's the point of converting to double before?

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-pfus: It's worth having TODOs then.

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-pfus: Maybe it could be optimized even further, by reusing converter for all columns with the same type? Maybe the cache key should be a column type?

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/DoubleToRealConverter.java
sfc-gh-rkowalski: I guess the intent was to have the actual value in the exception

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-rkowalski: Currently converters are per-column adapters, not stateless per-type helpers. Reusing one would read from the wrong buffers

PR #210, opened 2026-01-22
Extract common IT base

sfc-gh-rkowalski:

PR #211, opened 2026-01-22
Arrow-backed ResultSet.next() implementation

sfc-gh-rkowalski:
- Implement cursor-based `ResultSet.next()` using Arrow batch management, with explicit cursor/batch/schema state and correct `wasNull`, `getRow`, and `isLast` semantics.
- Add Arrow resource lifecycle handling for batch transitions and safe cleanup on close.
- Move Arrow converter classes into `...arrow.converters` package.
- Add tests to validate new behavior.


sfc-gh-rkowalski:
* **#212** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/212?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#211** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/211?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/211?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#210** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/210?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#208** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/208?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
```diff
@@ -481,6 +564,9 @@ public boolean last() throws SQLException {
   @Override
   public int getRow() throws SQLException {
     checkClosed();
+    if (currentRow < 0 || afterLast) {
+      return 0;
```
sfc-gh-rkowalski, line 392: Should be 1

jdbc/src/main/java/net/snowflake/client/jdbc/ErrorCode.java
sfc-gh-merbel, line 3: Are we documenting anywhere the error codes that we're returning?

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/IntToFixedConverter.java
sfc-gh-merbel, line 119: We have a duplicates in that package for this method implementation, maybe we can extract it?

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
```diff
@@ -51,52 +48,74 @@
...
-    checkClosed();
-    try {
-      reader.loadNextBatch();
-      currentRow = 0;
-      ensureSchemaInitialized();
-    } catch (IOException e) {
-      throw new RuntimeException(e);
+    if (closed) {
+      return false;
+    }
```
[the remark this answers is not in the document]
sfc-gh-rkowalski: We keep that behavior for backwards compatibility

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
```diff
@@ -51,52 +48,74 @@
...
+    boolean hasNext = fetchNextRow();
+    if (!hasNext) {
+      cursor.setAfterLast();
+      cursor.setOnLastRow(false);
```
sfc-gh-pfus: Is it correct?

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
```diff
@@ -116,45 +136,62 @@ public String getString(int columnIndex) throws SQLException {
 
   @Override
   public boolean getBoolean(int columnIndex) throws SQLException {
-    String value = getString(columnIndex);
-    if (wasNull) {
-      return false;
+    checkClosed();
```
sfc-gh-pfus: That's a lot of boilerplate repeated in every function 🙃

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
```diff
@@ -204,15 +252,19 @@ public BigDecimal getBigDecimal(int columnIndex, int scale) throws SQLException
     if (value == null) {
       return null;
     }
-    return value.setScale(scale, BigDecimal.ROUND_HALF_UP);
+    return value.setScale(scale, RoundingMode.HALF_UP);
```
sfc-gh-pfus, line 153: Why this change happened?

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
```diff
@@ -221,41 +273,17 @@ public byte[] getBytes(int columnIndex) throws SQLException {
...
-    String value = getString(columnIndex);
-    if (wasNull) {
-      return null;
-    }
-    try {
-      return Date.valueOf(value);
-    } catch (IllegalArgumentException e) {
-      throw new SQLException("Cannot convert '" + value + "' to Date", e);
-    }
+    throw new SQLFeatureNotSupportedException("getDate not supported");
```
sfc-gh-pfus: Why? Next PR? TODO?

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
```diff
@@ -1206,95 +1253,67 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
-      case REAL:
-        return Types.DOUBLE;
-      case BOOLEAN:
-        return Types.BOOLEAN;
-      case BINARY:
-        return Types.BINARY;
-      default:
-        return Types.OTHER;
-    }
+  private interface CloseAction {
```
sfc-gh-pfus: Maybe `java.io.Closeable` is enough?

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/ArrowBatchManager.java
```diff
@@ -0,0 +1,90 @@
...
+    } catch (IOException e) {
+      throw new SQLException("Unable to advance Arrow results", e);
+    }
+  }
+
+  public void prefetchNextBatchForIsLast() throws SQLException {
+    if (batch.hasPrefetchedBatch() || cursor.isAfterLast()) {
+      return;
+    }
+    VectorSchemaRoot snapshot = resources.copyVectorSchemaRoot(resources.getActiveRoot());
```
sfc-gh-pfus: Are you sure about this? This seems heavy.

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/ArrowBatchManager.java
```diff
@@ -0,0 +1,90 @@
...
+        batch.incrementRowInBatch();
+        return true;
+      }
+      return moveToNextBatch();
+    } catch (IOException e) {
+      throw new SQLException("Unable to advance Arrow results", e);
+    }
+  }
+
+  public void prefetchNextBatchForIsLast() throws SQLException {
```
sfc-gh-pfus: Can you add a comment why this is needed?

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/ArrowResources.java
```diff
@@ -0,0 +1,127 @@
...
+import java.sql.SQLException;
+import java.util.List;
+import org.apache.arrow.c.ArrowArrayStream;
+import org.apache.arrow.memory.BufferAllocator;
+import org.apache.arrow.vector.FieldVector;
+import org.apache.arrow.vector.VectorSchemaRoot;
+import org.apache.arrow.vector.ipc.ArrowReader;
+
+public final class ArrowResources {
+  private VectorSchemaRoot currentRoot;
```
sfc-gh-pfus: Finals?

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/ArrowResources.java
```diff
@@ -0,0 +1,127 @@
...
+    VectorSchemaRoot copy = VectorSchemaRoot.create(source.getSchema(), allocator);
+    int rowCount = source.getRowCount();
+    copy.setRowCount(rowCount);
+    List<FieldVector> sourceVectors = source.getFieldVectors();
+    List<FieldVector> targetVectors = copy.getFieldVectors();
+    for (int vectorIndex = 0; vectorIndex < sourceVectors.size(); vectorIndex++) {
+      FieldVector sourceVector = sourceVectors.get(vectorIndex);
+      FieldVector targetVector = targetVectors.get(vectorIndex);
+      targetVector.allocateNew();
+      for (int rowIndex = 0; rowIndex < rowCount; rowIndex++) {
```
sfc-gh-pfus: That doesn't seem right even more... Do I get it correctly, that we copy all arrow cells one by one?

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/BatchState.java
```diff
@@ -0,0 +1,55 @@
...
+  void setCurrentBatchRowCount(int currentBatchRowCount) {
+    this.currentBatchRowCount = currentBatchRowCount;
+  }
+
+  boolean hasLoadedBatch() {
+    return hasLoadedBatch;
+  }
+
+  void resetHasLoadedBatch() {
+    this.hasLoadedBatch = true;
```
sfc-gh-pfus: The initial value is false :) Should we manage all this functions separately? It seems that we should keep the state consistent, probably it's worth setting everything in one call.

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/SchemaState.java
```diff
@@ -0,0 +1,110 @@
...
+
+public final class SchemaState {
+  private static final DataConversionContext EMPTY_CONTEXT = new DataConversionContext() {};
+
+  private String[] columnNames;
+  private int[] columnTypes;
+  private ArrowVectorConverter[] converterCache;
+
+  void ensureInitialized(VectorSchemaRoot root) throws SQLException {
+    if (columnNames != null && columnTypes != null && converterCache != null) {
```
sfc-gh-pfus: Can't we just initialize in constructor and keep a correct state always?

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/SchemaState.java
```diff
@@ -0,0 +1,110 @@
...
+    converterCache = null;
+    columnNames = null;
+    columnTypes = null;
+  }
+
+  private int mapLogicalTypeToSqlType(SnowflakeType logicalType) {
+    if (logicalType == null) {
+      return Types.OTHER;
+    }
+    switch (logicalType) {
```
sfc-gh-pfus, line 91: How about other types? Are we supposed to focus only on basic ones?

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-rkowalski: Yes, more complex types are not supported yet. I'd better have it throw exception instead of working in too naive way like it was before. I will add TODOs

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-rkowalski: `setScale(scale, BigDecimal.ROUND_HALF_UP)` is deprecated now. `RoundingMode` should be used instead

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/ArrowBatchManager.java
sfc-gh-rkowalski: Yes, it adds overhead. The copy is used to snapshot the current batch so the code can peek at the next batch (to answer `isLast()`) without losing the current batch’s data. It reuses the same `VectorSchemaRoot` and underlying buffers for each batch, so advancing would overwrite the current vectors.
But you are right to mark this. I'll double check if that is correct and cannot be mitigated without additional overhead.

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-rkowalski: `onLastRow` here means “the cursor is positioned on the last valid row.” If `hasNext` is false, next() is moving the cursor past the end of the result set, so there is no current row at all

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-rkowalski: I agree, unfortunately we do have to check these things before each call. Improving this is on my todo list

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/SchemaState.java
sfc-gh-rkowalski: For now yes. I'll add TODO here

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/ArrowResources.java
sfc-gh-rkowalski: Yes, the safest option was to do a deep-copy here. Potentially we could share the buffers and get rid of copying. I'll take a second look at that.

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/ArrowBatchManager.java
sfc-gh-rkowalski: I was able to change this from deep-copy to zero-copy transfer still keeping the same functionality

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/ArrowBatchManager.java
sfc-gh-rkowalski: Added docs

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/ArrowResources.java
sfc-gh-rkowalski: This cannot be final since I'm mutating currentRoot when moving to next batch. I could create new ArrowResources each time but I wonder if it wouldn't add much overhead.

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/BatchState.java
sfc-gh-rkowalski: I've refactored how resetting works

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/SchemaState.java
sfc-gh-rkowalski: Got rid of `ensureInitialized` and moved this to constructor

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/ArrowBatchManager.java
sfc-gh-rkowalski: I decided to remove this code and not implement `isLast()` yet. I think we will be able to implement it with metadata from backend but its not ready in core yet. Added todo for it.

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-rkowalski: I decided to clean it up in this PR ;)

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
```diff
@@ -51,151 +48,101 @@
...
-  private ArrowReader reader;
-  private BufferAllocator allocator;
-  private ArrowVectorConverter[] converterCache;
-  private static final DataConversionContext EMPTY_CONTEXT = new DataConversionContext() {};
-
-  public SnowflakeResultSet(SnowflakeStatement statement, ExecuteResult result) {
+  public SnowflakeResultSet(SnowflakeStatement statement, ExecuteResult result)
+      throws SQLException {
     this.statement = statement;
     ByteString stream_ptr_bytes = result.getStream().getValue();
```
sfc-gh-pfus: Can we format this var name in Java style, not Rust? :D

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
```diff
@@ -51,151 +48,101 @@
...
     long ptr =
         ByteBuffer.wrap(stream_ptr_bytes.toByteArray()).order(ByteOrder.LITTLE_ENDIAN).getLong();
```
sfc-gh-pfus: I'm thinking, will it work on AIX? If this is a long (pointer address), probably on big endian system it will be reversed?

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
```diff
@@ -51,151 +48,101 @@
...
-    this.stream = ArrowArrayStream.wrap(ptr);
-    this.allocator = new RootAllocator(Long.MAX_VALUE);
-    this.reader = Data.importArrayStream(this.allocator, this.stream);
+    ArrowArrayStream stream = ArrowArrayStream.wrap(ptr);
+    RootAllocator allocator = new RootAllocator(Long.MAX_VALUE);
```
sfc-gh-pfus: Default constructor already uses this value. Do we want to keep it or maybe delegate the default to Arrow? 

The other discussion is - is it safe to allocate without limits. Probably yes, but...

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-pfus, line 1: Probably this is not a good PR to do this, but it seems like we should think about our code organization.

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
```diff
@@ -51,151 +48,101 @@
...
+    cursor.incrementRow();
     return true;
   }
 
   @Override
   public void close() throws SQLException {
+    if (closed) {
+      return;
+    }
     closed = true;
```
sfc-gh-pfus: I think we should do this after actually closing. Just in case if first close fails, but the second succeeds.

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
```diff
@@ -436,25 +341,25 @@ public BigDecimal getBigDecimal(String columnLabel) throws SQLException {
   @Override
   public boolean isBeforeFirst() throws SQLException {
     checkClosed();
-    return currentRow == -1;
+    return cursor.getCurrentRow() < 0 && !cursor.isAfterLast();
```
sfc-gh-pfus, line 346: Any comment why the second condition is needed? For empty resultset? Worth adding to the code.

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
```diff
@@ -436,25 +341,25 @@ public BigDecimal getBigDecimal(String columnLabel) throws SQLException {
...
     checkClosed();
-    return currentRow > 0;
+    return cursor.isAfterLast();
   }
 
   @Override
   public boolean isFirst() throws SQLException {
     checkClosed();
-    return currentRow == 0;
+    return cursor.getCurrentRow() == 0 && !cursor.isAfterLast();
```
sfc-gh-pfus, line 359: Same here - why the second part is needed? Current row is not incremented when we call next and there is no next row? If yes, this seems to be tricky, I'd expect that currentRow is null/-1 or sth when we are after all rows.

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/BatchState.java
```diff
@@ -0,0 +1,36 @@
+package net.snowflake.client.internal.core.arrow.cursor;
+
+public final class BatchState {
```
sfc-gh-pfus: Maybe it can be a part of CursorState?

jdbc/src/test/java/com/snowflake/jdbc/SnowflakeIntegrationTestBase.java
```diff
@@ -51,6 +52,20 @@ protected String buildJdbcUrl(Properties props) {
     return props.getProperty("url", defaultUrl);
   }
 
+  protected void ensureDatabaseAndSchema(Connection conn) throws Exception {
+    Properties props = loadConnectionProperties();
+    String database = props.getProperty("db");
+    String schema = props.getProperty("schema");
+    try (Statement stmt = conn.createStatement()) {
+      if (database != null && !database.isEmpty()) {
+        stmt.execute("use database " + database);
```
sfc-gh-pfus, line 61: Just curious - why can't we do it during connection init? It's slow and it makes our connections dirty (if we reuse them).

jdbc/src/test/java/com/snowflake/jdbc/SnowflakeResultSetCursorTest.java
```diff
@@ -0,0 +1,140 @@
...
+public class SnowflakeResultSetCursorTest extends SnowflakeIntegrationTestBase {
+
+  @Test
+  public void testCursorPosition() throws Exception {
+    String tableName = "RS_CURSOR_TEST_" + System.currentTimeMillis();
+    try (Connection conn = openConnection();
+        Statement stmt = conn.createStatement()) {
+      ensureDatabaseAndSchema(conn);
+      try {
+        stmt.execute("create or replace temporary table " + tableName + " (id int)");
```
sfc-gh-pfus: Why do we need a table? Can't we just use a static select?

jdbc/src/test/java/com/snowflake/jdbc/SnowflakeResultSetCursorTest.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  @Test
+  public void testUnsupportedGettersThrow() throws Exception {
+    try (Connection conn = openConnection();
+        Statement stmt = conn.createStatement()) {
+      ensureDatabaseAndSchema(conn);
+      try (ResultSet rs = stmt.executeQuery("select 1")) {
+        assertTrue(rs.next());
+        assertThrows(SQLFeatureNotSupportedException.class, () -> rs.getDate(1));
```
sfc-gh-pfus, line 113: I like this 👍

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-rkowalski: Could be. This is something worth checking later. I'll add TODO right away

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-rkowalski: Done

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-rkowalski: I agree. I should have done this earlier but I plan to replicate exact structure from `snowflake-jdbc` with proper API separation

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-rkowalski: Exactly. After calling `next()` on an empty result `currentRow` stays -1, but `afterLast` becomes true
Without the `!isAfterLast()` check, `isBeforeFirst()` would incorrectly return true for an exhausted empty result set. I'll add comment

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-rkowalski: `currentRow` is only incremented via `incrementRow()` when `next()` succeeds, so in practice `currentRow staying` at 0 after exhaustion might not happen, so `!isAfterLast()` check is defensive for this use cases

sfc-gh-rkowalski: I've added comment too

jdbc/src/test/java/com/snowflake/jdbc/SnowflakeIntegrationTestBase.java
sfc-gh-rkowalski: We can and we should. I have a TODO for providing more robust IT infrastructure so we can reuse same connection if possible

jdbc/src/main/java/net/snowflake/client/internal/core/arrow/cursor/BatchState.java
sfc-gh-rkowalski: Sure, why not. They have similar purpose. I'll merge them into one

jdbc/src/test/java/com/snowflake/jdbc/SnowflakeResultSetCursorTest.java
sfc-gh-rkowalski: Right, good catch!

jdbc/src/main/java/com/snowflake/jdbc/SnowflakeResultSet.java
sfc-gh-rkowalski: I've switched to default constructor for simplicity. Long.MAX_VALUE is existing in `snowflake-jdbc` from the dawn of time so I think its safe enough 🤔

sfc-gh-pfus (approved):

PR #215, opened 2026-01-23
SNOW-2867890: Rename ud_connector to connector

sfc-gh-turbaszek:
Rename `snowflake.ud_connector` to `snowflake.connector` namespace so UD driver can be easily installed in place of old driver.

To achieve that old driver is installed as pre-install step in tests env. The installation is done to separate directory than side-packages.


.github/workflows/test-python.yml
```diff
@@ -155,108 +155,109 @@ jobs:
...
-            --junitxml=reports/reference-${{ env.REFERENCE_OS_VERSION }}-${{ env.REFERENCE_PYTHON_VERSION }}.xml
-        working-directory: python
-
-      - name: Upload test results as GitHub artifact
-        uses: actions/upload-artifact@v4
-        if: always()  # Upload test results even if tests fail (needed for comparison)
-        with:
-          name: results-reference-${{ env.REFERENCE_OS_VERSION }}-${{ env.REFERENCE_PYTHON_VERSION }}
-          path: python/reports/*-${{ env.REFERENCE_OS_VERSION }}-${{ env.REFERENCE_PYTHON_VERSION }}.*
+#  python_tests_reference:
```
sfc-gh-fpawlowski, line 158: nit: if u would like to make sure not to forget about each place we have such commented-out code you can leave a TODO with jira ticket associated with fix of reference tests. But not a must have - there arent many such places anyway 👍

tests/performance/drivers/python/app/connection.py
```diff
@@ -52,14 +52,25 @@ def execute_setup_queries(cursor, setup_queries):
     print("✓ Setup queries completed")
 
 
+def _load_from_sources():
```
sfc-gh-fpawlowski, line 55: If i understood correctly our discussion - this is temporal and we should be able to achieve installation of the old driver without using different src in the following PRs?

sfc-gh-fpawlowski (approved):

tests/performance/drivers/python/app/connection.py
sfc-gh-turbaszek: Correct, this will be removed totally

PR #216, opened 2026-01-26
NO-SNOW: Fix ArrowIterator implementation

sfc-gh-turbaszek:
* Add missing `pytz` dependency
* Fix `SnowLogger` path in cpp code

### DIFF
--- a/python/src/snowflake/ud_connector/cursor.py
+++ b/python/src/snowflake/ud_connector/cursor.py
@@ -53,6 +53,9 @@ class Cursor:
         self.execute_result: Any = None
         self._iterator: Iterator[Row] | None = None
         self.execute_result = None
+        # Keep arrow_context and stream alive to prevent C++ from accessing freed memory
+        self._arrow_context: Any = None
+        self._stream: Any = None
 
     @property
     def description(self) -> Any:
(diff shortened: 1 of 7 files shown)


python/src/snowflake/ud_connector/cursor.py
```diff
@@ -124,8 +124,6 @@ def execute(self, operation: str, parameters: Sequence[Any] | dict[str, Any] | N
         self.execute_result = self.connection.db_api.statement_execute_query(
             StatementExecuteQueryRequest(stmt_handle=stmt_handle)
         ).result
-        # Reset streaming state for a new result
-        self._iterator = None
```
sfc-gh-pczajka, line 128: This will break _ensure_iterator, used by fetch() -> it will use old iterator if "execute()" is called multiple times.

sfc-gh-turbaszek: Good catch, fixed

sfc-gh-pczajka (approved):
LGTM. Yests are failing, however not because of this PR:
```
snowflake.connector.errors.ProgrammingError: 099404 (54000): 01c1fdaf-0813-c162-000c-a90ead9a9453: Exceeded maximum of 15 programmatic access tokens.
```

PR #217, opened 2026-01-27
NO-SNOW: Run python test in parallel

sfc-gh-turbaszek:
As per note in [seq docs](https://docs.snowflake.com/en/sql-reference/functions/seq) following query does not produce deterministic result:
`"SELECT seq8() as id FROM TABLE(GENERATOR(ROWCOUNT => 1000000)) v ORDER BY id"`
additionally, any `select * from table` is not deterministic. [Snowflake does not guarantee](https://docs.snowflake.com/en/sql-reference/sql/select#usage-notes) that rows will be sorted. To have deterministic result `ORDER BY` must be used.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -69,6 +69,8 @@ test = [
     "pytest-json-report",
     "pytest-split",
     "pytest-cov",
+    "pytest-randomly",
+    "pytest-xdist",
     "hypothesis",
     "requests",
     "brotli",
@@ -89,6 +91,7 @@ addopts = [
     "--strict-markers",
     "--strict-config",
     "-v",
+    "-n 8",
 ]
 markers = [
     "skip_reference(reason: str=''): skip when --connector=reference",
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -89,7 +89,7 @@ addopts = [
     "--strict-markers",
     "--strict-config",
     "-vv",
-    "-n 0",
+    "-n 16",
     "--durations=0",
 ]
 markers = [
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -69,6 +69,8 @@ test = [
     "pytest-split",
     "pytest-cov",
     "pytest-xdist",
+    "pytest-randomly",
+    "pytest-timeout",
     "hypothesis",
     "requests",
     "brotli",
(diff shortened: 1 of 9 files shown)


.github/workflows/test-python.yml
```diff
@@ -123,8 +123,11 @@ jobs:
           PARAMETERS_SECRET: ${{ secrets.PARAMETERS_SECRET }}
 
       - name: Install Hatch
+        working-directory: python
         run: uv tool install hatch
 
+      - name: Install universal driver from wheel
+        run: hatch run test.py${{ matrix.py }}:all-cov
```
sfc-gh-pczajka: "Install project and run tests?"

python/pyproject.toml
```diff
@@ -191,7 +194,7 @@ python = ["3.9", "3.10", "3.11", "3.12", "3.13"]
 
 [tool.hatch.envs.test.scripts]
 # Install from source
-install-wheel = "pip install dist/*.whl"
+install-wheel = "pip install {env:WHEEL_PATH:dist/*.whl}"
```
sfc-gh-pczajka: Should we call this command in test workflow?

python/tests/config.py
```diff
@@ -12,7 +16,7 @@ def get_test_parameters() -> dict[str, Any]:
         or environment variables as fallback.
     """
     # First try environment variable
-    parameter_path = os.environ.get("PARAMETER_PATH")
+    parameter_path = os.environ.get("PARAMETER_PATH", PROJECT_ROOT / "parameters.json")
```
sfc-gh-turbaszek, line 19: Thanks to this change you don't have to care about setting env variable.

python/tests/e2e/types/test_number.py
```diff
@@ -191,27 +191,37 @@ def test_should_download_large_result_set_with_multiple_chunks_from_generator_fo
     ):
         # Given Snowflake client is logged in
 
-        # When Query "SELECT seq8()::<type>(38,0), (seq8() + 0.12345)::<type>(20,5)
-        # FROM TABLE(GENERATOR(ROWCOUNT => 1000000)) v" is executed
+        # When Query
```
sfc-gh-pcyrek, line 194: Should we keep this code?

sfc-gh-pcyrek (approved):
lgtm

python/tests/e2e/types/test_number.py
sfc-gh-turbaszek: Yes, this comment is also in gherkins. In follow up PR I will clean up the gherkins to have proper query, that will require changing ODBC test. But to limit scope and unblock python tests I left this as is.

PR #218, opened 2026-01-27
NO-SNOW: Refactor reference tests to run in in isolation

sfc-gh-turbaszek:
This PR refactors the Python reference test infrastructure to run tests in proper isolation. Previously, tests could dynamically load different connector versions at runtime, which could lead to import side effects and cross-contamination between driver implementations.

Now, each test environment has only a single driver installed - either the universal driver or the reference driver (snowflake-connector-python). This ensures complete isolation: when testing the universal driver, no reference driver code is present in the environment, and vice versa.

The `IS_UNIVERSAL_DRIVER` constant in `python/tests/compatibility.py` auto-detects which driver is installed at test runtime by attempting to import a universal-driver-specific module (snowflake.connector._internal.api_client.client_api.database_driver_client). This approach enables:

* Conditional imports: Tests import exceptions from different module paths based on driver type (e.g., snowflake.connector.exceptions vs snowflake.connector.errors)
* Adapter selection: `ConnectorFactory.create_adapter()` returns the appropriate adapter type
* Test skipping: pytest_runtest_setup uses markers (`@pytest.mark.skip_universal`, `@pytest.mark.skip_reference`) to skip tests incompatible with the current driver
* Test metadata: Reports include which driver version was used
* Behavioral differences: NEW_DRIVER_ONLY() and OLD_DRIVER_ONLY() helpers allow test assertions to vary based on known behavioral differences (tracked by BC IDs)

CI runs the same test suite twice in separate jobs - once with the universal driver wheel installed, and once with the reference driver installed - then compares the results to detect regressions or behavioral differences.

### DIFF
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -22,8 +22,17 @@ jobs:
   detect-changes:
     uses: ./.github/workflows/detect-changes.yml
 
-  # Build Python wheel with compiled extensions for each Python version
+  # For backcompat
   build_rust_core:
+    needs: detect-changes
+    if: github.event_name == 'push' || needs.detect-changes.outputs.run_python == 'true'
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: "Dummy step"
+        run: "echo 'Dummy step to pass required check'"
+
+  # Build Python wheel with compiled extensions for each Python version
+  build_wheel:
     needs: detect-changes
     if: github.event_name == 'push' || needs.detect-changes.outputs.run_python == 'true'
     strategy:
@@ -86,7 +95,7 @@ jobs:
 
   # Combined unit + integration tests with universal connector
   python_tests:
-    needs: [detect-changes, build_rust_core, format]
+    needs: [detect-changes, build_wheel, format]
     if: github.event_name == 'push' || needs.detect-changes.outputs.run_python == 'true'
     strategy:
       fail-fast: false
@@ -165,7 +174,7 @@ jobs:
 
   # Integration tests only with reference connector (for comparison)
   python_tests_reference:
-    needs: [detect-changes, build_rust_core, format]
+    needs: [detect-changes, build_wheel, format]
     if: github.event_name == 'push' || needs.detect-changes.outputs.run_python == 'true'
     runs-on: ubuntu-latest
     continue-on-error: true  # Reference tests expected to fail due to behavioral differences
@@ -264,7 +273,7 @@ jobs:
         working-directory: python
 
   python-status:
-    needs: [detect-changes, build_rust_core, python_tests, python_tests_reference, python_compare_results]
+    needs: [detect-changes, build_wheel, python_tests, python_tests_reference, python_compare_results]
     if: always()
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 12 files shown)


python/tests/integ/test_connection.py
```diff
@@ -6,7 +6,13 @@
...
 import pytest
 
-from snowflake.connector.exceptions import NotSupportedError
+from tests.compatibility import IS_UNIVERSAL_DRIVER
+
+
+if IS_UNIVERSAL_DRIVER:
+    from snowflake.connector.exceptions import NotSupportedError
+else:
+    from snowflake.connector.errors import NotSupportedError
```
sfc-gh-turbaszek, line 15: Such code will clearly shows differences between old and new driver.

.github/workflows/test-python.yml
```diff
@@ -22,11 +22,25 @@ jobs:
...
+
+  # Build Python wheel with compiled extensions for each Python version
+  build_wheel:
+    needs: detect-changes
+    if: github.event_name == 'push' || needs.detect-changes.outputs.run_python == 'true'
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        py: ["3.9", "3.10", "3.11", "3.12", "3.13"]
```
sfc-gh-pczajka, line 42: Do we still want to support 3.9?

.github/workflows/test-python.yml
```diff
@@ -22,11 +22,25 @@ jobs:
...
 
-  # Build Rust core library once and share across Python jobs
+  # For backcompat
   build_rust_core:
     needs: detect-changes
     if: github.event_name == 'push' || needs.detect-changes.outputs.run_python == 'true'
-    runs-on: ubuntu-latest
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: "Dummy step"
```
sfc-gh-pczajka, line 31: I'd build wheel here, giving this job a "name" to something like "building wheel (including rust core)", keeping the ID not to break dependencies. That would keep the job's point (check whether rust still builds), WDYTH?

python/tests/compatibility.py
```diff
@@ -1,20 +1,22 @@
-from typing import Optional
+def _check_if_universal():
+    try:
+        # Import universal driver specific code
+        from snowflake.connector._internal.api_client.client_api import database_driver_client  # noqa
```
sfc-gh-pczajka, line 4: Do we want to add a check for this check (in case this path changes in 2 years)?

.github/workflows/test-python.yml
sfc-gh-turbaszek: It was there before changes, so I haven't remove it

.github/workflows/test-python.yml
sfc-gh-turbaszek: Been there, done that. Does not work :D

python/tests/compatibility.py
sfc-gh-turbaszek: I'm not sure. In 2 yers perspective we probably won't have reference tests

sfc-gh-pczajka (approved):

PR #221, opened 2026-01-29
SNOW-3022768: ODBC API tests for data source connection functions

sfc-gh-ddas:
Added granular ODBC unit and integration tests to verify compliance of ODBC wrapper with the ODBC 3.x and 2.x specs and consistent behaviour with existing ODBC driver.

This PR only contains ["Connecting to data sources" category functions](https://docs.google.com/spreadsheets/d/1BjOHFnLQ2AooXGqWncsaqCLIw8z9XD0v3fwwF-S-E3g/edit?gid=999869428#gid=999869428).

### DIFF
--- a/odbc_tests/CMakeLists.txt
+++ b/odbc_tests/CMakeLists.txt
@@ -34,8 +34,10 @@ if (NOT DEFINED DRIVER_TYPE)
   set(DRIVER_TYPE "NEW")
 endif()
 
-# Find odbc library
+# Find required libraries
 find_package(ODBC REQUIRED)
+find_package(Threads REQUIRED)
+
 message(STATUS "ODBC_INCLUDE_DIRS: ${ODBC_INCLUDE_DIRS}")
 message(STATUS "ODBC_LIBRARIES: ${ODBC_LIBRARIES}")
 
@@ -43,7 +45,7 @@ add_subdirectory(common)
 
 function(add_odbc_test name)
   add_executable(${name} ${ARGN})
-  target_link_libraries(${name} PRIVATE Catch2::Catch2WithMain ODBC::ODBC common)
+  target_link_libraries(${name} PRIVATE Catch2::Catch2WithMain ODBC::ODBC common Threads::Threads)
   target_include_directories(${name} PRIVATE ${picojson_SOURCE_DIR})
   catch_discover_tests(${name} TEST_PREFIX "${name}: ")
 endfunction()
--- a/odbc_tests/common/include/EnvOverride.hpp
+++ b/odbc_tests/common/include/EnvOverride.hpp
@@ -1,0 +1,79 @@
+#ifndef ENV_OVERRIDE_HPP
+#define ENV_OVERRIDE_HPP
+
+#include <optional>
+#include <string>
+
+// RAII class for temporarily overriding environment variables
+// TODO: Unix-only (uses setenv/unsetenv). Windows requires _putenv_s/_putenv("VAR=")
+class EnvOverride {
+ public:
+  // Sets the environment variable to the new value, saving the original.
+  EnvOverride(const std::string& name, const std::string& value) : name_(name) {
+    // Save original value
+    if (const char* original = std::getenv(name.c_str()); original != nullptr) {
+      original_value_ = std::string(original);
+    }
+    // Set new value
+    setenv(name.c_str(), value.c_str(), 1);
+  }
+
+  // Unsets the environment variable, saving the original.
+  explicit EnvOverride(const std::string& name) : name_(name) {
+    // Save original value
+    if (const char* original = std::getenv(name.c_str()); original != nullptr) {
+      original_value_ = std::string(original);
+    }
+    // Unset the variable
+    unsetenv(name.c_str());
+  }
+
+  ~EnvOverride() {
+    if (original_value_.has_value()) {
+      // Restore original value
+      setenv(name_.c_str(), original_value_->c_str(), 1);
+    } else {
+      // Variable was not set originally, unset it
+      unsetenv(name_.c_str());
+    }
+  }
+
+  // Non-copyable
+  EnvOverride(const EnvOverride&) = delete;
+  EnvOverride& operator=(const EnvOverride&) = delete;
+
+  // Movable
+  EnvOverride(EnvOverride&& other) noexcept
+      : name_(std::move(other.name_)), original_value_(std::move(other.original_value_)) {
+    other.name_.clear();  // Mark as moved-from
+  }
+
+  EnvOverride& operator=(EnvOverride&& other) noexcept {
+    if (this != &other) {
+      // Restore our original value before taking on new responsibility
+      if (!name_.empty()) {
+        if (original_value_.has_value()) {
+          setenv(name_.c_str(), original_value_->c_str(), 1);
+        } else {
+          unsetenv(name_.c_str());
+        }
+      }
+      name_ = std::move(other.name_);
+      original_value_ = std::move(other.original_value_);
+      other.name_.clear();
+    }
+    return *this;
+  }
+
+  // Get the variable name
+  [[nodiscard]] const std::string& name() const { return name_; }
+
+  // Get the original value (if it was set)
+  [[nodiscard]] const std::optional<std::string>& original_value() const { return original_value_; }
+
+ private:
+  std::string name_;
+  std::optional<std::string> original_value_;
+};
+
+#endif  // ENV_OVERRIDE_HPP
--- a/odbc_tests/common/include/ODBCConfig.hpp
+++ b/odbc_tests/common/include/ODBCConfig.hpp
@@ -1,0 +1,115 @@
+#ifndef ODBC_CONFIG_HPP
+#define ODBC_CONFIG_HPP
+
+#include <picojson.h>
+
+#include <map>
+#include <memory>
+#include <optional>
+#include <string>
+#include <vector>
+
+#include "EnvOverride.hpp"
+
+// Forward declarations
+class DriverConfig;
+class DataSourceConfig;
+class ConfigInstallation;
+
+// ============================================================================
+// DriverConfig - Manages ODBC driver configuration
+// ============================================================================
+
+class DriverConfig {
+ public:
+  // Factory method
+  static std::shared_ptr<DriverConfig> Default();
+
+  // Builder methods
+  DriverConfig& set(const std::string& key, const std::string& value);
+  DriverConfig& remove(const std::string& key);
+
+  // Accessors
+  [[nodiscard]] const std::map<std::string, std::string>& parameters() const;
+  [[nodiscard]] static std::string name();
+  static std::string get_driver_path();
+
+ private:
+  std::map<std::string, std::string> parameters_;
+};
+
+// ============================================================================
+// DataSourceConfig - Manages ODBC data source configuration
+// ============================================================================
+
+class DataSourceConfig {
+ public:
+  // Factory methods
+  static DataSourceConfig Snowflake(const std::string& connection_name = "testconnection");
+  static DataSourceConfig SnowflakeNoAuth(const std::string& connection_name = "testconnection");
+
+  // Builder methods
+  DataSourceConfig& set(const std::string& key, const std::string& value);
+  DataSourceConfig& remove(const std::string& key);
+  DataSourceConfig& driver_config(const std::optional<std::shared_ptr<DriverConfig>>& dc);
+  DataSourceConfig& name(const std::string& name);
+
+  // Accessors
+  [[nodiscard]] const std::string& name() const;
+  [[nodiscard]] const std::map<std::string, std::string>& parameters() const;
+  [[nodiscard]] std::optional<std::shared_ptr<DriverConfig>> driver_config() const;
+
+  // Installation
+  ConfigInstallation install();
+
+ private:
+  std::string name_;
+  std::map<std::string, std::string> parameters_;
+  std::optional<std::shared_ptr<DriverConfig>> driver_config_;
+
+  // Helper methods
+  static picojson::object load_parameters(const std::string& connection_name);
+  static std::string get_string(const picojson::object& obj, const std::string& key,
+                                 const std::string& default_value = "");
+};
+
+// ============================================================================
+// ConfigInstallation - RAII class for managing installed ODBC configuration
+// ============================================================================
+
+class ConfigInstallation {
+ public:
+  // Factory method
+  static ConfigInstallation install(const std::vector<DataSourceConfig>& data_sources);
+
+  // Destructor
+  ~ConfigInstallation();
+
+  // Non-copyable
+  ConfigInstallation(const ConfigInstallation&) = delete;
+  ConfigInstallation& operator=(const ConfigInstallation&) = delete;
+
+  // Movable
+  ConfigInstallation(ConfigInstallation&& other) noexcept;
+  ConfigInstallation& operator=(ConfigInstallation&& other) noexcept;
+
+  // Accessors
+  [[nodiscard]] const std::string& config_dir() const;
+  [[nodiscard]] std::string dsn_name(size_t index = 0) const;
+
+ private:
+  // Private constructor (use factory method)
+  explicit ConfigInstallation(const std::vector<DataSourceConfig>& data_sources);
+
+  // Helper methods
+  static std::string create_temp_dir();
+  void write_odbcinst_ini() const;
+  void write_odbc_ini() const;
+
+  // Members
+  std::string config_dir_;
+  std::vector<DataSourceConfig> data_sources_;
+  std::vector<EnvOverride> env_overrides_;
+};
+
+#endif  // ODBC_CONFIG_HPP
--- a/odbc_tests/common/include/ODBCFixtures.hpp
+++ b/odbc_tests/common/include/ODBCFixtures.hpp
@@ -1,0 +1,91 @@
+#ifndef ODBCFIXTURES_HPP
+#define ODBCFIXTURES_HPP
+
+#include <sql.h>
+#include <sqlext.h>
+#include <optional>
+
+#include <catch2/catch_test_macros.hpp>
+#include <utility>
+#include "HandleWrapper.hpp"
+#include "ODBCConfig.hpp"
+
+// ============================================================================
+// Base Fixtures (Parameterized via Constructor)
+// ============================================================================
+
+class EnvFixture {
+public:
+  std::optional<ConfigInstallation> config;
+  std::optional<EnvironmentHandleWrapper> env_wrapper;
+  SQLHENV env = SQL_NULL_HENV;
+
+  // Constructor with optional DSN configuration
+  explicit EnvFixture(std::optional<DataSourceConfig> dsn_config = std::nullopt) {
+    // Install DSN BEFORE creating ENV handle (critical for UnixODBC caching)
+    if (dsn_config.has_value()) {
+      config = dsn_config->install();
+    }
+
+    // Create ENV handle (will see installed DSN)
+    env_wrapper.emplace();
+    env = env_wrapper->getHandle();
+    SQLRETURN ret = SQLSetEnvAttr(env, SQL_ATTR_ODBC_VERSION, reinterpret_cast<SQLPOINTER>(SQL_OV_ODBC3), 0);
+    REQUIRE(ret == SQL_SUCCESS);
+  }
+
+  // Disable copy and move (RAII resource management)
+  EnvFixture(const EnvFixture&) = delete;
+  EnvFixture& operator=(const EnvFixture&) = delete;
+  EnvFixture(EnvFixture&&) = delete;
+  EnvFixture& operator=(EnvFixture&&) = delete;
+
+  SQLHENV env_handle() const { return env; }
+};
+
+class DbcFixture : public EnvFixture {
+public:
+  std::optional<ConnectionHandleWrapper> dbc_wrapper;
+  SQLHDBC dbc = SQL_NULL_HDBC;
+
+  // Constructor with optional DSN configuration
+  explicit DbcFixture(std::optional<DataSourceConfig> dsn_config = std::nullopt)
+    : EnvFixture(std::move(dsn_config)) {
+    dbc_wrapper.emplace(env_wrapper->createConnectionHandle());
+    dbc = dbc_wrapper->getHandle();
+  }
+
+  // Disable copy and move (RAII resource management)
+  DbcFixture(const DbcFixture&) = delete;
+  DbcFixture& operator=(const DbcFixture&) = delete;
+  DbcFixture(DbcFixture&&) = delete;
+  DbcFixture& operator=(DbcFixture&&) = delete;
+
+  SQLHDBC dbc_handle() const { return dbc; }
+};
+
+// ============================================================================
+// Convenience Fixtures for TEST_CASE_METHOD (Pre-configured DSNs)
+// ============================================================================
+
+class EnvDefaultDSNFixture : public EnvFixture {
+public:
+  EnvDefaultDSNFixture() : EnvFixture(DataSourceConfig::Snowflake()) {}
+};
+
+class DbcDefaultDSNFixture : public DbcFixture {
+public:
+  DbcDefaultDSNFixture() : DbcFixture(DataSourceConfig::Snowflake()) {}
+};
+
+class EnvNoAuthDSNFixture : public EnvFixture {
+public:
+  EnvNoAuthDSNFixture() : EnvFixture(DataSourceConfig::SnowflakeNoAuth()) {}
+};
+
+class DbcNoAuthDSNFixture : public DbcFixture {
+public:
+  DbcNoAuthDSNFixture() : DbcFixture(DataSourceConfig::SnowflakeNoAuth()) {}
+};
+
+#endif // ODBCFIXTURES_HPP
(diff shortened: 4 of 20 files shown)


sfc-gh-jszczerbinski (changes requested):
In general looks goods, but I don't like the how we setup DSN. For UD I would like running tests not to require almost any setup - unlike old ODBC.

I think we should use RAII to setup DSNs. Look how Schema::use_random_schema() works. We should follow a same pattern and have a DSN object that:
- Creates schema file when instantiated, and sets env variables so the driver knows where the schema is.
- Deletes schema file and set env variables to values that were set before.

We can reuse EnvOverride from old odbc to manage env vars: https://github.com/snowflakedb/snowflake-odbc/blob/26e2affc02c10b13d78c210267f1efd8aa631d15/Tests/EnvUtil.hpp#L47

We should also implement function that accepts parameters (from parameters.json) and constructs DSN file and returns DSN object.

Let's also talk tomorrow about overall approach - I think you overdid how much coverage we have - at least at this point. I think it's still added value since we are gonna need those tests at some point. I just don't think it should be priority now.

odbc_tests/common/include/DSNConfig.hpp
```diff
@@ -0,0 +1,255 @@
...
+    write_odbc_ini();
+
+    // Set environment variables (RAII - will be restored on destruction)
+    env_overrides_.emplace_back("ODBCSYSINI", config_dir_);
+    env_overrides_.emplace_back("ODBCINI", config_dir_ + "/odbc.ini");
+  }
+
+  ~DSNConfig() {
+    // Clean up config files and directory
+    std::error_code ec;
```
sfc-gh-jszczerbinski: Let's log something when we fail to remove the files - we are not checking ec at all right now.

sfc-gh-jszczerbinski: Maybe even add CHECK statement?

odbc_tests/common/include/test_macros.hpp
```diff
@@ -0,0 +1,42 @@
...
+    REQUIRE(ret == SQL_SUCCESS_WITH_INFO);                                     \
+    auto _records = get_diag_records(handleType, handle);                      \
+    REQUIRE(!_records.empty());                                                \
+    INFO("SQLSTATE: " << _records[0].sqlState                                  \
+                      << ", Message: " << _records[0].messageText);            \
+    REQUIRE(_records[0].sqlState == expectedState);                            \
+  } while (0)
+
+// Skip test if running against the new Universal Driver
+#define SKIP_NEW_DRIVER_NOT_IMPLEMENTED()                                      \
```
sfc-gh-jszczerbinski: Let's move this to compatibility.hpp

odbc_tests/common/include/test_macros.hpp
```diff
@@ -0,0 +1,42 @@
...
+  do {                                                                         \
+    if (TESTING_NEW_DRIVER) {                                                  \
```
sfc-gh-jszczerbinski: We dont need this "TESTING_NEW_DRIVER" use 

```
get_driver_type
```

odbc_tests/common/src/test_utils.cpp
```diff
@@ -0,0 +1,21 @@
+#include "test_utils.hpp"
+
+std::vector<DiagRec> get_diag_records(SQLSMALLINT handle_type, SQLHANDLE handle) {
```
sfc-gh-jszczerbinski: There is a similar function in get_diag_rec.hpp, make them share implementation and use overloading instead of diffrent names

odbc_tests/common/include/DSNFixture.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+#include <memory>
+#include <mutex>
+#include <stdexcept>
+
+#include "DSNConfig.hpp"
+#include "test_utils.hpp"
+
+// Lazy singleton for DSNConfig.
+class DSNFixture {
```
sfc-gh-jszczerbinski: Let's not do this singleton:   
\- Creating files on FS is inexpensive compared to establishing connections to snowflake.  
\- Poor test isolation <- test that require config, but do not create it.

odbc_tests/common/include/DSNConfig.hpp
```diff
@@ -0,0 +1,255 @@
...
+#include <fstream>
+#include <random>
+#include <sstream>
+#include <string>
+#include <vector>
+
+#include "EnvOverride.hpp"
+
+// RAII class for managing ODBC DSN configuration
+class DSNConfig {
```
sfc-gh-jszczerbinski: I don't like the API here, it's too rigid. Let's do sth that enables following use cases:  
Creating default config:

```
auto config = DataSourceConfig::Snowflake() // Just data about data source configuration
// Equivalent to ConfigInstallation::install({config});
ConfigInstallation config_installed = config.install() // Config installation is RAII object that manages the installed config

SQLConnect(.., "DSN=" + config.name(), ...)
// Rest of the test
```

Modyfing config before creation:

```
auto config = DataSourceConfig::Snowflake()
   .set("Port", 8081) // Sets some config key to integer value and returns reference
   .set("Host", "localhost") // Set another key to string value
auto config_installed = config.install()
```

Multiple configs:

```
auto default_config = DataSourceConfig::Snowflake();
auto noauth_config = DataSourceConfig::SnowflakeNoAuth();
auto config_installed = ConfigInstallation::install({default_config, noauth_config});
```

Custom driver config:

```
auto driver_config = DriverConfig::default() // This the config we use install() calls
   .set("Driver", "/custom/path/to/driver.so"); // Overriding driver config value
auto default_config = DataSourceConfig::Snowflake().driver_config(driver_config);
auto config_installed = ConfigInstallation::install({ds_config});
```

No driver config:

```
auto ds_config = DriverConfig::Snowflake() // Let's test if we can pass path to so
  .driver_config({})
  .set("Driver", get_driver_path()); // instead of driver config name
auto config_installed = 
   ConfigInstallation::install({ds_config});
```

In summary I think we need couple of objects that are composable:

```
class DataSourceConfig {
  static DataSourceConfig Snowflake();
   static DataSourceConfig SnowflakeNoAuth() {
     return Snowflake()
       .delete("UID");
       .delete("PWD");
   }
   DataSourceConfig& set(std::string key, ...);
   DataSourceConfig& delete(std::string key);
   DataSourceConfig& driver_config(std::optional<std::shared_ptr<DriverConfig>> dc);

   ConfigInstallation install() {
      return ConfigInstallation::install({*this});
   }
private:
   std::optional<std::shared_ptr<DriverConfig>> driver_config;
   std::string name;
   std::map<std::string, ParameterValue> parameters;
}

class DriverConfig {  // Similar for DataSourceConfig (we can set, get and delete values)
  static std::shared_ptr<DriverConfig> default();
}

// This would have to implementations
// For unix: managing ini files and env vars
// For windows: modifying the registry <- we can add that one later
class ConfigInstallation {
   ConfigInstallation install(std::vector<DataSourceConfig>)
}

```

odbc_tests/common/include/DSNConfig.hpp
```diff
@@ -0,0 +1,255 @@
...
+    int tracing = 0;
+  };
+
+  // Create DSNConfig from parameters.json for a specific connection.
+  // Creates two DSNs: one with full credentials and one without (for error testing).
+  static DSNConfig from_parameters(const std::string& connection_name = "testconnection") {
+    const auto params = load_parameters(connection_name);
+
+    DSNEntry full_dsn;
+    full_dsn.name = "SnowflakeDSII";
```
sfc-gh-jszczerbinski: Remove DSII from name <- Let's not use Simba terms

odbc_tests/common/include/EnvOverride.hpp
```diff
@@ -0,0 +1,78 @@
...
+    if (original_value_.has_value()) {
+      // Restore original value
+      setenv(name_.c_str(), original_value_->c_str(), 1);
+    } else {
+      // Variable was not set originally, unset it
+      unsetenv(name_.c_str());
+    }
+  }
+
+  // Non-copyable
```
sfc-gh-jszczerbinski, line 41: Rule of 5 implemented: very nice :)

odbc_tests/common/include/test_macros.hpp
```diff
@@ -0,0 +1,42 @@
...
+#include <catch2/catch_test_macros.hpp>
+
+#include "get_diag_rec.hpp"
+
+#ifndef TESTING_NEW_DRIVER
+#define TESTING_NEW_DRIVER 0
+#endif
+
+// Helper macro to check for expected ODBC error with specific SQLSTATE
+#define CHECK_EXPECTED_ERROR(ret, expectedState, handle, handleType)           \
```
sfc-gh-jszczerbinski: Let's follow the Catch convention: use require for short-circuiting assertions

odbc_tests/common/include/test_utils.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+// Get diagnostic records from a raw ODBC handle
+// This is a convenience function for tests that work with raw handles
+// instead of HandleWrapper objects
+std::vector<DiagRec> get_diag_records(SQLSMALLINT handle_type, SQLHANDLE handle);
+
+// Check if a test DSN is configured
+// Returns true if ODBC DSN configuration is available
+// Will return true when a DSNConfig object is in scope
+inline bool is_test_dsn_configured() {
```
sfc-gh-jszczerbinski: Do we need this if we have configurations managed by RAII?

odbc_tests/common/include/test_utils.hpp
```diff
@@ -0,0 +1,42 @@
...
+  const char* odbcsysini = std::getenv("ODBCSYSINI");
+  const char* odbcini = std::getenv("ODBCINI");
+  return (odbcsysini != nullptr && odbcini != nullptr);
+}
+
+// Check if PARAMETER_PATH is set
+inline bool is_parameter_path_configured() {
```
sfc-gh-jszczerbinski: Do we need this? We already check that in get_test_parameters

sfc-gh-jszczerbinski (changes requested):
In general the PR looks good:  
\- We need to get ds configuration right. Not getting it right will cause us to change a lot of tests later on. <- that's why i left so many comments.  
\- The tests look good to me, a little overkill, but as we discussed that's fine for this PR.  
\- I think you are missing pthreads on Linux - that's why coverage build is failing. Add them in "modern" CMake way - use https://cmake.org/cmake/help/latest/module/FindThreads.html

odbc_tests/CMakeLists.txt
```diff
@@ -34,6 +34,24 @@ if (NOT DEFINED DRIVER_TYPE)
   set(DRIVER_TYPE "NEW")
 endif()
 
+set(CATCH_TAGS "")
```
sfc-gh-jszczerbinski: I think I deleted this part on main

sfc-gh-jszczerbinski: Let's keep it that way since we don't need   add_compile_definitions(TESTING_NEW_DRIVER=0)

odbc_tests/common/include/DSNConfig.hpp
```diff
@@ -0,0 +1,255 @@
...
+    std::string warehouse;
+    std::string role;
+    int port = 443;
+    bool ssl = true;
+    int tracing = 0;
+  };
+
+  // Create DSNConfig from parameters.json for a specific connection.
+  // Creates two DSNs: one with full credentials and one without (for error testing).
+  static DSNConfig from_parameters(const std::string& connection_name = "testconnection") {
```
sfc-gh-jszczerbinski: I don't think we should create more than one DS for each test.

odbc_tests/common/include/DSNConfig.hpp
sfc-gh-ddas: Refactored, please let me know if this is what you had in mind.

odbc_tests/common/include/DSNConfig.hpp
sfc-gh-ddas: Updated to use the new API.

odbc_tests/common/include/DSNConfig.hpp
sfc-gh-ddas: Removed. I had no idea DSII is a Simba term.

odbc_tests/common/include/DSNConfig.hpp
sfc-gh-ddas: Added logs when temp dir fails to be deleted. Did not add a CHECK because that would throw an exception which is not allowed in destructors.

odbc_tests/common/include/DSNFixture.hpp
sfc-gh-ddas: Removed singleton.

odbc_tests/common/include/test_macros.hpp
sfc-gh-ddas: Changed.

odbc_tests/common/include/test_macros.hpp
sfc-gh-ddas: Done.

odbc_tests/common/include/test_macros.hpp
sfc-gh-ddas: Removed.

odbc_tests/common/src/test_utils.cpp
sfc-gh-ddas: Fixed.

odbc_tests/CMakeLists.txt
sfc-gh-ddas: Removed.

odbc_tests/common/include/ODBCConfig.hpp
```diff
@@ -0,0 +1,148 @@
...
+
+class DriverConfig {
+ public:
+  // Compiler-generated special members
+  DriverConfig() = default;
+  ~DriverConfig() = default;
+  DriverConfig(const DriverConfig&) = default;
+  DriverConfig& operator=(const DriverConfig&) = default;
+  DriverConfig(DriverConfig&&) noexcept = default;
+  DriverConfig& operator=(DriverConfig&&) noexcept = default;
```
sfc-gh-makowalski: Shouldn't we just omit those if they are all defaulted?

odbc_tests/common/include/ODBCConfig.hpp
```diff
@@ -0,0 +1,148 @@
...
+
+class DataSourceConfig {
+ public:
+  // Compiler-generated special members
+  DataSourceConfig() = default;
+  ~DataSourceConfig() = default;
+  DataSourceConfig(const DataSourceConfig&) = default;
+  DataSourceConfig& operator=(const DataSourceConfig&) = default;
+  DataSourceConfig(DataSourceConfig&&) noexcept = default;
+  DataSourceConfig& operator=(DataSourceConfig&&) noexcept = default;
```
sfc-gh-makowalski: Same here

odbc_tests/common/src/ODBCConfig.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+ParameterValue::ParameterValue(const char* value) : value_(std::string(value)) {}
+
+ParameterValue::ParameterValue(const int value) : value_(std::to_string(value)) {}
+
+ParameterValue::ParameterValue(const bool value) : value_(value ? "true" : "false") {}
+
+std::string ParameterValue::as_string() const {
+  return value_;
+}
```
sfc-gh-makowalski: Do we really need it? Can't we just use strings?

odbc_tests/common/src/ODBCConfig.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+ConfigInstallation::ConfigInstallation(const std::vector<DataSourceConfig>& data_sources)
+    : data_sources_(data_sources) {
+  // TODO: Windows - Use registry
+  config_dir_ = create_temp_dir();
+  write_odbcinst_ini();
+  write_odbc_ini();
+  env_overrides_.emplace_back("ODBCSYSINI", config_dir_);
+  env_overrides_.emplace_back("ODBCINI", config_dir_ + "/odbc.ini");
```
sfc-gh-makowalski: Could we use `std::filesystem::path` instead of concatenating with `+`? (for future cross platform compatibility)

sfc-gh-makowalski: There are probably other places where it could also be used. So if you decide to implement it, just ask an agent to identify them.

odbc_tests/tests/odbc-api/connecting/alloc_handle_tests.cpp
```diff
@@ -0,0 +1,766 @@
...
+  ret = SQLSetEnvAttr(env, SQL_ATTR_ODBC_VERSION, reinterpret_cast<SQLPOINTER>(SQL_OV_ODBC3), 0);
+  REQUIRE(ret == SQL_SUCCESS);
+
+  ret = SQLAllocHandle(SQL_HANDLE_DBC, env, &dbc);
+  REQUIRE(ret == SQL_SUCCESS);
+  REQUIRE(dbc != SQL_NULL_HDBC);
+
+  SQLFreeHandle(SQL_HANDLE_DBC, dbc);
+  SQLFreeHandle(SQL_HANDLE_ENV, env);
+}
```
sfc-gh-makowalski, line 102: Please consider usage of Catch2 fixtures. For example we can simplify this test case to:

```
TEST_CASE_METHOD(EnvFixture, "SQLAllocHandle DBC: Basic allocation succeeds", "[odbc-api][alloc_handle][dbc][connecting]") {
  SQLHDBC dbc = SQL_NULL_HDBC;

  SQLRETURN ret = SQLAllocHandle(SQL_HANDLE_DBC, env, &dbc);
  REQUIRE(ret == SQL_SUCCESS);
  REQUIRE(dbc != SQL_NULL_HDBC);

  SQLFreeHandle(SQL_HANDLE_DBC, dbc);
}
```

The example fixtures I got from claude:

```
// odbc_tests/common/include/ODBCFixtures.hpp

#include <sql.h>
#include <sqlext.h>
#include <catch2/catch_test_macros.hpp>

// Helper to avoid verbose casts
inline SQLCHAR* sql_str(const std::string& s) {
  return reinterpret_cast<SQLCHAR*>(const_cast<char*>(s.c_str()));
}

// Fixture: Environment handle only
struct EnvFixture {
  SQLHENV env = SQL_NULL_HENV;

  EnvFixture() {
    SQLRETURN ret = SQLAllocHandle(SQL_HANDLE_ENV, SQL_NULL_HANDLE, &env);
    REQUIRE(ret == SQL_SUCCESS);
    ret = SQLSetEnvAttr(env, SQL_ATTR_ODBC_VERSION, 
                        reinterpret_cast<SQLPOINTER>(SQL_OV_ODBC3), 0);
    REQUIRE(ret == SQL_SUCCESS);
  }

  ~EnvFixture() {
    if (env != SQL_NULL_HENV) {
      SQLFreeHandle(SQL_HANDLE_ENV, env);
    }
  }
};

// Fixture: Environment + Connection handle
struct DbcFixture : EnvFixture {
  SQLHDBC dbc = SQL_NULL_HDBC;

  DbcFixture() {
    SQLRETURN ret = SQLAllocHandle(SQL_HANDLE_DBC, env, &dbc);
    REQUIRE(ret == SQL_SUCCESS);
  }

  ~DbcFixture() {
    if (dbc != SQL_NULL_HDBC) {
      SQLFreeHandle(SQL_HANDLE_DBC, dbc);
    }
  }
};

// Fixture: Full connection (connected state)
struct ConnectedFixture : DbcFixture {
  ConnectedFixture() {
    std::string connStr = get_connection_string();
    SQLRETURN ret = SQLDriverConnect(dbc, nullptr, sql_str(connStr), SQL_NTS,
                                      nullptr, 0, nullptr, SQL_DRIVER_NOPROMPT);
    REQUIRE(ret == SQL_SUCCESS);
  }

  ~ConnectedFixture() {
    SQLDisconnect(dbc);
  }
};
```

LMK what you think about it.

sfc-gh-makowalski (changes requested):

odbc_tests/common/include/ODBCConfig.hpp
sfc-gh-ddas: Makes sense 😅

odbc_tests/common/src/ODBCConfig.cpp
sfc-gh-ddas: You're right

odbc_tests/common/src/ODBCConfig.cpp
sfc-gh-ddas: Fixed.

odbc_tests/tests/odbc-api/connecting/alloc_handle_tests.cpp
sfc-gh-ddas: I think fixtures will be useful for the other API tests in upcoming PRs, however since in this PR we are specifically testing the connection functions, with a fixture here we would be calling the function under test inside the fixture.

odbc_tests/tests/odbc-api/connecting/driver_connect_tests.cpp
```diff
@@ -0,0 +1,1567 @@
...
+    if (*result == SQL_SUCCESS || *result == SQL_SUCCESS_WITH_INFO) {
+      // Verify with simple query
+      SQLHSTMT stmt = SQL_NULL_HSTMT;
+      if (SQLAllocHandle(SQL_HANDLE_STMT, dbc, &stmt) == SQL_SUCCESS) {
+        SQLExecDirect(stmt, reinterpret_cast<SQLCHAR*>(const_cast<char*>("SELECT 1")), SQL_NTS);
+        SQLFreeHandle(SQL_HANDLE_STMT, stmt);
+      }
+      SQLDisconnect(dbc);
+      *result = SQL_SUCCESS;
+    }
```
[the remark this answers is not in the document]
sfc-gh-ddas: In all early return paths, either no resources are allocated or they are freed correctly.

odbc_tests/tests/odbc-api/connecting/alloc_handle_tests.cpp
sfc-gh-ddas: Refactored to use fixtures after offline discussion.

odbc_tests/common/include/ODBCFixtures.hpp
```diff
@@ -0,0 +1,97 @@
...
+#include "ODBCConfig.hpp"
+
+// ============================================================================
+// Base Fixtures (Parameterized via Constructor)
+// ============================================================================
+
+class EnvFixture {
+public:
+  std::optional<ConfigInstallation> config;
+  SQLHENV env = SQL_NULL_HENV;
```
sfc-gh-jszczerbinski, line 21: Can we use 

```
EnvironmentHandleWrapper
```

sfc-gh-jszczerbinski: We won't need override destructor then

odbc_tests/common/src/ODBCConfig.cpp
```diff
@@ -0,0 +1,343 @@
...
+std::string DriverConfig::get_driver_path() {
+  if (const char* driver_path_env = std::getenv("DRIVER_PATH");
+      driver_path_env != nullptr && driver_path_env[0] != '\0') {
+    return driver_path_env;
+  }
+  return "/usr/lib/snowflake/odbc/lib/libSnowflake.so";
+}
+
+// ============================================================================
+// DataSourceConfig
```
sfc-gh-jszczerbinski: Let's put them in corresponding .cpp files, move DataSourceConfig to DataSourceConfig.cpp and the same for ConfigInstallation

odbc_tests/common/include/ODBCFixtures.hpp
```diff
@@ -0,0 +1,97 @@
...
+  // Disable copy and move (RAII resource management)
+  EnvFixture(const EnvFixture&) = delete;
+  EnvFixture& operator=(const EnvFixture&) = delete;
+  EnvFixture(EnvFixture&&) = delete;
+  EnvFixture& operator=(EnvFixture&&) = delete;
+};
+
+class DbcFixture : public EnvFixture {
+public:
+  SQLHDBC dbc = SQL_NULL_HDBC;
```
sfc-gh-jszczerbinski, line 49: Same heve with ConnectionHandleWrapper?

odbc_tests/common/src/ODBCConfig.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+std::string ConfigInstallation::create_temp_dir() {
+  std::random_device rd;
+  std::mt19937 gen(rd());
+  std::uniform_int_distribution<> dis(0, 999999);
+
+  // Include process ID to avoid race conditions when multiple test processes run in parallel
+  const int pid = GET_PROCESS_ID();
+  const std::string dir_name = "odbc_dsn_" + std::to_string(pid) + "_" + std::to_string(dis(gen));
+  const std::filesystem::path full_path = std::filesystem::temp_directory_path() / dir_name;
```
sfc-gh-jszczerbinski: Should we put it in temp? Those files contain sensitive data. If we place them in current working directory they should land cmake-build dir near the test binary (this is where ctest runs the tests):  
\- it will be in safer dir  
\- we could easily locate dsns when debugging a test

odbc_tests/common/include/DSNConfig.hpp
sfc-gh-jszczerbinski: Looks good

sfc-gh-jszczerbinski (approved):
Small comments, please fix them. But in general LGTM

odbc_tests/common/include/ODBCFixtures.hpp
sfc-gh-ddas: Done.

sfc-gh-makowalski (approved):

PR #222, opened 2026-01-29
NO-SNOW: Restructure hatch setup

sfc-gh-turbaszek:
Restructures the hatch env setup to distinguish between local (dev) and CI (test) envs. For dev we want to work with local sources while on CI we want to have e2e SDLC tested by installing driver from wheel.

Additionally this PR fixes uv and hatch wiring. Now hatch should us uv for dep management. That should help with CI speed.

### DIFF
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -264,7 +264,7 @@ jobs:
 
       - name: Compare integration test results
         run: |
-          hatch run ci:compare \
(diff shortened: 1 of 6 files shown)


sfc-gh-fpawlowski (approved):
looks good 🚢 
Just to make sure - so we will call hatch run dev:all since now (on local envs), right?

sfc-gh-turbaszek:
Yes, or selected test

PR #227, opened 2026-01-30
SNOW-3035531: Adding old python coverage reports

sfc-gh-pcyrek:
SNOW-3035531: Adding old python coverage reports


scripts/old_python/coverage_report.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# Write summary in lcov format (compatible with existing benchstore parser)
+with open(summary_file, 'w') as f:
+    f.write(f"lines......: {line_rate:.1f}% ({lines_covered} of {lines_valid} lines)\n")
+    f.write(f"functions..: no data found\n")
+
+print(f"")
+print(f"Old Python Connector Coverage Summary")
+print(f"=" * 40)
+print(f"  Line coverage:   {line_rate:.1f}% ({lines_covered}/{lines_valid} lines)")
```
sfc-gh-makowalski, line 58: cant this be a script that we just run here?

sfc-gh-makowalski (changes requested):

scripts/old_python/coverage_report.sh
sfc-gh-pcyrek: Technically, yeah, but we want to keep it externally. Basically, it measures external code; we don't want to make any kind of merge gate for the current code. This is for monitoring purposes only, and we agree that an external source is okay for that purpose.

sfc-gh-makowalski (approved):

PR #229, opened 2026-02-01
Fix validator duplicated names of feature files bug

sfc-gh-fpawlowski:
# fix(validator): Handle multiple feature files and language-specific folders

## Summary

Fixes three bugs in the Gherkin test validator that caused false validation failures.

### 1. Handle multiple feature files with same name

**Problem:** When multiple feature files had the same base name in different directories (e.g., `shared/session/logout.feature`, `core/session/logout.feature`, `python/session/logout.feature`), the validator treated them as the same feature, causing:
- False "orphaned tests" detection
- Incorrect step matching across unrelated features

**Solution:** 
- Use full relative path as unique feature identifier instead of just file stem
- Added `is_feature_relevant_for_language()` to correctly filter features by folder:
  - `shared/` features apply to all languages
  - `core/` features apply to Rust only
  - `python/` features apply to Python only
  - etc.

### 2. Only require test files when scenarios have implementation tags

**Problem:** Features in language-specific folders (e.g., `jdbc/`, `odbc/`) failed validation even when no scenarios had implementation tags (`@jdbc_e2e`, `@odbc_e2e`).

**Solution:** Only require test files when scenarios actually have implementation tags:
- Feature-level tag (`@jdbc`) = "This feature describes JDBC behavior"
- Scenario-level tag (`@jdbc_e2e`) = "This scenario has a test implementation"

Features that document planned behavior without implementations no longer cause failures.
This aligns validate_feature_with_path() with the existing logic in collect_all_scenarios_and_languages(), which already only adds languages to requirements when scenarios have implementation tags.

### 3. Fix Python method boundary detection

**Problem:** Python method boundary detection incorrectly ended methods when encountering nested `import` statements or nested `def` functions inside a test method.

**Solution:** Only match `method_end_patterns` when the line is at the same or lower indentation level as the method start. This prevents nested code from prematurely ending method boundaries.

## Testing

```bash
# Validator passes on all feature files
tests/tests_format_validator/target/debug/tests_format_validator
```

## Impact

| Scenario | Before | After |
|----------|--------|-------|
| `jdbc/session/logout.feature` (no `@jdbc_e2e` tags) | ❌ FAILED | ✅ PASSED |
| `odbc/session/logout.feature` (no `@odbc_e2e` tags) | ❌ FAILED | ✅ PASSED |
| Python tests with nested imports | ❌ Parsing errors | ✅ Correct |
| Multiple `logout.feature` files | ❌ False orphans | ✅ Independent |


tests/tests_format_validator/src/validator.rs
```diff
@@ -265,6 +292,21 @@ impl GherkinValidator {
...
+        let lang_folder = match language {
+            Language::Rust => "core",
+            Language::Python => "python",
+            Language::Jdbc => "jdbc",
+            Language::Odbc => "odbc",
+            _ => return false,
+        };
+
+        // Feature is relevant if it's in shared/ folder OR in the language-specific folder
+        feature_id.starts_with("shared/") || feature_id.starts_with(&format!("{}/", lang_folder))
```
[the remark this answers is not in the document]
sfc-gh-pczajka: That's a nice spot. Should we raise an error that there is "unknown prefix" rather than silently ignoring it?
Also: would it be easy to move forward if new wrapper is introduced?

tests/tests_format_validator/src/step_finder.rs
```diff
@@ -414,16 +414,26 @@ impl MethodBoundaryFinder {
                 }
 
                 // Fallback: Check if any end pattern matches (mainly for Python)
+                // Only match if the line is at the same or lower indentation level as the method start
```
sfc-gh-pczajka, line 417: This logic makes sense assuming that other languages keep indentation convention. I'm fine if we assume that.

In ideal world every language/language group would have their own set of rules - in Python logic could be simplified to "method continues as long as indentation is bigger that it's definition, and ends immediately if it drops to the same level (being careful about multiline strings)", but I'm not sure the refactor is worth the effort now.

tests/tests_format_validator/src/validator.rs
```diff
@@ -265,6 +292,21 @@ impl GherkinValidator {
         ))
     }
 
+    /// Check if a feature (identified by its path-based ID) is relevant for a given language.
+    /// A feature is relevant if it's in the shared/ folder OR in the language-specific folder.
+    fn is_feature_relevant_for_language(&self, feature_id: &str, language: &Language) -> bool {
```
sfc-gh-pczajka: Do we have to add this logic? Wouln'd tags + IDs be enough?

tests/tests_format_validator/src/validator.rs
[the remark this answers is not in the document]
sfc-gh-pczajka: EDIT: yes, we would just need to add it here

tests/tests_format_validator/src/validator.rs
```diff
@@ -191,6 +221,7 @@ impl GherkinValidator {
         std::collections::HashMap<(String, String), Vec<Language>>,
     )> {
         let mut scenarios = Vec::new();
+        // Use feature_id (relative path) as key to distinguish features with same name
```
sfc-gh-pczajka: Unnecessary comment

```suggestion
```

tests/tests_format_validator/src/validator.rs
```diff
@@ -437,32 +548,37 @@ impl GherkinValidator {
...
                             scenario_language_requirements
-                                .get(&(feature_name.clone(), scenario_name.clone()))
+                                .get(&(fid.clone(), scenario_name.clone()))
                                 .map(|langs| langs.contains(language))
                                 .unwrap_or(false)
                         } else {
                             false
                         }
                     })
-            } else {
```
sfc-gh-pczajka: Why is this removed?

tests/tests_format_validator/src/validator.rs
```diff
@@ -931,7 +1068,11 @@ impl GherkinValidator {
                         let mut method_missing_steps = Vec::new();
 
                         for step_text in &scenario_steps {
-                            if !method_steps.contains(step_text) {
+                            // Check if any implemented step starts with the expected step
```
sfc-gh-pczajka: What this changes? It is not documented in the description

tests/tests_format_validator/src/validator.rs
sfc-gh-fpawlowski: Achieved the same with
`matching_feature_ids.iter().any(|feature_id| {...}`

throught that .any call

tests/tests_format_validator/src/validator.rs
sfc-gh-fpawlowski: Good catch, changed back!

tests/tests_format_validator/src/validator.rs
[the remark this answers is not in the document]
sfc-gh-fpawlowski: > Resolved by removing is_feature_relevant_for_language() entirely (commit dabb8ea).
>
> The function was redundant - the existing tag-based language requirements already handle this correctly. Features are validated based on their @lang_e2e scenario tags, not folder location. We also added validate_feature_prefix() which raises an explicit error for unknown directories (e.g., typos like shares/ or rust/), failing fast on misconfiguration rather than silently ignoring.

tests/tests_format_validator/src/validator.rs
```diff
@@ -183,6 +183,36 @@ impl GherkinValidator {
...
+    fn get_feature_id(&self, feature_path: &Path) -> String {
+        // Get path relative to features_dir
+        if let Ok(relative) = feature_path.strip_prefix(&self.features_dir) {
+            // Remove .feature extension and convert to string
+            relative
+                .with_extension("")
+                .to_str()
+                .unwrap_or_else(|| feature_path.file_stem().unwrap().to_str().unwrap())
+                .to_string()
+        } else {
```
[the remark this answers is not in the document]
sfc-gh-fpawlowski: > Fixed in commit 5a23df6. The function now normalizes path separators to forward slashes:

tests/tests_format_validator/src/validator.rs
```diff
@@ -304,24 +346,34 @@ impl GherkinValidator {
...
+                    });
 
-                    if !feature_requires_language {
-                        // Feature doesn't require this language - determine why by checking the feature file directly
-                        let reason = self.determine_orphan_reason(feature_name, language)?;
+                    if !any_feature_requires_language {
+                        // No matching feature requires this language - determine why
+                        // Use the first matching feature for the reason
+                        let reason =
+                            self.determine_orphan_reason(matching_feature_ids[0], language)?;
```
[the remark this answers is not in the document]
sfc-gh-fpawlowski: > Fixed in commit 6aa2045. We now sort matching_feature_ids deterministically before selecting:

tests/tests_format_validator/src/validator.rs
```diff
@@ -1018,8 +1100,9 @@ impl GherkinValidator {
         let norm_impl = normalize(implemented_step);
         let norm_feature = normalize(feature_step);
 
-        // Require exact match after normalization - no partial matches allowed
-        norm_impl == norm_feature
+        // Match if implemented step starts with the expected feature step
+        // This handles cases where comments after the step are treated as continuations
+        norm_impl.starts_with(&norm_feature)
```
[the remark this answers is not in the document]
sfc-gh-fpawlowski: > Reverted to exact matching using contains(). The starts_with approach was removed.
>
> Regarding the continuation concern: You're correct that next-line comments can affect step extraction - this is due to the continuation logic in step_finder.rs which treats comment lines following a Gherkin step as continuations. However, this is pre-existing behavior not introduced by this PR.
>
> In practice, validation passes on all 33 current features because no test files have arbitrary comments immediately after Gherkin steps. If this becomes an issue, we can address it in a follow-up by either:
> 1. Removing the continuation logic entirely (steps = single line only)
> 2. Requiring a special marker for continuations (e.g., // ...)
>
> For this PR, the step matching is exact equality - same as before.

tests/tests_format_validator/src/step_finder.rs
```diff
@@ -414,16 +414,26 @@ impl MethodBoundaryFinder {
...
+
+                    // Only consider end patterns at same or lower indentation (not nested code)
+                    if line_indent <= start_line_indent {
+                        for pattern in self.config.method_end_patterns {
+                            if line == *pattern || line.starts_with(pattern) {
+                                method_end_line = Some(i);
+                                break;
+                            }
+                        }
+                        if method_end_line.is_some() {
```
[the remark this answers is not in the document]
sfc-gh-fpawlowski: > Already addressed in commit a97db90. Two regression tests were added in step_finder.rs:
>
> 1. test_python_method_boundary_with_nested_constructs (line 805)
> - Tests a Python method containing nested import, nested def helper(), and nested async def
> - Asserts all nested constructs are included in the method boundary
> - Asserts the next top-level def test_next_method is NOT included
>
> 2. test_python_method_boundary_ends_at_top_level_def (line 886)
> - Verifies that top-level def statements still correctly terminate method boundaries
> - Ensures the fix doesn't break normal method detection
>
> Both tests run as part of cargo test and are passing.

tests/tests_format_validator/src/validator.rs
sfc-gh-fpawlowski: Great catch, simplified

tests/tests_format_validator/src/step_finder.rs
sfc-gh-fpawlowski: Strongly agreed, should be refactored in the future

tests/tests_format_validator/src/validator.rs
```diff
@@ -183,6 +200,63 @@ impl GherkinValidator {
         Ok(untagged_features)
     }
 
+    /// Get a unique feature ID that includes the relative path to distinguish
+    /// features with the same name in different directories (e.g., shared/session/logout vs core/session/logout)
+    fn get_feature_id(&self, feature_path: &Path) -> String {
+        // Get path relative to features_dir
+        let raw_id = if let Ok(relative) = feature_path.strip_prefix(&self.features_dir) {
```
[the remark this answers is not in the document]
sfc-gh-fpawlowski: > Already addressed in commit 9cd04f4. Three regression tests were added in validator_tests.rs:
>
> 1. should_not_conflate_features_with_same_name_in_different_folders
> - Creates shared/session/logout.feature and core/session/logout.feature
> - Verifies they're treated as distinct features with independent language requirements
> - Asserts orphan detection doesn't conflate them
>
> 2. should_not_require_tests_when_no_implementation_tags_exist
> - Creates jdbc/session/logout.feature with only @jdbc tag (no @jdbc_e2e)
> - Verifies validation passes without requiring a test file
> - Documents "planned feature" behavior
>
> 3. should_require_tests_when_implementation_tags_exist
> - Creates jdbc/session/logout.feature with @jdbc_e2e scenario tag
> - Verifies validation correctly requires a test file
> - Confirms implementation tags trigger test requirements
>
> Also added create_feature_file_in_folder() helper for creating features in specific top-level folders (shared/, core/, jdbc/, etc.).

sf_core/src/crl/config.rs
```diff
@@ -3,9 +3,10 @@ use crate::config::settings::Settings;
 use chrono::Duration;
 use std::path::PathBuf;
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Default)]
 pub enum CertRevocationCheckMode {
     /// Default - disables CRL checking (TLS handshake still in place)
+    #[default]
     Disabled,
```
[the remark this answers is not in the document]
sfc-gh-fpawlowski: moved

tests/tests_format_validator/src/validator.rs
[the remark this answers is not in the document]
sfc-gh-fpawlowski: fixed already

tests/tests_format_validator/tests/validator_tests.rs
```diff
@@ -869,6 +869,204 @@ fn should_handle_multiple_breaking_changes_in_single_test_method() -> Result<()>
...
+        .collect();
+
+    assert!(
+        feature_paths.iter().any(|p| p.contains("shared/session")),
+        "Should find shared/session/logout.feature"
+    );
+    assert!(
+        feature_paths.iter().any(|p| p.contains("core/session")),
+        "Should find core/session/logout.feature"
+    );
```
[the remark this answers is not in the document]
sfc-gh-fpawlowski: > Fixed. Added path normalization in the test to handle Windows backslashes

tests/tests_format_validator/src/validator.rs
```diff
@@ -988,40 +1097,11 @@ impl GherkinValidator {
...
-            })
+            .filter(|feature_step| !implemented_steps.contains(feature_step))
             .cloned()
             .collect()
     }
 
-    fn steps_match(&self, implemented_step: &str, feature_step: &str) -> bool {
-        // Normalize both steps for comparison - only remove punctuation, keep all words
-        let normalize = |s: &str| {
-            s.to_lowercase()
```
sfc-gh-pczajka: Did we got rid of normalization at all?

.github/workflows/test-odbc.yml
sfc-gh-pczajka, line 62: Woah, what happened here?

sfc-gh-fpawlowski: This is actually already present on the main branch - as well as the crl.config.rs changes. I have no idea why Github locks and shows those changes here as well - when you do diff locally it does not show those changes.. I can create a new PR but we will loose history of comments. I already tried closing and opening this one.

tests/tests_format_validator/src/validator.rs
sfc-gh-fpawlowski: I did it after the yesterday discussion regarding the replacement of 'contains' with iteration + applying 'steps_match' function. However, I actually prefer minimal changes so I just reverted only this thing back. Should be good now - and I checked that both places where we added steps_match application require it.

.github/workflows/test-odbc.yml
sfc-gh-fpawlowski: Fixed it by changing base branch to smth else and then back to main (2 times as the first time I did such switch did not solve it..) - no code changes were required - simply Github lock

sfc-gh-pczajka (approved):

sfc-gh-pcyrek (approved):
lgtm

PR #232, opened 2026-02-02
Lower numeric large tests size to 3-4 chunks

sfc-gh-pczajka:
Lower large test size for numeric types to 3-4 chunks, which is:
- 20k for decfloat
- 30k for number
- 50k for int and float

Values were adjusted experimentally from local machine

### DIFF
--- a/odbc_tests/tests/e2e/types/int.cpp
+++ b/odbc_tests/tests/e2e/types/int.cpp
@@ -68,13 +68,13 @@ TEST_CASE("should download large result set with multiple chunks for int and syn
   // Given Snowflake client is logged in
   Connection conn;
 
-  // When Query "SELECT seq8()::<type> as id FROM TABLE(GENERATOR(ROWCOUNT => 1000000)) v ORDER BY id" is executed
+  // When Query "SELECT seq8()::<type> as id FROM TABLE(GENERATOR(ROWCOUNT => 50000)) v ORDER BY id" is executed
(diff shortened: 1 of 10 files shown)


odbc_tests/tests/e2e/types/int.cpp
```diff
@@ -68,13 +68,13 @@ TEST_CASE("should download large result set with multiple chunks for int and syn
   // Given Snowflake client is logged in
   Connection conn;
 
-  // When Query "SELECT seq8()::<type> as id FROM TABLE(GENERATOR(ROWCOUNT => 1000000)) v ORDER BY id" is executed
+  // When Query "SELECT seq8()::<type> as id FROM TABLE(GENERATOR(ROWCOUNT => 50000)) v ORDER BY id" is executed
```
sfc-gh-rkowalski, line 71: How reliable is that command? Do we always get the same chunks or they differ in size/number?

sfc-gh-rkowalski (approved):
Overall looks good

odbc_tests/tests/e2e/types/int.cpp
sfc-gh-pczajka: They might differ depending on server magic, some session parameters and (but don't quote me) connection speed. I assumed that office connection is optimistic + I left a bit of buffer here (3-4 chunks, the test requires at least 2)

PR #234, opened 2026-02-02
SNOW-2997748: Add fetchmany

sfc-gh-turbaszek:


python/src/snowflake/connector/cursor.py
```diff
@@ -170,30 +168,26 @@ def _get_stream_ptr(self) -> int:
...
-            stream_ptr = self._get_stream_ptr()
-            arrow_context = ArrowConverterContext()
-            self._iterator = ArrowStreamIterator(
-                stream_ptr,
-                arrow_context,
-                # TODO: SNOW-2997742, SNOW-2997786, temporarily hardcoded
-                use_dict_result=False,
-                use_numpy=False,
-            )
+    def _ensure_iterator(self) -> ArrowStreamIterator:
```
sfc-gh-pczajka: Why to change this logic? If so - could we rename it? It does not ensure anymore, it always creates one

python/tests/integ/test_cursor.py
```diff
@@ -170,6 +167,117 @@ def test_fetchall_empty_result(self, cursor):
...
+            ORDER BY 1
+            """
+        )
+        batch = cursor.fetchmany(2)
+        remaining = cursor.fetchall()
+
+        assert batch == [(0,), (1,)]
+        assert remaining == [(2,), (3,), (4,)]
+
+    def test_fetchmany_with_multiple_columns(self, cursor):
```
sfc-gh-pczajka, line 273: Could we add testcases for:
- `fetchmany` x `fetchall` (fetchall should get the remainder)
- `fetchone`/`fetchmany` x second `execute` (second execute should reset state - fetches should start anew)

python/src/snowflake/connector/cursor.py
sfc-gh-turbaszek: To please mypy instead of adding `type: ignore`

python/tests/integ/test_cursor.py
sfc-gh-turbaszek: Added

python/src/snowflake/connector/cursor.py
sfc-gh-turbaszek: Renamed

sfc-gh-pczajka (approved):

PR #239, opened 2026-02-03
[SNOW-3035385] add windows odbc tests

sfc-gh-makowalski:


.github/workflows/test-odbc.yml
```diff
@@ -136,6 +162,21 @@ jobs:
...
+          dir target\debug\
+          $env:DRIVER_PATH = "$PWD\target\debug\${{ matrix.driver_lib }}"
+          $env:PARAMETER_PATH = "$PWD\parameters.json"
+          $NPROC = (Get-CimInstance Win32_ComputerSystem).NumberOfLogicalProcessors
+          
+          cd odbc_tests
+          New-Item -ItemType Directory -Force -Path cmake-build | Out-Null
+          cmake -B cmake-build -D DRIVER_TYPE=NEW .
+          cmake --build cmake-build --config Debug --parallel $NPROC
+          ctest -j $NPROC -C Debug --test-dir cmake-build --output-on-failure
```
sfc-gh-turbaszek: The only thing I would consider is potentially extracting this to some scripts in case we would like to run those tests on windows machine during debugging

sfc-gh-turbaszek (approved):

.github/workflows/test-odbc.yml
sfc-gh-makowalski: That's a good idea and it looks cleaner in the yml file. I'll add it

PR #240, opened 2026-02-03
SNOW-3045931: add-python-github-actions-woa-64-workflows-for-ud-build-and-test

sfc-gh-fpawlowski:


.github/workflows/test-jdbc.yml
```diff
@@ -31,27 +40,48 @@ jobs:
...
         env:
           PARAMETERS_SECRET: ${{ secrets.PARAMETERS_SECRET }}
 
+      - name: Decode secrets (Windows)
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          gpg --batch --yes --passphrase $env:PARAMETERS_SECRET --decrypt --output parameters.json .github/secrets/parameters_aws.json.gpg
+        env:
+          PARAMETERS_SECRET: ${{ secrets.PARAMETERS_SECRET }}
```
sfc-gh-turbaszek, line 43: Why not use
```
      - name: Decode secrets
        run: ./scripts/decode_secrets.sh
        env:
          PARAMETERS_SECRET: ${{ secrets.PARAMETERS_SECRET }}
```
as we do in other workflows?

.github/workflows/test-python.yml
```diff
@@ -131,32 +165,57 @@ jobs:
...
           PARAMETERS_SECRET: ${{ secrets.PARAMETERS_SECRET }}
 
+      - name: Decode secrets (Windows)
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          gpg --batch --yes --passphrase $env:PARAMETERS_SECRET --decrypt --output parameters.json .github/secrets/parameters_aws.json.gpg
+        env:
+          PARAMETERS_SECRET: ${{ secrets.PARAMETERS_SECRET }}
+
```
sfc-gh-turbaszek: Oh, I see. Can we wrap this in similar script and reuse the step instead of duplicating this command?

sfc-gh-fpawlowski: done, good catch

.github/workflows/test-python.yml
```diff
@@ -26,28 +26,41 @@ jobs:
...
+  # Build Python wheel with compiled extensions for each OS/Python version
   build_wheel:
     needs: detect-changes
     if: github.event_name == 'push' || needs.detect-changes.outputs.run_python == 'true'
     strategy: &matrix
       fail-fast: false
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, windows-11-arm]
         py: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
```
sfc-gh-turbaszek, line 45: I know this is out of scope bu after this PR we will have 9 parallel runs. and MacOS and normal windows is not yet here. There's a max limit for number of PATs (100). So we are getting close to it. Imo we should consider optimizing the testing strategy. I would opt for 3.10 + 3.14. Everything else can be run on main after merge Let's plan for it.

sfc-gh-turbaszek (approved):
Small comment but otherwise looks good to me 🚀

PR #245, opened 2026-02-03
SNOW-2467887: Metadata support in python

sfc-gh-turbaszek:
Add support for sfqid, rowcount and column metadata


python/src/snowflake/connector/cursor.py
```diff
@@ -168,6 +201,44 @@ def _get_stream_ptr(self) -> int:
...
+            internal_size = col.byte_length if col.HasField("byte_length") else None
+
+            # precision and scale for numeric types
+            precision = col.precision if col.HasField("precision") else None
+            scale = col.scale if col.HasField("scale") else None
+
+            # nullable flag
+            null_ok = col.nullable
+
+            description.append((name, type_code, display_size, internal_size, precision, scale, null_ok))
```
sfc-gh-fpawlowski: Nit: we could create such NamedTuple with a method from_column().
It shouldnt break our contract as NamedTuple is only an additional layer on top of normal tuple. 
Pep249 only says it needs to be a Sequence.

python/src/snowflake/connector/cursor.py
```diff
@@ -42,47 +53,57 @@ def __init__(self, connection: Connection) -> None:
...
     @property
-    def rowcount(self) -> int:
+    def rowcount(self) -> int | None:
         """
         Read-only attribute specifying the number of rows that the last
         .execute*() produced or affected.
 
         Returns:
-            int: Number of rows affected, or -1 if not determined
+            int: Number of rows affected, or None if not determined
```
sfc-gh-fpawlowski, line 141: To be sure: this default is caused by the backward compatibility, right?

sf_core/src/apis/database_driver_v1/statement.rs
```diff
@@ -24,6 +26,71 @@ use std::{collections::HashMap, sync::Arc};
...
+            && !rowset.is_empty()
+            && !rowset[0].is_empty()
+        {
+            let mut affected_rows = 0i64;
+
+            // Look for specific column names that indicate affected rows
+            for (idx, col) in row_types.iter().enumerate() {
+                let col_name = col.name.to_lowercase();
+
+                if (col_name == "number of rows updated"
```
sfc-gh-fpawlowski: Should we have this list of column names expected to contain data of rows affected extracted into array of constants or smth like that?

sf_core/src/apis/database_driver_v1/statement.rs
```diff
@@ -24,6 +26,71 @@ use std::{collections::HashMap, sync::Arc};
...
+
+            return affected_rows;
+        }
+        // DML with no affected rows
+        return 0;
+    }
+
+    // For SELECT and other queries, use total
+    // Return -1 if total is not available
+    data.total.unwrap_or(-1)
```
sfc-gh-fpawlowski: should we name this default as UNKNOWN_ROWS_AFFECTED_DEFAULT

python/src/snowflake/connector/cursor.py
sfc-gh-turbaszek: Yes + plus PEP 249 states 
```
Future versions of the DB API specification could redefine the latter case to have the object return None instead of -1.
```

sf_core/src/apis/database_driver_v1/statement.rs
sfc-gh-turbaszek: Done

sf_core/src/apis/database_driver_v1/statement.rs
sfc-gh-turbaszek: Done

python/src/snowflake/connector/cursor.py
sfc-gh-turbaszek: Done

sfc-gh-fpawlowski (approved):

PR #248, opened 2026-02-05
SNOW-3046149: Fix sdist to include Rust sources for building from source

sfc-gh-turbaszek:
The sdist was previously invalid because it lacked sf_core sources needed to build the native extension. This change:

- Add symlinks in python/ for Rust sources (sf_core, proto_utils, Cargo.toml)
- Configure hatch to dereference directory symlinks and include Rust sources
- Use force-include for Cargo.toml.sdist (minimal workspace config)
- Update hatch_build.py to use local Cargo.toml path (works for both layouts)
- Add packaging tests to verify sdist can be built and installed from source

The sdist now includes all necessary Rust sources and can be used to build wheels on systems with a Rust toolchain installed.


python/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "snowflake-connector-python-ud"  # TODO: change this once ready, needs switching tests to UD
```
sfc-gh-fpawlowski, line 6: Side note - we may want to have some kind of epic with PuPR readiness and collect them such items (and then add Jira ticket numer here). Similar to how it was with asyncio.

Anyway, nothing to be done in scope of this PR.

sfc-gh-fpawlowski (approved):
lgtm, but packaging tests seem failing

PR #263, opened 2026-02-10
SNOW-3022768: ODBC API tests for statement preparation functions

sfc-gh-ddas:
Added granular ODBC unit and integration tests to verify compliance of ODBC wrapper with the ODBC 3.x and 2.x specs and consistent behaviour with existing ODBC driver.

This PR only contains ["Preparing SQL requests" category functions](https://docs.google.com/spreadsheets/d/1BjOHFnLQ2AooXGqWncsaqCLIw8z9XD0v3fwwF-S-E3g/edit?gid=999869428#gid=999869428).

### DIFF
--- a/odbc_tests/tests/odbc-api/preparing/bind_parameter_tests.cpp
+++ b/odbc_tests/tests/odbc-api/preparing/bind_parameter_tests.cpp
@@ -1,0 +1,315 @@
+#include <sql.h>
+#include <sqlext.h>
+#include <sqltypes.h>
+
+#include <cstring>
+#include <string>
+
+#include <catch2/catch_test_macros.hpp>
+
+#include "ODBCConfig.hpp"
+#include "ODBCFixtures.hpp"
+#include "compatibility.hpp"
+#include "get_diag_rec.hpp"
+#include "test_macros.hpp"
+#include "test_setup.hpp"
+
+// ============================================================================
+// SQLBindParameter - Basic Functionality
+// ============================================================================
+
+TEST_CASE_METHOD(StmtDefaultDSNFixture, "SQLBindParameter: Binds integer input parameter",
+                 "[odbc-api][bindparameter][preparing]") {
+  SKIP_NEW_DRIVER_NOT_IMPLEMENTED();
+
+  SQLRETURN ret = SQLPrepare(stmt_handle(), reinterpret_cast<SQLCHAR*>(const_cast<char*>("SELECT ? AS val")), SQL_NTS);
+  REQUIRE(ret == SQL_SUCCESS);
+
+  SQLINTEGER param_value = 42;
+  SQLLEN indicator = 0;
+  ret =
+      SQLBindParameter(stmt_handle(), 1, SQL_PARAM_INPUT, SQL_C_SLONG, SQL_INTEGER, 0, 0, &param_value, 0, &indicator);
+  REQUIRE(ret == SQL_SUCCESS);
+
+  ret = SQLExecute(stmt_handle());
+  REQUIRE(ret == SQL_SUCCESS);
+
+  SQLINTEGER result = 0;
+  SQLLEN result_indicator = 0;
+  ret = SQLBindCol(stmt_handle(), 1, SQL_C_SLONG, &result, 0, &result_indicator);
+  REQUIRE(ret == SQL_SUCCESS);
+
+  ret = SQLFetch(stmt_handle());
+  REQUIRE(ret == SQL_SUCCESS);
+  REQUIRE(result == 42);
+}
+
+TEST_CASE_METHOD(StmtDefaultDSNFixture, "SQLBindParameter: Binds string input parameter",
+                 "[odbc-api][bindparameter][preparing]") {
+  SKIP_NEW_DRIVER_NOT_IMPLEMENTED();
+
+  SQLRETURN ret = SQLPrepare(stmt_handle(), reinterpret_cast<SQLCHAR*>(const_cast<char*>("SELECT ? AS val")), SQL_NTS);
+  REQUIRE(ret == SQL_SUCCESS);
+
+  char param_value[] = "hello";
+  SQLLEN indicator = SQL_NTS;
+  ret = SQLBindParameter(stmt_handle(), 1, SQL_PARAM_INPUT, SQL_C_CHAR, SQL_VARCHAR, strlen(param_value), 0,
+                         param_value, 0, &indicator);
+  REQUIRE(ret == SQL_SUCCESS);
+
+  ret = SQLExecute(stmt_handle());
+  REQUIRE(ret == SQL_SUCCESS);
+
+  char result[64] = {};
+  SQLLEN result_indicator = 0;
+  ret = SQLBindCol(stmt_handle(), 1, SQL_C_CHAR, result, sizeof(result), &result_indicator);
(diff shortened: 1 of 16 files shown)


sfc-gh-jszczerbinski (approved):
LGTM

odbc_tests/tests/odbc-api/preparing/bind_parameter_tests.cpp
```diff
@@ -0,0 +1,450 @@
...
+  SKIP_NEW_DRIVER_NOT_IMPLEMENTED();
+
+  SQLRETURN ret =
+      SQLConnect(dbc_handle(), reinterpret_cast<SQLCHAR*>(const_cast<char*>(config.value().dsn_name().c_str())),
+                 SQL_NTS, nullptr, 0, nullptr, 0);
+  REQUIRE(ret == SQL_SUCCESS);
+
+  SQLHSTMT stmt = SQL_NULL_HSTMT;
+  ret = SQLAllocHandle(SQL_HANDLE_STMT, dbc_handle(), &stmt);
+  REQUIRE(ret == SQL_SUCCESS);
```
sfc-gh-makowalski: Maybe let's move this ...

odbc_tests/tests/odbc-api/preparing/bind_parameter_tests.cpp
```diff
@@ -0,0 +1,450 @@
...
+  SQLLEN result_indicator = 0;
+  ret = SQLBindCol(stmt, 1, SQL_C_SLONG, &result, 0, &result_indicator);
+  REQUIRE(ret == SQL_SUCCESS);
+
+  ret = SQLFetch(stmt);
+  REQUIRE(ret == SQL_SUCCESS);
+  REQUIRE(result == 42);
+
+  SQLFreeHandle(SQL_HANDLE_STMT, stmt);
+  SQLDisconnect(dbc_handle());
```
sfc-gh-makowalski: ... to a fixture? 😄

odbc_tests/tests/odbc-api/preparing/bind_parameter_tests.cpp
```diff
@@ -0,0 +1,450 @@
...
+  REQUIRE(ret == SQL_SUCCESS);
+
+  char result[64] = {};
+  SQLLEN result_indicator = 0;
+  ret = SQLBindCol(stmt, 1, SQL_C_CHAR, result, sizeof(result), &result_indicator);
+  REQUIRE(ret == SQL_SUCCESS);
+
+  ret = SQLFetch(stmt);
+  REQUIRE(ret == SQL_SUCCESS);
+  REQUIRE(std::string(result) == "hello");
```
sfc-gh-makowalski, line 70: Shouldn't we `#include <string>` for this to work? AI tells me that it works right now because it's included transitively by Catch2 or some other header, but in case they change we should include it explicitly.

odbc_tests/tests/odbc-api/preparing/bind_parameter_tests.cpp
```diff
@@ -0,0 +1,450 @@
...
+  SQLRETURN ret =
+      SQLConnect(dbc_handle(), reinterpret_cast<SQLCHAR*>(const_cast<char*>(config.value().dsn_name().c_str())),
+                 SQL_NTS, nullptr, 0, nullptr, 0);
+  REQUIRE(ret == SQL_SUCCESS);
+
+  SQLHSTMT stmt = SQL_NULL_HSTMT;
+  ret = SQLAllocHandle(SQL_HANDLE_STMT, dbc_handle(), &stmt);
+  REQUIRE(ret == SQL_SUCCESS);
+
+  ret = SQLPrepare(stmt, reinterpret_cast<SQLCHAR*>(const_cast<char*>("SELECT ? AS val")), SQL_NTS);
```
sfc-gh-makowalski: Maybe we could add a helper like:
```
#define SQLSTR(s) reinterpret_cast<SQLCHAR*>(const_cast<char*>(s))
```

so we could turn this:
```
ret = SQLPrepare(stmt, reinterpret_cast<SQLCHAR*>(const_cast<char*>("SELECT ? AS val")), SQL_NTS);
```
into:
```
ret = SQLPrepare(stmt, SQLSTR("SELECT ? AS val"), SQL_NTS);
```

Imo it looks a bit more pleasant, but I imagine not everyone likes these kinds of macros and this PR is not the first where we have this line of code.

odbc_tests/tests/odbc-api/preparing/bind_parameter_tests.cpp
```diff
@@ -0,0 +1,450 @@
...
+
+  SQLINTEGER param_value = 1;
+  SQLLEN indicator = 0;
+  // HY003: Invalid application buffer type (9999 is not a valid C data type)
+  ret = SQLBindParameter(stmt, 1, SQL_PARAM_INPUT, 9999, SQL_INTEGER, 0, 0, &param_value, 0, &indicator);
+  REQUIRE_EXPECTED_ERROR(ret, "HY003", stmt, SQL_HANDLE_STMT);
+
+  SQLFreeHandle(SQL_HANDLE_STMT, stmt);
+  SQLDisconnect(dbc_handle());
+}
```
sfc-gh-makowalski: Why do we do `SQLConnect(...)` here but we omit it in test for invalid handle (line 226). Is one of them caught internally and the other one by the server?

sfc-gh-makowalski (changes requested):

odbc_tests/tests/odbc-api/preparing/bind_parameter_tests.cpp
sfc-gh-ddas: IMO, the current setup is good to keep the intent explicit.

odbc_tests/tests/odbc-api/preparing/bind_parameter_tests.cpp
sfc-gh-ddas: Inlcuded. I removed it because Clion showed the include directive as greyed out. Is the double include a good pattern in C++?

odbc_tests/tests/odbc-api/preparing/bind_parameter_tests.cpp
sfc-gh-ddas: Both should be caught internally, however the handle validation runs first before the connection logic.

odbc_tests/tests/odbc-api/preparing/bind_parameter_tests.cpp
sfc-gh-ddas: Added a new fixture, this does however increase some overhead on CI because now the tests that should skip on new driver will also establish connection before skipping. I deemed this as acceptable though because eventually we intend to remove the skip macros from each test anyhow.

odbc_tests/tests/odbc-api/preparing/bind_parameter_tests.cpp
sfc-gh-makowalski: I wouldn't call it double include, but rather "include what you use". If we refactor the header file that included it some time from now then we could introduce a bug here. I'd like to avoid that.

sfc-gh-makowalski: Resolve if you agree!

sfc-gh-makowalski (approved):

PR #265, opened 2026-02-11
Implement JSON parameter binding in Rust core

sfc-gh-pczajka:
- Add protobuf StringPtr-based JSON binding support (zero-copy path)
- Implement parse_json_bindings with Cow<'static, RawValue> for zero allocation when receiving JSON from language wrappers
- Switch query_request bindings from serde_json::Value to Cow<RawValue> for raw JSON pass-through (no Value tree built)
- Update async_exec and snowflake query functions for new binding types
- Add comprehensive unit tests for JSON bindings (error cases, zero-copy verification, Arrow backwards-compat, Request serialization round-trip)
- Enable serde_json raw_value feature
- Update ODBC wrapper to pass bindings: None for compatibility

Follow-up PR:
- #266


sf_core/src/apis/database_driver_v1/statement.rs
```diff
@@ -217,11 +139,12 @@ pub unsafe fn statement_bind(
...
     pub stream: Box<FFI_ArrowArrayStream>,
     pub rows_affected: i64,
-    pub query_id: String,
-    pub columns: Vec<ColumnMetadata>,
 }
 
-pub fn statement_execute_query(stmt_handle: Handle) -> Result<ExecuteResult, ApiError> {
+pub fn statement_execute_query(
+    stmt_handle: Handle,
+    proto_bindings: Option<crate::protobuf_gen::database_driver_v1::query_bindings::BindingType>,
```
sfc-gh-jszczerbinski: Let's not have generate protobuf types here. Most likely we will use this interface at some point for ADBC without protobuf and protobuf will be optional compilation feature. Let's not have protobuf as depedency where it doesn't need to be

sf_core/src/apis/database_driver_v1/statement.rs
```diff
@@ -523,6 +477,94 @@ pub enum StatementError {
...
+
+    // Validate pointer is not null
+    if ptr_value == 0 {
+        return UnsupportedBindParameterTypeSnafu {
+            type_: "Null pointer in bindings".to_string(),
+        }
+        .fail();
+    }
+
+    // Dereference pointer to get JSON bytes as a &'static slice.
```
sfc-gh-jszczerbinski: This is not safe, the bytes don't have static lifetime.

sf_core/src/protobuf_gen/database_driver_v1.rs
sfc-gh-jszczerbinski, line 632: ???

sfc-gh-jszczerbinski: Why are we rewriting the whole file here?

sfc-gh-jszczerbinski (changes requested):
I think we should have a better way of passing wrapper owned strings. The current solution is a recipe for incorrect code and crashes. We cannot have 'static lifetime for things that are only guaranteed to exist for the lifetime of function call

sf_core/src/apis/database_driver_v1/statement.rs
sfc-gh-pczajka: Proposition:  
wrapper around buffer pointer, which does not destroy the contents.  
avoid using static  
do we need to validate?

sfc-gh-pczajka: not public pointer

sf_core/src/apis/database_driver_v1/statement.rs
sfc-gh-pczajka: Done :+1:

sf_core/src/protobuf_gen/database_driver_v1.rs
sfc-gh-pczajka: CI/CD check enforced regeneration

sfc-gh-jszczerbinski: I think the issue is that we are not formatting it after generation

sf_core/src/apis/database_driver_v1/statement.rs
```diff
@@ -20,88 +18,43 @@ use arrow::{
...
+/// # Important
+/// This points to **bytes**, not string objects:
+/// - JSON: UTF-8 encoded JSON bytes (e.g., from Python `json_str.encode('utf-8')`)
+/// - CSV: Raw CSV file bytes
+///
+/// Avoids protobuf dependency so this interface can be used with ADBC and other protocols.
+#[derive(Debug, Clone)]
+pub struct DataPtr {
+    /// 8-byte pointer value encoded as little-endian bytes
+    pub value: Vec<u8>,
```
sfc-gh-jszczerbinski: At this point we should have parsed pointer from protobuf message and this field should be `value: *const u8`

sfc-gh-jszczerbinski: We should also enforce lifetime
- the DataPtr should not derive Clone - as we do not copy the underlying memory.
- Let's keep the fields private so the user of this struct don't have access to a raw ptr
- Let's have pub fn new() so we can construct the value on the caller side
- Let's have pub fn slice<'a>(&'a self) -> &'a [u8] so the callee can access the data.

sf_core/src/apis/database_driver_v1/statement.rs
```diff
@@ -523,6 +503,85 @@ pub enum StatementError {
...
+///
+/// The caller (language wrapper) MUST guarantee:
+/// 1. The pointer points to valid UTF-8 encoded JSON bytes
+/// 2. The data remains valid and unmodified for the entire `statement_execute_query` call
+/// 3. `statement_execute_query` is called synchronously (blocks until HTTP completes)
+///
+/// The `'static` lifetime is a lie (data doesn't live forever), but it's safe because
+/// the synchronous call guarantees the data lives long enough. This is documented and
+/// understood by wrapper implementors.
+fn parse_json_bindings<'a>(data_ptr: &'a DataPtr) -> Result<&'a RawValue, StatementError> {
```
sfc-gh-jszczerbinski: Let's split this function:
- Byte array to ptr conversion should occur in protobuf layer
- Let's put ptr + length to slice conversion in DataPtr method so we can ensure that the slice doesn't outlive DataPtr object.
- Let's do the Json validation in this function

sfc-gh-jszczerbinski (changes requested):
I made some changes to resolve lifetime issues: Those were mostly caused by `with_valid_session`, because we were returning Future that captured local references provided by the closure of Fn object.

Still there are some issue in the PR, let's resolve them.

sf_core/src/apis/database_driver_v1/statement.rs
sfc-gh-pczajka: done

sf_core/src/apis/database_driver_v1/statement.rs
sfc-gh-pczajka: done

sfc-gh-jszczerbinski (approved):

PR #266, opened 2026-02-11
Small bindings implementation in Python

sfc-gh-pczajka:
This PR is part of stack:
 - #265
 - this PR

Implement small parameter binding in Python. In scope:
- `?` and `:1` syntaxes
- JSON-only parameter serialization

Not in scope:
- csv serialization
- python-specific formatting
- `paramastyle` connection parameter


python/src/snowflake/connector/_internal/binding_serializer.py
```diff
@@ -0,0 +1,253 @@
...
+_ZERO_EPOCH_DATE = date(1970, 1, 1)
+_ZERO_EPOCH = datetime.fromtimestamp(0, timezone.utc).replace(tzinfo=None)
+
+
+class BindingSerializer:
+    """Serializes Python parameters to Snowflake binding JSON format."""
+
+    # Python type to Snowflake type mapping
+    # Mirrors PYTHON_TO_SNOWFLAKE_TYPE from the reference connector's converter.py
+    TYPE_MAP = {
```
sfc-gh-turbaszek: It's yet another type mapping. How does this compares to https://github.com/snowflakedb/universal-driver/blob/main/python/src/snowflake/connector/_internal/type_codes.py ?

Any chances to unify? Imo, we should keep any kind of such mapping in single place to have good overview how many of such thing we have.

Potentially, maybe it would be worth to have `TypeMapping` class that we use everywhere mapping happens.

python/src/snowflake/connector/cursor.py
```diff
@@ -25,6 +29,26 @@
 Row = tuple[Any, ...]
 
 
+def _convert_proto_error(exc: ProtoApplicationException) -> DatabaseError:
```
sfc-gh-turbaszek: Is this change related to the PR?

python/src/snowflake/connector/cursor.py
```diff
@@ -104,40 +130,96 @@ def close(self) -> None:
...
+                string_ptr = StringPtr(
+                    value=ptr_bytes,  # 8-byte pointer value
+                    length=length,
+                )
+
+                # Create request with bindings (no CopyFrom - direct assignment)
+                request = StatementExecuteQueryRequest(stmt_handle=stmt_handle, bindings=QueryBindings(json=string_ptr))
+            else:
+                request = StatementExecuteQueryRequest(stmt_handle=stmt_handle)
+        else:
```
sfc-gh-turbaszek, line 250: Should we introduce a helper method to encapsulate  logic of building QueryBindings?

python/src/snowflake/connector/cursor.py
```diff
@@ -104,40 +130,96 @@ def close(self) -> None:
...
+            if len(params) != first_len:
+                raise ProgrammingError(
+                    f"Parameter sequence at index {i} has {len(params)} elements, expected {first_len}"
+                )
+
+        # Transpose from row-major to column-major format
+        # Input:  [(row1_col1, row1_col2), (row2_col1, row2_col2), ...]
+        # Output: [[row1_col1, row2_col1, ...], [row1_col2, row2_col2, ...]]
+        num_columns = first_len
+        transposed = [[row[col_idx] for row in seq_of_parameters] for col_idx in range(num_columns)]
```
sfc-gh-turbaszek, line 299: Why this is needed?

tests/definitions/shared/query/parameter_binding.feature
```diff
@@ -0,0 +1,165 @@
...
+    Given Snowflake client is logged in
+    When Query with 20 positional parameters is executed with values [0..19]
+    Then Result should contain all 20 values in order
+
+  @python_e2e
+  Scenario: should bind parameters with OR clause for multiple value matching
+    Given Snowflake client is logged in
+    And A temporary table with columns (id NUMBER, name VARCHAR) exists
+    And Rows [1, "Alice"], [2, "Bob"], [3, "Charlie"], [4, "David"], [5, "Eve"] are inserted
+    When Query "SELECT FROM {table_name} WHERE id = ? OR id = ? OR id = ? ORDER BY id" is executed with parameters [1, 3, 5]
```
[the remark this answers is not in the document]
sfc-gh-pczajka: That's necessary as validator does not validate

python/src/snowflake/connector/cursor.py
sfc-gh-pczajka: Old syntax requires user to provide input row by row, the json needs to be generated in columnar format

python/src/snowflake/connector/cursor.py
sfc-gh-pczajka: Reverted; hopefully we will remeber about that when converting the errors :)

python/src/snowflake/connector/cursor.py
sfc-gh-pczajka: done

sfc-gh-turbaszek (approved):

PR #270, opened 2026-02-11
sf_core: replace per-call tokio runtime creation with a shared global runtime

sfc-gh-boler:
Each call to connection_init(), statement_execute_query(), and get_chunk_data_sync() previously created and tore down a full multi-threaded tokio runtime. This was especially wasteful for chunk downloads.

Introduce sf_core/src/async_bridge.rs, which holds a single lazily-initialized tokio runtime. Initialization returns a Result to avoid panics across FFI boundaries.

The CRL background worker runtimes (crl/worker.rs, crl/cache.rs) are intentionally left unchanged -- they run infinite event loops on dedicated threads with single-threaded runtimes and serve a different purpose.

### DIFF
--- a/sf_core/src/async_bridge.rs
+++ b/sf_core/src/async_bridge.rs
@@ -1,0 +1,29 @@
+use std::sync::OnceLock;
+
+static RUNTIME: OnceLock<tokio::runtime::Runtime> = OnceLock::new();
+
+/// Get the global async runtime, initializing on first access.
+///
+/// # Notes
+/// The returned runtime must not be used with `block_on` from within
+/// an existing tokio runtime context (e.g. `#[tokio::test]`), as that
+/// will panic.
+pub(crate) fn runtime() -> Result<&'static tokio::runtime::Runtime, std::io::Error> {
+    if let Some(rt) = RUNTIME.get() {
+        return Ok(rt);
+    }
+    // Build may be called more than once if multiple threads race here,
+    // but `get_or_init` guarantees only one value is stored; the extra
+    // runtime is simply dropped.
+    let rt = build()?;
+    Ok(RUNTIME.get_or_init(|| rt))
+}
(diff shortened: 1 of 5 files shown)


sf_core/src/async_bridge.rs
```diff
@@ -0,0 +1,29 @@
+use std::sync::OnceLock;
+
+static RUNTIME: OnceLock<tokio::runtime::Runtime> = OnceLock::new();
```
sfc-gh-jszczerbinski, line 3: Should we use https://docs.rs/lazy_static/latest/lazy_static/ here?

sfc-gh-jszczerbinski (approved):
Long time overdue, consider using lazy_static macro

sf_core/src/async_bridge.rs
sfc-gh-boler: We are targeting a new enough version of rust where the external crate is not necessary. 

I guess we could use LazyLock, but then we would be back to panicing in the (unlikely) case of a failed initializatoin.

sfc-gh-jszczerbinski: We are already using it and it's small dependency. I think the code is a little simpler when using it:

```rust

lazy_static! {   
    static ref runtime : Result<tokio::runtime::Runtime, std::io::Error> =  {
        tokio::runtime::Builder::new_multi_thread()
           .enable_all()
           .thread_name("sf-core-async-bridge-worker")        
           .build()
    }
}
```

sfc-gh-jszczerbinski: I think either way is fine.

sfc-gh-boler: I'll keep OnceLock and follow up with a cleanup PR to replace lazy_static with stdlib constructs.

PR #273, opened 2026-02-11
NO-SNOW: Add missing stubs for Python

sfc-gh-turbaszek:
## Summary

- Generate `.pyi` type stubs for protobuf-generated Python code by extending the proto generator to invoke `mypy-protobuf` (with fallback to `protoc --pyi_out`). This resolves mypy `attr-defined` errors on all imports from `database_driver_v1_pb2`.
- Add a hand-written `.pyi` stub for the Cython `ArrowStreamIterator` extension module, resolving the `import-not-found` error in `cursor.py`.
- Remove `# type: ignore` suppression comments from `cursor.py` and `connection.py` that are no longer needed, and fix two minor type annotation issues (`_rowcount` and `execute_result`).

## Details

**Proto generator** (`proto_generator/src/generators/python_generator.rs`):
- After generating `_pb2.py` with `protoc --python_out`, the generator now also produces a `.pyi` stub.
- Prefers `protoc --mypy_out` (via the `mypy-protobuf` plugin) which correctly handles proto3 `optional` fields, typed `HasField`/`WhichOneof` overloads, and includes docstrings.
- Falls back to `protoc --pyi_out` (built-in) if `mypy-protobuf` is not installed.

**Cython stub** (`arrow_stream_iterator.pyi`):
- Hand-written stub for the `ArrowStreamIterator` class declaring `__init__`, `__iter__`, and `__next__` signatures, matching the `.pyx` source.

**Type annotation fixes** (`cursor.py`):
- `_rowcount` typed as `int | None` (was assigned `None` without annotation).
- `execute_result` typed as `ExecuteResult | None` (was annotated as `ExecuteResult` but initialized to `None`).


sfc-gh-pczajka (approved):
LGTM

PR #284, opened 2026-02-13
SNOW-2399378 introduce perf tests with recorded http traffic

sfc-gh-akolodziejczyk:


tests/performance/drivers/python/app/connection.py
```diff
@@ -53,15 +59,27 @@ def execute_setup_queries(cursor, setup_queries):
...
+    legacy_path = Path(__file__).parent / "old_driver_src"
+    if not legacy_path.exists():
+        raise ImportError(f"Old driver directory not found: {legacy_path}")
+    
+    # Add old_driver_src to the beginning of sys.path to prioritize it
+    legacy_path_str = str(legacy_path.absolute())
+    if legacy_path_str not in sys.path:
+        sys.path.insert(0, legacy_path_str)
+    
+    # Remove any cached snowflake.connector modules to ensure fresh import
```
sfc-gh-pcyrek: Any better option than deleting modules? If it's a Dockerized, is it even necessary to clean?

sfc-gh-akolodziejczyk: Good point, removed it. As we run tests for new OR old driver, never both in the same container, it is not needed.

tests/performance/tests/test_select_1M_recorded_http.py
```diff
@@ -0,0 +1,70 @@
+"""Performance test for 1M rows with WireMock - for stability testing"""
+import pytest
+from runner.test_types import TestType
+
+ITERATIONS = 30
+WARMUP_ITERATIONS = 2
```
sfc-gh-pcyrek, line 6: Sometimes it is 1, sometimes it is 2, does it make a difference here?

sfc-gh-pcyrek (approved):
LGTM

tests/performance/tests/test_select_1M_recorded_http.py
sfc-gh-akolodziejczyk: for now 2 is for lower volume tests, 1 for bigger, but we will adjust it in the future

PR #286, opened 2026-02-13
SNOW-2314136: Logout session gherkin scenarios

sfc-gh-fpawlowski:
## Pull request overview

Adds new Gherkin feature specifications for “session logout” behavior across the shared test suite and driver-specific suites (Core, Python, JDBC, ODBC), intended as design-only scenario definitions.

**Changes:**
- Introduces shared logout scenarios (token cleanup, idempotent close, post-close invalidation, concurrency).
- Adds Python/JDBC/ODBC driver-specific logout behavior specs (defaults, timeouts, error strategies, phase-migration behavior).
- Adds a detailed Core HTTP-layer logout integration feature covering request construction, retry/timeout behavior, and strategy-driven error handling.


tests/definitions/core/session/logout.feature
```diff
@@ -0,0 +1,405 @@
...
+    And Total elapsed time is between 5 and 6 seconds
+
+  Scenario: should cancel individual request when per-request socket timeout exceeded
+    # Tests that per-request timeout is passed to socket and interrupts slow responses
+    Given Mock HTTP server holds connection open for 8 seconds on first attempt then succeeds immediately
+    And UD Core connection is logged in
+    And Per-request socket timeout is set to 2 seconds
+    And Total retry budget timeout is set to 10 seconds
+    When Logout is initiated
+    Then First request is cancelled after 2 seconds due to socket timeout
```
sfc-gh-boler: What exactly is a "socket timeout" in this case and how is it configured?

tests/definitions/python/session/logout.feature
```diff
@@ -0,0 +1,176 @@
...
+    And Session is logged out if conditions allow
+
+  Scenario: should emit deprecation warning on first auto-cleanup run per process
+    # Phase 1 (doc for: SNOW-2314152) deprecation. Prepares users for explicit close() requirement.
+    Given Snowflake Python client is created with auto_cleanup enabled
+    And No auto-cleanup has run yet in this process
+    When Process exits without explicit close
+    Then atexit handler runs
+    And Deprecation warning is emitted once
+    When Another connection is created and process exits
```
sfc-gh-boler: With two "When" blocks it's a bit hard to understand when the second When/Then pair happens.

tests/definitions/shared/session/logout.feature
```diff
@@ -0,0 +1,69 @@
...
+  #                    Post-Logout Session Invalidation
+  # ===========================================================================
+
+  Scenario: should reject queries client-side after connection is closed
+    Given Snowflake client is logged in
+    And Simple query SELECT 1 executes successfully
+    When Connection is closed
+    And Query is attempted on closed connection
+    Then Query throws ConnectionClosedException
+    And Error message contains "Connection is closed"
```
sfc-gh-boler: Don't think we can require that exact wording from all drivers, because it may be part of the language-dependent interface. 

The same applies to `ConnectionClosedException`. Instead of "Query throws ConnectionClosedException", how about "The query fails with a connection-closed error"?

tests/definitions/core/session/logout.feature
```diff
@@ -0,0 +1,405 @@
...
+  # ===========================================================================
+
+  Scenario: should timeout after 5 seconds by default when server does not respond
+    # Tests that default timeout is applied when no override provided
+    # Mock server holds connection open (10s) to verify timeout interrupts after 5s
+    Given Mock HTTP server holds connection open for 10 seconds without responding
+    And UD Core connection is logged in with no timeout override
+    When Logout is initiated
+    Then Logout request times out after approximately 5 seconds
+    And Close throws timeout error
```
sfc-gh-boler: I don't see a language-specific override for this one, but it would be a breaking change for some drivers. This gherkin needs to support strategies?

tests/definitions/python/session/logout.feature
```diff
@@ -0,0 +1,176 @@
...
+  Scenario: should have enable_server_session_keep_alive_auto_detection default to true
+    # Phase 2 (doc for: SNOW-2314152) default for backward compatibility. Phase 3 defaults to false.
+    Given Snowflake Python client is created without enable_server_session_keep_alive_auto_detection parameter
+    When Connection configuration is checked
+    Then enable_server_session_keep_alive_auto_detection defaults to true
+    And Auto-detection is enabled by default
+
+  Scenario: should use best-effort error handling strategy by default
+    Given Snowflake Python client is created with default parameters
+    And Server will return 500 Internal Server Error on logout
```
sfc-gh-boler: We need to either constrain it to disable retries or clarify that the mock server must return only 500s. Otherwise we may: call close(), get 500, retry, get 200, close succeeds.

tests/definitions/shared/session/logout.feature
```diff
@@ -0,0 +1,69 @@
...
+
+  # Core-level HTTP protocol details are in core/session/logout.feature
+  # Auto-detection scenarios moved to fire-and-forget ticket (SNOW-2923705)
+  # Resource cleanup (heartbeat/telemetry/QCC) scenarios delegated to respective tickets
+
+  # ===========================================================================
+  #                          Token Cleanup
+  # ===========================================================================
+
+  Scenario Outline: should cleanup all tokens on close regardless of whether logout was sent
```
sfc-gh-boler, line 11: Why "Outline"?

tests/definitions/core/session/logout.feature
```diff
@@ -0,0 +1,405 @@
...
+    # Same logic: no retries → no point refreshing token
+    Given Core logout function called with best-effort strategy
+    And Mock HTTP server returns SESSION_TOKEN_EXPIRED 390112
+    And Retry policy allows 0 retries
+    When Logout is executed
+    Then No token refresh request is sent to server
+    And SESSION_TOKEN_EXPIRED is logged as WARN
+    And Close succeeds
+
+  Scenario Outline: should attempt token refresh on 390112 when retries allowed for each <strategy_type>
```
sfc-gh-boler, line 203: Why "Outline"?

tests/definitions/shared/session/logout.feature
```diff
@@ -0,0 +1,69 @@
...
+    # Tests that tokens are cleared regardless of logout decision
+    Given Snowflake client is logged in
+    And <server_session_keep_alive> is set to any value
+    When Connection is closed
+    Then Session token in Connection.tokens is null
```
sfc-gh-boler, line 16: How are we going to inspect that?

sfc-gh-boler (changes requested):
Meta point: it would be great to link to the relevant design docs and specs. Without that, understanding and validating the gherkins in hard. :)

tests/definitions/core/session/logout.feature
sfc-gh-fpawlowski: This was supposed to only check core's default. Each driver can override it - their respective timeouts are checked in the integration-manner - e.g.:
1. JDBC: https://github.com/snowflakedb/universal-driver/pull/286/changes#diff-95c6052390e6e32e7d631f71c6c43fdf568241e7491aa5c8ced22bebcecf80faR8
2. Python: https://github.com/snowflakedb/universal-driver/pull/286/changes#diff-7b0f82a91e8943c5470440bc66e3757b3872428f465ecdf8343c58a0c92a99a7R8

tests/definitions/core/session/logout.feature
sfc-gh-fpawlowski: Socket timeout was config exposed and configurable in the old JDBC driver. It was the Apache HttpClient setting controlling how long to wait for data between TCP packets on an established connection.

However, after creating an initial implementation of this logout feature, I believe tests and introduction of such potential socketTimeout should be tackled in: https://snowflakecomputing.atlassian.net/browse/SNOW-2314153

I decided that for now we should only honor the total timeout for logout and retry count max - as socket Timeout was not in the initial design doc scope as well.

tests/definitions/core/session/logout.feature
sfc-gh-fpawlowski: _Scenario Outline_ is gherkin's syntax for parametrization. The _Examples_ block below is recognised and each column's name appearing as <col_name> in gherkin's body should be treated as any value from that column.

tests/definitions/python/session/logout.feature
sfc-gh-fpawlowski: Good catch! Added 👍

tests/definitions/python/session/logout.feature
sfc-gh-fpawlowski: Agreed, rewritten to keep single "When" block

tests/definitions/shared/session/logout.feature
sfc-gh-fpawlowski: As earlier - we want to test as parametrised.

tests/definitions/shared/session/logout.feature
sfc-gh-fpawlowski: Yes, that's my oversight - I locally asked claude to change that to error type being returned with protobuf - but seems like it got stashed and not committed. Fixed now 👍

tests/definitions/shared/session/logout.feature
sfc-gh-fpawlowski: I intended to use ConnectionGetInfoResponse -> so effectively in rust the connection_get_info function.

sfc-gh-boler: Why does the wrapper need to see those tokens at all? I assumed they would stay within the core and wouldn't be exposed via ConnectionInfo.

tests/definitions/core/session/logout.feature
sfc-gh-boler: Thanks!

tests/definitions/core/session/logout.feature
sfc-gh-boler: Agreed! Can you update the PR and remove the gherkin?

tests/definitions/core/session/logout.feature
```diff
@@ -0,0 +1,405 @@
...
+    And Close completes successfully after logout response arrives
+
+  Scenario: should fail in-flight query when server response arrives after closing process started
+    # The server completes the query — the HTTP connection is not cancelled.
+    # The query fails because post-response processing cannot operate on
+    # invalidated services after close.
+    Given Mock HTTP server delays query response by 3 seconds then returns query result
+    And Mock HTTP server accepts logout requests with 200
+    And UD Core connection is logged in
+    And Socket timeout is set to 10 seconds
```
sfc-gh-boler: Per the earlier thread, we shouldn't use socket timeouts anymore.

But can you explain how the 3 seconds delay and 10 seconds timeout work in this test? As it is, the test seems slow and brittle.

tests/definitions/core/session/logout.feature
```diff
@@ -0,0 +1,405 @@
...
+
+  Scenario: should timeout after 5 seconds by default when server does not respond
+    # Tests that default timeout is applied when no override provided
+    # Mock server holds connection open (10s) to verify timeout interrupts after 5s
+    Given Mock HTTP server holds connection open for 10 seconds without responding
+    And UD Core connection is logged in with no timeout override
+    When Logout is initiated
+    Then Logout request times out after approximately 5 seconds
+    And Close throws timeout error
+    And Total elapsed time is between 5 and 6 seconds
```
sfc-gh-boler: This check is probably redundant. If we return with the timeout error, it must have been the timeout that triggered. Bound it to 1 second window will cause flaky tests.

tests/definitions/core/session/logout.feature
```diff
@@ -0,0 +1,405 @@
...
+      | best-effort   | 3            | 1        | 2                 |
+      | strict        | 5            | 4        | 5                 |
+      | best-effort   | 5            | 4        | 5                 |
+
+  Scenario Outline: should honor provided timeout config and succeed for each <strategy_type>
+    # Wrappers pass their historical defaults (Python: 5s, JDBC/ODBC: 300s)
+    # Note: Failure path scenarios (timeout exceeded) are below, split by strategy
+    Given Core logout function called with <strategy_type> strategy
+    And Timeout configured to <timeout_seconds> seconds
+    And Mock HTTP server delays response by <delay_seconds> seconds then returns 200
```
sfc-gh-boler, line 277: What's the point of delaying the request if the delay is always shorter than the timeout?

tests/definitions/core/session/logout.feature
```diff
@@ -0,0 +1,405 @@
...
+      | 3            |
+
+  # -- Failure path: timeout (outcome differs per strategy) --
+
+  Scenario Outline: should throw on timeout with strict strategy
+    Given Core logout function called with strict strategy
+    And Timeout configured to <timeout_seconds> seconds
+    And Mock HTTP server delays response by <delay_seconds> seconds
+    When Logout is executed
+    Then Request times out after <timeout_seconds> seconds
```
sfc-gh-boler: Is this some observable behaviour or are we only observing "Close throws timeout error"?

tests/definitions/core/session/logout.feature
```diff
@@ -0,0 +1,405 @@
...
+      | timeout_seconds | delay_seconds |
+      | 3               | 5             |
+      | 5               | 10            |
+
+  Scenario Outline: should log WARN and succeed on timeout with best-effort strategy
+    Given Core logout function called with best-effort strategy
+    And Timeout configured to <timeout_seconds> seconds
+    And Mock HTTP server delays response by <delay_seconds> seconds
+    When Logout is executed
+    Then Request times out after <timeout_seconds> seconds
```
sfc-gh-boler: Is this some observable behaviour or are we only observing "Close throws timeout error"?

tests/definitions/core/session/logout.feature
```diff
@@ -0,0 +1,405 @@
...
+    #   Attempt 4: should never start (budget exhausted)
+    Given Mock HTTP server responds with 503 after 2 second delay on each attempt
+    And UD Core connection is logged in
+    And Total retry budget timeout is set to 5 seconds
+    # Any number above 3 should be sufficient for max retries
+    And Retry policy allows 10 attempts
+    When Logout is initiated
+    Then Fewer than 4 attempts are made
+    And The last attempt timeouts because remaining budget is less than server response time
+    And Total wall-clock time does not exceed 7 seconds for closing the connection
```
sfc-gh-boler: Similarly to one of the other test cases, this check exposes us to flaky test without bring much value. We already know that timeouts work because the test retried only 3 times, and then we get a timeout error.

tests/definitions/core/session/logout.feature
```diff
@@ -0,0 +1,405 @@
...
+    Given Core logout function called
+    And Mock HTTP server returns SESSION_TOKEN_EXPIRED 390112 on first attempt
+    And Token refresh endpoint delays response by 3 seconds
+    And Mock HTTP server returns 200 after token refresh
+    And Total retry budget timeout is set to 5 seconds
+    When Logout is executed
+    Then Token refresh is attempted
+    And Token refresh time is counted against total timeout budget
+    And Remaining budget for retry logout is reduced by token refresh duration
+    And Total wall-clock time does not exceed 7 seconds for closing the connection
```
sfc-gh-boler: This probably can be simplified to verifying that logout succeeded (after a retry). The wall-clock check seems unnecessary.

tests/definitions/core/session/logout.feature
```diff
@@ -0,0 +1,405 @@
...
+      | best-effort   | 5            | 4        | 5                 |
+
+  Scenario Outline: should honor provided timeout config and succeed for each <strategy_type>
+    # Wrappers pass their historical defaults (Python: 5s, JDBC/ODBC: 300s)
+    # Note: Failure path scenarios (timeout exceeded) are below, split by strategy
+    Given Core logout function called with <strategy_type> strategy
+    And Timeout configured to <timeout_seconds> seconds
+    And Mock HTTP server delays response by <delay_seconds> seconds then returns 200
+    When Logout is executed
+    Then Request completes within <timeout_seconds> seconds
```
sfc-gh-boler: This scenario effectively tests that we don't have more aggressive built in timeouts than what the user specifies. But that hard to test in general.

Wouldn't it be enough to confirm that logout and close succeeded?

tests/definitions/core/session/logout.feature
```diff
@@ -0,0 +1,405 @@
...
+    # Tests that default timeout is applied when no override provided
+    # Mock server holds connection open (10s) to verify timeout interrupts after 5s
+    Given Mock HTTP server holds connection open for 10 seconds without responding
+    And UD Core connection is logged in with no timeout override
+    When Logout is initiated
+    Then Logout request times out after approximately 5 seconds
+    And Close throws timeout error
+    And Total elapsed time is between 5 and 6 seconds
+
+  Scenario: should cancel individual request when per-request socket timeout exceeded
```
sfc-gh-boler: If we are removing socket timeouts, then this needs refactoring.

sfc-gh-boler (changes requested):
Overall this LGTM, but I had a closer look at the tests with dependency on wall-clock time and that's where the majority of comments is.

tests/definitions/shared/session/logout.feature
sfc-gh-fpawlowski: SessionToken is part of API that some drivers exposed in the past. E.g. Python driver lets users access tokens via connection.rest.token (all publicly accessible even in Python's convention).

Moreover - some of our internal libraries use drivers simply to manage the token's lifecycle - e.g. SnowAPIs.

tests/definitions/core/session/logout.feature
sfc-gh-fpawlowski: Ahh yes sure, seems like my local changes got lost and were not pushed at the end fully. My bad -reapplying!

tests/definitions/core/session/logout.feature
sfc-gh-fpawlowski: Right, removed 👍

tests/definitions/core/session/logout.feature
sfc-gh-fpawlowski: Should be true, removed as well.

tests/definitions/core/session/logout.feature
sfc-gh-fpawlowski: Resolved by relying on observations - split into 2 scenarios.

tests/definitions/core/session/logout.feature
sfc-gh-fpawlowski: The point of using multiple timeout values here was to make sure we are not just lucky with one particular number.
What I wanted these rows to guard against was not only the obvious “request shorter than timeout should succeed” behavior, but also hidden bugs in the timeout/retry logic where success happens to work for one chosen value while the underlying algorithm is still wrong. For example:
- the configured timeout could be ignored and replaced by some constant,
- timeout handling could accidentally depend on a cap/floor,
- retry-loop logic could interact with timeout budget in a way that only works for one specific timeout/delay ratio,
- or there could be some odd internal relationship that makes one example pass even though the general success path is not implemented correctly.

A concrete example is a buggy implementation that takes the total timeout and silently distributes it across the allowed retry attempts. In that case, timeouts may still appear to happen in roughly the “right” area, but for the wrong reason. For one chosen timeout/delay pair we might accidentally pass because the divided per-attempt budget still happens to be long enough, while a different configured timeout would expose that the algorithm is not actually treating timeout as the intended total budget. The point of using several timeout values was to reduce the chance that we are just matching one lucky ratio.

Also, the opposite side of the boundary is already covered by the timeout-failure scenarios like Scenario Outline: should throw on timeout with strict strategy and the corresponding best-effort one. So this success-side outline was meant to complement those by showing that values above the boundary still behave correctly, rather than trying to cover both sides in one place.

In the ideal world, I would cover this with a denser matrix across timeout values, retry budgets, and relevant edge conditions, so we could exercise those intersections much more explicitly. I just do not think that level of scope is necessary for this particular task/PR.

tests/definitions/core/session/logout.feature
sfc-gh-fpawlowski: For this scope I agree that's sufficient. Good point 👍

tests/definitions/core/session/logout.feature
sfc-gh-fpawlowski: I believed we could use 2 green threads (one with only sleep(timeout_seconds + medium size epsilon) and another with our execution) and have them both change atomic variable at their end. If our one was later - it would mean timeout is not respected. We could rely in such case on Tokio e.g. 

But again - I agree with you that this may not be need at this stage of the project. When we have the retries logic in the final state we may want to extend (and refine) our final network-testing approach.

tests/definitions/core/session/logout.feature
sfc-gh-fpawlowski: Fixed

sfc-gh-fpawlowski (commented):
Re-responded to comments - I think I messed something up in my local env first time answering them as the socket-timeout changes were already done in my local env, just not pushed. My bad - should be alright now.

tests/definitions/core/session/logout.feature
sfc-gh-fpawlowski: Removed before but somehow didnt push - my bad. Now really removed :))

tests/definitions/core/session/logout.feature
sfc-gh-boler: Let's say you have a hardcoded timeout or distributed across a number of attempts (btw, number of retries is not specified in the gherkin). If response delay is too low, then only some very low hardcoded timeouts will be caught by this. The test will catch more issues if the delay is close to timeout_seconds, e.g 5/4, 10/9, etc. Concretely, a hardcoded timeout of 15s will pass all listed scenarios, but if the delay is closer to the expected timeout, all scenarios will catch it. Similarly if timeout budget is distributed over retries.



BTW, the 300s delay will obviously make the tests slow, so we may want to scale it down a bit.



Food for thought!

sfc-gh-boler (approved):
lgtm

(by brain is friend from thinking about timeouts and delays so much)

PR #289, opened 2026-02-13
SNOW-2872454: Add config managment

sfc-gh-turbaszek:


sf_core/src/config/config_manager.rs
```diff
@@ -0,0 +1,856 @@
...
+    load_all_config_sections_with_options(true)
+}
+
+/// Apply environment variable overrides to settings based on section name
+///
+/// For connections (section starts with "connections."):
+///   - Uses SNOWFLAKE_<KEY> format (e.g., SNOWFLAKE_ACCOUNT)
+/// For other sections:
+///   - Uses SNOWFLAKE_<SECTION>_<KEY> format (e.g., SNOWFLAKE_LOG_LEVEL)
+fn apply_env_overrides_for_section(section_name: &str, settings: &mut HashMap<String, Setting>) {
```
sfc-gh-jszczerbinski: So can you elaborate why we need this? Is that a feature on the python side? How is it used?

sf_core/src/apis/database_driver_v1/connection.rs
```diff
@@ -7,15 +7,59 @@ use super::Handle;
...
+use crate::config::config_manager;
 use crate::config::rest_parameters::{ClientInfo, LoginParameters};
 use crate::config::retry::RetryPolicy;
 use crate::rest::snowflake::{self, RestError, SessionTokens, SnowflakeResponseError};
 use crate::tls::client::create_tls_client_with_config;
 use reqwest;
 
+/// Load configuration from TOML files for a named connection
+pub fn connection_load_from_config(
+    conn_handle: Handle,
```
sfc-gh-jszczerbinski: Pass mut & Connection so we don't lock connection twice

sf_core/src/apis/database_driver_v1/connection.rs
```diff
@@ -7,15 +7,59 @@ use super::Handle;
...
+    connection_name: &str,
+) -> Result<(), ApiError> {
+    // Load settings from config files
+    let config_settings = config_manager::load_connection_config(connection_name)
```
sfc-gh-jszczerbinski: This overrides env variables, what happens if we have multiple connections?

sf_core/src/c_api.rs
```diff
@@ -66,3 +71,133 @@ pub unsafe extern "C" fn sf_core_api_call_proto(
...
     }
 }
+
+/// Load configuration for a specific connection from TOML files
+/// Returns JSON serialized HashMap<String, Setting> on success
+/// # Safety
+/// This function dereferences raw pointers `connection_name`, `result`, and `result_len`.
+/// The caller must ensure that `connection_name`, `result`, and `result_len` are valid.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn sf_core_config_load_connection(
```
sfc-gh-jszczerbinski: Why not use protobuf here? This is not performance critical, we are not catching panic here on ffi boundary. This function also cannot return more complex errors

sf_core/src/config/config_manager.rs
sfc-gh-jszczerbinski, line 1: All those files seem python specific, can we put them in python mod inside of config mod?

sfc-gh-jszczerbinski (changes requested):
I don't think we should use C FFI for this API, can you make protobuf interface instead?

sf_core/src/config/config_manager.rs
sfc-gh-turbaszek: This is good question. This is feature on CLI which was never ported to driver. I can remove it from this PR to keep things simple. But this is frequently used feature on CI.

sfc-gh-jszczerbinski (approved):

PR #298, opened 2026-02-17
PY - Small bindings syntax compatibility with old driver

sfc-gh-pczajka:
* add `paramtyle` connection parameter
* unskip binding reference tests
----
PR stack:
* #265 
* #266 
* this PR


python/src/snowflake/connector/connection.py
```diff
@@ -22,17 +22,50 @@
...
+
+
+# Paramstyles that enable server-side binding in the universal driver.
+_SUPPORTED_PARAMSTYLES = {"qmark", "numeric"}
+
+# TODO: to be added in follow-up PR
+_CLIENT_SIDE_PARAMSTYLES = {"format", "pyformat"}
+
+
+def _resolve_paramstyle(value: str | None) -> str | None:
```
sfc-gh-turbaszek, line 35: Can we move it to `_internal` module?

sfc-gh-pczajka: It's addressed in #314

sfc-gh-turbaszek (approved):

PR #301, opened 2026-02-17
NO-SNOW: Add metadata support for PUT/GET

sfc-gh-turbaszek:


python/tests/e2e/put_get/test_put_get_basic_operations.py
```diff
@@ -158,25 +155,30 @@ def test_should_return_correct_column_metadata_for_put(connection):
...
-            "target_compression",
-            "status",
-            "message",
-        ]
 
-        for i, expected_name in enumerate(expected_columns):
-            actual_name = columns[i][0].lower()
-            assert actual_name == expected_name, f"Column {i} should be named '{expected_name}', got '{actual_name}'"
+        # Verify column names and type codes (TEXT=2, FIXED=0)
+        expected = [
```
sfc-gh-fpawlowski: nit: subjectively - naming was better before - expected_columns.

Same in the next test.

python/tests/e2e/put_get/test_put_get_basic_operations.py
```diff
@@ -158,25 +155,30 @@ def test_should_return_correct_column_metadata_for_put(connection):
...
+            ("target", 2),
+            ("source_size", 0),
+            ("target_size", 0),
+            ("source_compression", 2),
+            ("target_compression", 2),
+            ("status", 2),
+            ("message", 2),
+        ]
+        for i, (expected_name, expected_type_code) in enumerate(expected):
+            assert columns[i][0].lower() == expected_name, (
```
sfc-gh-fpawlowski: nit: accessing the name twice, instead of extracting to variable before like in the old code:
            actual_name = columns[i][0].lower()


Same thing below with type_code.
Same thing in the next test.

python/pyproject.toml
```diff
@@ -187,7 +188,7 @@ installer = "uv"
 check = [
     'ruff format --check {args:src/snowflake tests}',
     'ruff check {args:src/snowflake tests}',
-    'mypy --install-types --non-interactive -p snowflake.connector',
+    'mypy -p snowflake.connector',
```
sfc-gh-fpawlowski, line 191: nit: This should be already on main - seems Github got locked again on the wrong diff.

sf_core/src/apis/database_driver_v1/query.rs
```diff
@@ -132,11 +157,8 @@ macro_rules! int64_array {
     };
 }
 
-/// Converts upload results to Arrow format
-pub fn upload_results_reader(
-    upload_results: Vec<UploadResult>,
-) -> Result<Box<dyn RecordBatchReader + Send>, ArrowError> {
-    let row_types: Vec<RowType> = vec![
+fn upload_row_types() -> Vec<RowType> {
```
sfc-gh-fpawlowski, line 160: Just to be clear - we have to hardcode it in driver because the responses are in JSON -> we cannot infer any metadata about their fields?

sfc-gh-fpawlowski (approved):
LGTM, left some nits - mostly naming or clarification requests

sf_core/src/apis/database_driver_v1/query.rs
sfc-gh-turbaszek: For PUT/GET there's no metadata from server, so the driver builds it from scratch. This way executing PUT/GET query does not differ from other queries from end user perspective.

PR #309, opened 2026-02-17
SNOW-3111596: Add connection URL parsing

sfc-gh-rkowalski:
## Summary

Implement Snowflake JDBC behavior compatible connection URL parsing behavior in `universal-driver/jdbc`.

## What changed
- Added dedicated parser and resolver components for connection URL/options handling:
  - `ConnectionString.java`
  - `ConnectionOptionsResolver.java`
- Refactored `SnowflakeDriver` URL handling to match snowflake-jdbc behavior:
  - `acceptsURL` validates via parser
  - `connect` returns `null` for non-Snowflake prefixes and throws for invalid Snowflake URLs
- Refactored parser flow into fluent internal builder-style steps for readability.

### DIFF
--- a/jdbc/src/main/java/net/snowflake/client/api/driver/SnowflakeDriver.java
+++ b/jdbc/src/main/java/net/snowflake/client/api/driver/SnowflakeDriver.java
@@ -7,6 +7,8 @@ import java.sql.DriverPropertyInfo;
 import java.sql.SQLException;
 import java.util.Properties;
 import java.util.logging.Logger;
+import net.snowflake.client.api.exception.SnowflakeSQLException;
+import net.snowflake.client.internal.api.implementation.connection.ConnectionString;
 import net.snowflake.client.internal.api.implementation.connection.SnowflakeConnectionImpl;
 
 /**
@@ -21,6 +23,7 @@ public class SnowflakeDriver implements Driver {
   private static final String DRIVER_VERSION = "4.0.0";
   private static final int MAJOR_VERSION = 4;
   private static final int MINOR_VERSION = 0;
+  public static final Properties EMPTY_PROPERTIES = new Properties();
 
   public static void empty() {}
 
@@ -38,19 +41,19 @@ public class SnowflakeDriver implements Driver {
 
   @Override
   public Connection connect(String url, Properties info) throws SQLException {
-    if (!acceptsURL(url)) {
+    if (ConnectionString.hasSupportedPrefix(url)) {
       return null;
     }
-
+    ConnectionString parsed = ConnectionString.parse(url, info);
+    if (!parsed.isValid()) {
+      throw new SnowflakeSQLException("Connection string is invalid. Unable to parse.");
+    }
     return new SnowflakeConnectionImpl(url, info);
   }
 
   @Override
   public boolean acceptsURL(String url) throws SQLException {
-    if (url == null) {
-      return false;
-    }
-    return url.startsWith("jdbc:snowflake:");
+    return ConnectionString.parse(url, EMPTY_PROPERTIES).isValid();
   }
 
   @Override
--- a/jdbc/src/main/java/net/snowflake/client/api/driver/SnowflakeDriver.java
+++ b/jdbc/src/main/java/net/snowflake/client/api/driver/SnowflakeDriver.java
@@ -23,7 +23,6 @@ public class SnowflakeDriver implements Driver {
   private static final String DRIVER_VERSION = "4.0.0";
   private static final int MAJOR_VERSION = 4;
   private static final int MINOR_VERSION = 0;
-  public static final Properties EMPTY_PROPERTIES = new Properties();
 
   public static void empty() {}
 
@@ -41,7 +40,7 @@ public class SnowflakeDriver implements Driver {
 
   @Override
   public Connection connect(String url, Properties info) throws SQLException {
-    if (ConnectionString.hasSupportedPrefix(url)) {
+    if (ConnectionString.hasUnsupportedPrefix(url)) {
       return null;
     }
     ConnectionString parsed = ConnectionString.parse(url, info);
@@ -53,7 +52,7 @@ public class SnowflakeDriver implements Driver {
 
   @Override
   public boolean acceptsURL(String url) throws SQLException {
-    return ConnectionString.parse(url, EMPTY_PROPERTIES).isValid();
+    return ConnectionString.parse(url).isValid();
   }
 
   @Override
--- a/jdbc/src/main/java/net/snowflake/client/api/driver/SnowflakeDriver.java
+++ b/jdbc/src/main/java/net/snowflake/client/api/driver/SnowflakeDriver.java
@@ -12,7 +12,6 @@ import net.snowflake.client.internal.api.implementation.connection.ConnectionStr
 import net.snowflake.client.internal.api.implementation.connection.SnowflakeConnectionImpl;
 import net.snowflake.client.internal.log.SFLogger;
 import net.snowflake.client.internal.log.SFLoggerFactory;
-import org.slf4j.LoggerFactory;
 
 /**
  * Snowflake JDBC Driver implementation
@@ -26,7 +25,6 @@ public class SnowflakeDriver implements Driver {
   private static final String DRIVER_VERSION = "4.0.0";
   private static final int MAJOR_VERSION = 4;
   private static final int MINOR_VERSION = 0;
-  private static final org.slf4j.Logger log = LoggerFactory.getLogger(SnowflakeDriver.class);
 
   public static void empty() {}
 
@@ -57,7 +55,10 @@ public class SnowflakeDriver implements Driver {
 
   @Override
   public boolean acceptsURL(String url) throws SQLException {
-    return ConnectionString.parse(url).isValid();
+    if (url == null) {
+      throw new SQLException("URL must not be null");
+    }
+    return url.startsWith("jdbc:snowflake:");
   }
 
   @Override
--- a/jdbc/src/main/java/net/snowflake/client/api/driver/SnowflakeDriver.java
+++ b/jdbc/src/main/java/net/snowflake/client/api/driver/SnowflakeDriver.java
@@ -10,6 +10,9 @@ import java.util.logging.Logger;
 import net.snowflake.client.api.exception.SnowflakeSQLException;
 import net.snowflake.client.internal.api.implementation.connection.ConnectionString;
 import net.snowflake.client.internal.api.implementation.connection.SnowflakeConnectionImpl;
+import net.snowflake.client.internal.log.SFLogger;
+import net.snowflake.client.internal.log.SFLoggerFactory;
+import org.slf4j.LoggerFactory;
 
 /**
  * Snowflake JDBC Driver implementation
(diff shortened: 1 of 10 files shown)


jdbc/src/main/java/net/snowflake/client/api/driver/SnowflakeDriver.java
```diff
@@ -38,19 +40,19 @@ public static void registerDriver() {
...
     return new SnowflakeConnectionImpl(url, info);
   }
 
   @Override
   public boolean acceptsURL(String url) throws SQLException {
-    if (url == null) {
-      return false;
-    }
-    return url.startsWith("jdbc:snowflake:");
+    return ConnectionString.parse(url).isValid();
```
sfc-gh-pfus: This is a bit of change. IIUC, this function checks if this URL is for YOUR database, not that you can connect with it.

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionOptionsResolver.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    for (Map.Entry<String, Object> entry : parsed.getParameters().entrySet()) {
+      String normalizedKey = entry.getKey().toLowerCase(Locale.ROOT);
+      if (normalizedKey.trim().isEmpty()) {
+        continue;
+      }
+      Object value = entry.getValue();
+      if (value == null) {
+        continue;
```
sfc-gh-pfus: 100% sure? Nulls are not valid (existing of a key doesn't has a notion?)

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
```diff
@@ -0,0 +1,256 @@
...
+import java.io.UnsupportedEncodingException;
+import java.net.URI;
+import java.net.URLDecoder;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Properties;
+
+public class ConnectionString implements Serializable {
```
sfc-gh-pfus: Why do we need serializable?

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
```diff
@@ -0,0 +1,256 @@
...
+          .validateSchemeHostAndPath()
+          .applyQueryParameters()
+          .normalizeScheme()
+          .applyInfoOverrides()
+          .deriveAccountFromHostIfMissing()
+          .ensureAccountPresent()
+          .normalizeHostForUnderscoreAccount()
+          .build();
+    } catch (Exception ex) {
+      return INVALID_CONNECT_STRING;
```
sfc-gh-pfus, line 74: No error printing, nothing is returned to the caller?

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
```diff
@@ -0,0 +1,256 @@
...
+    }
+  }
+
+  public boolean isValid() {
+    return !isNullOrEmpty(host);
```
sfc-gh-pfus, line 79: Is this good enough?

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
```diff
@@ -0,0 +1,256 @@
...
+
+  public Map<String, Object> getParameters() {
+    return Collections.unmodifiableMap(parameters);
+  }
+
+  public String getAccount() {
+    return account;
+  }
+
+  private static boolean isNullOrEmpty(String value) {
```
sfc-gh-pfus: Let's use an existing library or create a dedicated class.

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
```diff
@@ -0,0 +1,256 @@
...
+    return value == null || value.isEmpty();
+  }
+
+  private static String stringValue(Object value) {
+    return value == null ? null : value.toString();
```
sfc-gh-pfus: `String.valueOf` is not good enough?

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
```diff
@@ -0,0 +1,256 @@
...
+
+    ConnectionStringBuilder validateSchemeHostAndPath() {
+      if (!"snowflake".equals(scheme) && !"http".equals(scheme) && !"https".equals(scheme)) {
+        throw new IllegalArgumentException("Unsupported scheme");
+      }
+      if (isNullOrEmpty(host)) {
+        throw new IllegalArgumentException("Missing host");
+      }
+      if (port == -1) {
+        port = 443;
```
sfc-gh-pfus: I think it should be based on the protocol? And by the way - function name says `validate`, but it mutates the state 🙃

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
```diff
@@ -0,0 +1,256 @@
...
+      }
+      String path = uri.getPath();
+      if (!isNullOrEmpty(path) && !"/".equals(path)) {
+        throw new IllegalArgumentException("Invalid path");
```
sfc-gh-pfus, line 162: Is it not possible to choose a db or schema in path?

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
```diff
@@ -0,0 +1,256 @@
...
+    ConnectionStringBuilder applyQueryParameters() {
+      String queryData = uri.getRawQuery();
+      if (isNullOrEmpty(queryData)) {
+        return this;
+      }
+      String[] params = queryData.split("&");
+      for (String param : params) {
+        String[] keyVals = param.split("=", 2);
+        if (keyVals.length != 2 || keyVals[1].isEmpty()) {
+          continue;
```
sfc-gh-pfus, line 176: Same - is empty value not a valid value?

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
```diff
@@ -0,0 +1,256 @@
...
+        }
+        String key;
+        String value;
+        try {
+          key = URLDecoder.decode(keyVals[0], "UTF-8");
```
sfc-gh-pfus: Actually you could initialize key and value in `try` and call `applyParameter` too.

jdbc/src/main/java/net/snowflake/client/api/driver/SnowflakeDriver.java
sfc-gh-rkowalski: I'm not sure what JDBC specs says about it but thats how we behave in snowflake-jdbc https://github.com/snowflakedb/snowflake-jdbc/blob/e42f811367445328ee2372a903a9a41f437f09ef/src/main/java/net/snowflake/client/api/driver/SnowflakeDriver.java#L71

sfc-gh-rkowalski: > The java.sql.Driver.acceptsURL(String url) method checks if a JDBC driver can establish a connection to a specific database URL. It returns true if the driver recognizes the URL's sub-protocol (e.g., jdbc:postgresql: or jdbc:oracle:) and false otherwise. It is primarily used to identify appropriate drivers for a given connection string.

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
sfc-gh-rkowalski: We had this in snowflake-jdbc. I'm not sure if this was added there intentionally or not but I can remove it now and we'll add this in the future if needed

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
sfc-gh-rkowalski: I guess not but thats how snowflake-jdbc behaves 😄

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
sfc-gh-rkowalski: Since logger PR is merged I can add logging now 😄 but probably should've added todo here

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
sfc-gh-rkowalski: Added to `SnowflakeUtil`

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
sfc-gh-rkowalski: It is!

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionOptionsResolver.java
sfc-gh-rkowalski: Good to call that out, there shouldn't be nulls here

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
sfc-gh-rkowalski: This is legacy behavior is: no explicit port is forced to 443, even if final scheme becomes http 🥲
It doesn't make sense TBH, maybe we should change it?

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
sfc-gh-rkowalski: Yes, this is snowflake-jdbc behavior

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
sfc-gh-rkowalski: Empty query parameter values are ignored for compatibility with existing snowflake-jdbc behavior

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
sfc-gh-rkowalski: Right, inlined

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
sfc-gh-pfus: Maybe let's not create a huge bag for everything again. Maybe StringUtil?

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/ConnectionString.java
sfc-gh-pfus: We should, it doesn't make sense and won't work.

sfc-gh-pfus (approved):

PR #300, opened 2026-02-17
SNOW-3103846 refactor int.feature tests to us parametrized tests, support showing parametrized tests in html report

sfc-gh-akolodziejczyk:

PR #317, opened 2026-02-18
Pczajka/add test for different batch size handling

sfc-gh-pczajka:
Follow up of #300 - add test case for glueing batches of different variable size

### DIFF
--- a/odbc_tests/tests/e2e/types/int.cpp
+++ b/odbc_tests/tests/e2e/types/int.cpp
@@ -131,46 +131,7 @@ TEST_CASE("should select values from table for int and synonyms", "[int]") {
   }
 }
 
-TEST_CASE("should select large result set from table for int and synonyms", "[int]") {
-  // Given Snowflake client is logged in
-  Connection conn;
-  auto random_schema = Schema::use_random_schema(conn);
-
-  // And Table with <type> column exists with 50000 sequential values
-  conn.execute("DROP TABLE IF EXISTS int_large_table");
-  conn.execute("CREATE TABLE int_large_table (col BIGINT)");
-  conn.execute("INSERT INTO int_large_table SELECT seq8() FROM TABLE(GENERATOR(ROWCOUNT => 50000))");
-
-  // When Query "SELECT * FROM <table> ORDER BY col" is executed
-  auto stmt = conn.createStatement();
-  const auto sql = "SELECT * FROM int_large_table ORDER BY col";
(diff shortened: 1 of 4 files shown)


odbc_tests/tests/e2e/types/int.cpp
```diff
@@ -169,3 +169,54 @@ TEST_CASE("should select large result set from table for int and synonyms", "[in
 
   REQUIRE(row_count == 50000);
 }
+
+TEST_CASE("should manage different column sizes between batches for int and synonyms", "[int]") {
```
sfc-gh-akolodziejczyk: test looks good, but I think it would be good to run it few times and verify batches look how we expect, can you do that (i.e. by adding some logs for testing)

python/tests/e2e/types/test_int.py
```diff
@@ -289,6 +289,33 @@ def test_should_select_large_result_set_from_table_for_int_and_synonyms(self, ex
...
 
+    def test_should_handle_server_side_arrow_memory_optimization_for_int_columns(self, execute_query, tmp_schema):
+        # Given Snowflake client is logged in
+        assert_connection_is_open(execute_query)
+
+        # And Table with four INT columns exists
+        table_name = f"{tmp_schema}.different_int_column_sizes"
+        execute_query(f"CREATE TABLE {table_name} (col_int8 INT, col_int16 INT, col_int32 INT, col_int64 INT)")
+
+        # And Each column contains values of different magnitudes
```
sfc-gh-akolodziejczyk: we have this step empty for both python and odbc, can we adjust the definition?

tests/definitions/shared/types/int.feature
```diff
@@ -80,6 +80,21 @@ Feature: INT type support
     When Query "SELECT * FROM <table> ORDER BY col" is executed
     Then Result should contain 50000 sequentially numbered rows from 0 to 49999
 
+  @python_e2e @odbc_e2e
```
sfc-gh-akolodziejczyk: Maybe you can use Scenario Outline syntax?

sfc-gh-pczajka: I'm not sure I understand the difference. I can see that's it's used when test has multiple cases - which is not a case here

python/tests/e2e/types/test_int.py
sfc-gh-pczajka: done

odbc_tests/tests/e2e/types/int.cpp
sfc-gh-pczajka: done

sfc-gh-rkowalski (approved):

PR #323, opened 2026-02-19
Add basic execute query tests for Python

sfc-gh-pczajka:
SNOW-2997744 will be fixed in follow-up PR (and tests should be un-skipped)

### DIFF
--- a/python/tests/e2e/query/test_basic_execute_query.py
+++ b/python/tests/e2e/query/test_basic_execute_query.py
@@ -1,0 +1,204 @@
+"""Basic execute query tests for Universal Driver (Python-specific).
+
+This module tests basic query execution functionality including:
+- Simple SELECT queries returning single value
+- SELECT queries returning multiple columns
+- SELECT queries returning multiple rows
+- SELECT queries returning empty result set
+- SELECT queries returning NULL values
+- DDL statements (CREATE and DROP TABLE)
+- DML statements (INSERT and SELECT)
+- Error handling for invalid SQL
+- Sequential query execution on same connection
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from tests.e2e.types.utils import assert_sequential_values
+
+from ...compatibility import IS_UNIVERSAL_DRIVER
+
+
+class TestSelectQueries:
+    """Tests for basic SELECT query execution."""
+
+    def test_should_execute_simple_SELECT_returning_single_value(self, cursor):
+        """Test simple SELECT returning single value."""
+        # Given Snowflake client is logged in
+
+        # When Query "SELECT 1 AS value" is executed
+        cursor.execute("SELECT 1 AS value")
+        result = cursor.fetchone()
+
+        # Then the result should contain value 1
+        assert result is not None
+        assert result[0] == 1
+
+    def test_should_execute_SELECT_returning_multiple_columns(self, cursor):
+        """Test SELECT returning multiple columns."""
+        # Given Snowflake client is logged in
+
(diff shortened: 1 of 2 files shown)


sfc-gh-turbaszek (approved):

python/tests/e2e/query/test_basic_execute_query.py
```diff
@@ -0,0 +1,204 @@
...
+        cursor.execute(f"SELECT COUNT(*) FROM {table_name}")
+        result = cursor.fetchone()
+        assert result[0] == 0
+
+        # And DROP TABLE statement should complete successfully
+        cursor.execute(f"DROP TABLE {table_name}")
+
+        # Verify table no longer exists by attempting to query it
+        if IS_UNIVERSAL_DRIVER:
+            from snowflake.connector._internal.protobuf_gen.proto_exception import ProtoApplicationException
```
sfc-gh-fpawlowski: Do we want to name this behaviour difference?

python/tests/e2e/query/test_basic_execute_query.py
```diff
@@ -0,0 +1,204 @@
...
+
+            expected_error = ProtoApplicationException
+        else:
+            from snowflake.connector import ProgrammingError
+
+            expected_error = ProgrammingError
+
+        with pytest.raises(expected_error):
```
sfc-gh-fpawlowski: If we are asserting that table no longer exists by such call we may want to check that error code received means actually that. We might have received other issues like ROLE-based ones and unawarely suppose it means it was deleted. 

NIT: We could run smth like SHOW TABLE LIKE... or select tables from information_schema like ....

python/tests/e2e/query/test_basic_execute_query.py
```diff
@@ -0,0 +1,204 @@
...
+class TestErrorHandling:
+    """Tests for error handling."""
+
+    def test_should_return_error_for_invalid_sql_syntax(self, cursor):
+        """Test error handling for invalid SQL syntax."""
+        # Given Snowflake client is logged in
+
+        # When Invalid SQL "SELCT INVALID SYNTAX" is executed
+        # Then An error should be returned
+        if IS_UNIVERSAL_DRIVER:
```
sfc-gh-fpawlowski, line 154: Same as earlier - should this BehaviourDifference be named?

python/tests/e2e/query/test_basic_execute_query.py
sfc-gh-pczajka: This is not a desired state - to be changes after error unification PR. I'll leave a TODO

python/tests/e2e/query/test_basic_execute_query.py
sfc-gh-pczajka: I think that would oppose the gherkin scenario - "# Verify table no longer exists by attempting to query it"

python/tests/e2e/query/test_basic_execute_query.py
sfc-gh-pczajka: Same - This is not a desired state - to be changes after error unification PR. I'll leave a TODO

python/tests/e2e/query/test_basic_execute_query.py
sfc-gh-fpawlowski: Isn't the:
> And DROP TABLE statement should complete successfully

the last step in the gherkin scenario?

And 'by attempting to query it' is just the chosen way of checking that?

sfc-gh-fpawlowski (approved):

PR #324, opened 2026-02-19
SNOW-2467847: Implement JDBC query parameter binding

sfc-gh-rkowalski:
## Summary
- Adds query parameter binding support for JDBC prepared statements.
- Introduces prepared statement binding serialization and wires it into statement execution/transport flow.
- Extends and updates unit/integration coverage for binding behavior and related execution paths.
- NO batch support yet, this will be in additional PR since this one got pretty big already

## Testing
- Added/updated tests for:
  - Prepared statement binding behavior.
  - Unsupported feature handling for prepared statements.
  - Statement and result-set related paths impacted by binding support.
  - End-to-end integer type scenarios in shared feature definitions.

### DIFF
--- a/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
+++ b/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
@@ -123,7 +123,7 @@ final class PreparedStatementBindingSerializer {
       this.address = address;
     }
 
-    static NativeBuffer fromBytes(byte[] source) throws SQLException {
+    private static NativeBuffer fromBytes(byte[] source) throws SQLException {
       RootAllocator allocator = null;
       ArrowBuf arrowBuf = null;
       boolean success = false;
--- a/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
+++ b/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
@@ -1,0 +1,155 @@
+package net.snowflake.client.internal.api.implementation.statement;
+
+import com.google.protobuf.ByteString;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.charset.StandardCharsets;
+import java.sql.SQLException;
+import net.snowflake.client.internal.log.SFLogger;
+import net.snowflake.client.internal.log.SFLoggerFactory;
+import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.BinaryDataPtr;
+import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.QueryBindings;
+import org.apache.arrow.memory.ArrowBuf;
+import org.apache.arrow.memory.RootAllocator;
+import org.json.JSONStringer;
+
+final class PreparedStatementBindingSerializer {
+  private static final SFLogger logger =
+      SFLoggerFactory.getLogger(PreparedStatementBindingSerializer.class);
+
+  static final class ParameterValue {
+    private final String bindType;
+    private final Object value;
+
+    ParameterValue(String bindType, Object value) {
+      this.bindType = bindType;
+      this.value = value;
+    }
+
+    String bindType() {
+      return bindType;
+    }
+
+    Object value() {
+      return value;
+    }
+  }
+
+  static final class SerializedBindings implements AutoCloseable {
+    private final QueryBindings bindings;
+    private final NativeBuffer buffer;
+
+    SerializedBindings(QueryBindings bindings, NativeBuffer buffer) {
+      this.bindings = bindings;
+      this.buffer = buffer;
+    }
+
+    QueryBindings bindings() {
+      return bindings;
+    }
+
+    @Override
+    public void close() {
+      if (buffer != null) {
+        buffer.close();
+      }
+    }
+  }
+
+  private PreparedStatementBindingSerializer() {}
+
+  static SerializedBindings serialize(ParameterValue[] parameterValues) throws SQLException {
+    if (parameterValues.length == 0) {
+      logger.debug("No parameter placeholders found, skipping bindings serialization.");
+      return new SerializedBindings(null, null);
+    }
+    logger.debug("Serializing prepared bindings: placeholders={}", parameterValues.length);
+
+    JSONStringer jsonStringer = new JSONStringer();
+    jsonStringer.object();
+    for (int i = 0; i < parameterValues.length; i++) {
+      ParameterValue parameterValue = parameterValues[i];
+      int parameterIndex = i + 1;
+      if (parameterValue == null) {
+        logger.warn(
+            "Bindings serialization failed: missing parameter value for index {}", parameterIndex);
+        throw new SQLException("Missing value for parameter index: " + parameterIndex);
+      }
+
+      jsonStringer.key(String.valueOf(parameterIndex)).object();
+      jsonStringer.key("type").value(parameterValue.bindType());
+      if (parameterValue.value() == null) {
+        jsonStringer.key("value").value(null);
+      } else {
+        jsonStringer.key("value").value(String.valueOf(parameterValue.value()));
+      }
+      jsonStringer.endObject();
+    }
+    jsonStringer.endObject();
+
+    byte[] jsonBytes = jsonStringer.toString().getBytes(StandardCharsets.UTF_8);
+    NativeBuffer nativeBuffer = NativeBuffer.fromBytes(jsonBytes);
+    boolean success = false;
+    try {
+      byte[] ptrBytes = nativeBuffer.pointerAsLittleEndianBytes();
+      BinaryDataPtr jsonPtr =
+          BinaryDataPtr.newBuilder()
+              .setValue(ByteString.copyFrom(ptrBytes))
+              .setLength(jsonBytes.length)
+              .build();
+      QueryBindings queryBindings = QueryBindings.newBuilder().setJson(jsonPtr).build();
+      logger.debug(
+          "Prepared bindings serialized: payloadBytes={}, pointerBytes={}",
+          jsonBytes.length,
+          ptrBytes.length);
+      SerializedBindings serializedBindings = new SerializedBindings(queryBindings, nativeBuffer);
+      success = true;
+      return serializedBindings;
+    } finally {
+      if (!success) {
+        nativeBuffer.close();
+      }
+    }
+  }
+
+  private static final class NativeBuffer implements AutoCloseable {
+    private final RootAllocator allocator;
+    private final ArrowBuf arrowBuf;
+    private final long address;
+
+    private NativeBuffer(RootAllocator allocator, ArrowBuf arrowBuf, long address) {
+      this.allocator = allocator;
+      this.arrowBuf = arrowBuf;
+      this.address = address;
+    }
+
+    static NativeBuffer fromBytes(byte[] source) throws SQLException {
+      RootAllocator allocator = new RootAllocator(Long.MAX_VALUE);
+      ArrowBuf arrowBuf = allocator.buffer(source.length);
+      arrowBuf.setBytes(0, source);
+      long address = arrowBuf.memoryAddress();
+      if (address == 0L) {
+        arrowBuf.close();
+        allocator.close();
+        logger.warn(
+            "Failed to allocate native memory for binding data: payloadBytes={}", source.length);
+        throw new SQLException("Failed to allocate native memory for binding data");
+      }
+      logger.debug("Allocated native binding buffer: payloadBytes={}", source.length);
+      return new NativeBuffer(allocator, arrowBuf, address);
+    }
+
+    byte[] pointerAsLittleEndianBytes() {
+      return ByteBuffer.allocate(Long.BYTES)
+          .order(ByteOrder.LITTLE_ENDIAN)
+          .putLong(address)
+          .array();
+    }
+
+    @Override
+    public void close() {
+      arrowBuf.close();
+      allocator.close();
+    }
+  }
+}
--- a/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
+++ b/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
@@ -124,19 +124,32 @@ final class PreparedStatementBindingSerializer {
     }
 
     static NativeBuffer fromBytes(byte[] source) throws SQLException {
-      RootAllocator allocator = new RootAllocator(Long.MAX_VALUE);
-      ArrowBuf arrowBuf = allocator.buffer(source.length);
-      arrowBuf.setBytes(0, source);
-      long address = arrowBuf.memoryAddress();
-      if (address == 0L) {
-        arrowBuf.close();
-        allocator.close();
-        logger.warn(
-            "Failed to allocate native memory for binding data: payloadBytes={}", source.length);
-        throw new SQLException("Failed to allocate native memory for binding data");
+      RootAllocator allocator = null;
+      ArrowBuf arrowBuf = null;
+      boolean success = false;
+      try {
+        allocator = new RootAllocator(Long.MAX_VALUE);
+        arrowBuf = allocator.buffer(source.length);
+        arrowBuf.setBytes(0, source);
+        long address = arrowBuf.memoryAddress();
+        if (address == 0L) {
+          logger.warn(
+              "Failed to allocate native memory for binding data: payloadBytes={}", source.length);
+          throw new SQLException("Failed to allocate native memory for binding data");
+        }
+        logger.debug("Allocated native binding buffer: payloadBytes={}", source.length);
+        success = true;
+        return new NativeBuffer(allocator, arrowBuf, address);
+      } finally {
+        if (!success) {
+          if (arrowBuf != null) {
+            arrowBuf.close();
+          }
+          if (allocator != null) {
+            allocator.close();
+          }
+        }
       }
-      logger.debug("Allocated native binding buffer: payloadBytes={}", source.length);
-      return new NativeBuffer(allocator, arrowBuf, address);
     }
 
     byte[] pointerAsLittleEndianBytes() {
--- a/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/SnowflakePreparedStatementImpl.java
+++ b/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/SnowflakePreparedStatementImpl.java
@@ -25,11 +25,12 @@ import java.sql.Types;
 import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Map;
+import java.util.function.Function;
 import net.snowflake.client.api.statement.SnowflakePreparedStatement;
 import net.snowflake.client.internal.api.implementation.connection.SnowflakeConnectionImpl;
 import net.snowflake.client.internal.log.SFLogger;
 import net.snowflake.client.internal.log.SFLoggerFactory;
-import net.snowflake.client.internal.util.SnowflakeUtil;
+import net.snowflake.client.internal.util.HexUtil;
 
 /**
  * Snowflake JDBC PreparedStatement implementation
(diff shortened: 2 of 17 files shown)


jdbc/src/main/java/net/snowflake/client/internal/util/SnowflakeUtil.java
```diff
@@ -15,4 +15,18 @@ public class SnowflakeUtil {
   public static final String BYTES_STR = "byte array";
 
   private SnowflakeUtil() {}
```
sfc-gh-merbel: Did we consider adding lombok to jdbc wrapper?
@sfc-gh-rkowalski @sfc-gh-boler

sfc-gh-pfus: Yes! I insisted on this a few times already!

sfc-gh-rkowalski: I'm ok with adding it the main downsides for me are that it forces potential contributors to having code generation properly setup in their IDE and I'm not sure how would this impact AI coding tools

sfc-gh-merbel: so let's consider all pros and cons. 
Adding @sfc-gh-boler for the thread.

sfc-gh-boler: +1 to lombok, if used within reason. I believe it's built into jdtls, so most editors/IDEs should work fine.

Pretty sure AI will be able to understand it too.

sfc-gh-pfus: Lombok has been an unwritten standard for the last ~10 years (like SLF4J and so on). Everyone is able to easily set up a plugin for it in their favourite IDE.

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
```diff
@@ -0,0 +1,155 @@
...
+import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.QueryBindings;
+import org.apache.arrow.memory.ArrowBuf;
+import org.apache.arrow.memory.RootAllocator;
+import org.json.JSONStringer;
+
+final class PreparedStatementBindingSerializer {
+  private static final SFLogger logger =
+      SFLoggerFactory.getLogger(PreparedStatementBindingSerializer.class);
+
+  static final class ParameterValue {
```
sfc-gh-pfus, line 20: private?

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
```diff
@@ -0,0 +1,155 @@
...
+    private final String bindType;
```
sfc-gh-pfus, line 21: What does it cover?

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
```diff
@@ -0,0 +1,155 @@
...
+    String bindType() {
+      return bindType;
+    }
+
+    Object value() {
+      return value;
+    }
+  }
+
+  static final class SerializedBindings implements AutoCloseable {
```
sfc-gh-pfus, line 38: private?

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
```diff
@@ -0,0 +1,155 @@
...
+    private final QueryBindings bindings;
+    private final NativeBuffer buffer;
+
+    SerializedBindings(QueryBindings bindings, NativeBuffer buffer) {
+      this.bindings = bindings;
+      this.buffer = buffer;
```
sfc-gh-pfus, line 44: Hmmm, why we need this? It is never used except for closing.

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
```diff
@@ -0,0 +1,155 @@
...
+      QueryBindings queryBindings = QueryBindings.newBuilder().setJson(jsonPtr).build();
+      logger.debug(
+          "Prepared bindings serialized: payloadBytes={}, pointerBytes={}",
+          jsonBytes.length,
+          ptrBytes.length);
+      SerializedBindings serializedBindings = new SerializedBindings(queryBindings, nativeBuffer);
+      success = true;
+      return serializedBindings;
+    } finally {
+      if (!success) {
```
sfc-gh-pfus, line 109: This logic looks a bit fragile. Why we only close it if failure? Because autoclosable will close it otherwise?

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
```diff
@@ -0,0 +1,155 @@
...
+      this.address = address;
+    }
+
+    static NativeBuffer fromBytes(byte[] source) throws SQLException {
+      RootAllocator allocator = new RootAllocator(Long.MAX_VALUE);
+      ArrowBuf arrowBuf = allocator.buffer(source.length);
+      arrowBuf.setBytes(0, source);
+      long address = arrowBuf.memoryAddress();
+      if (address == 0L) {
+        arrowBuf.close();
```
sfc-gh-pfus: It also looks fragile and error-prone. I think we should handle closing in one single finally block (maybe in an upper layer?) and not based on some strange conditions.

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
```diff
@@ -0,0 +1,155 @@
...
+        throw new SQLException("Failed to allocate native memory for binding data");
+      }
+      logger.debug("Allocated native binding buffer: payloadBytes={}", source.length);
+      return new NativeBuffer(allocator, arrowBuf, address);
+    }
+
+    byte[] pointerAsLittleEndianBytes() {
+      return ByteBuffer.allocate(Long.BYTES)
+          .order(ByteOrder.LITTLE_ENDIAN)
+          .putLong(address)
```
sfc-gh-pfus, line 158: Double checking - it will work on AIX, right? IIRC Java is always big endian...

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/SnowflakePreparedStatementImpl.java
```diff
@@ -33,112 +38,132 @@
...
 
   private final String sql;
-  private Object[] parameters;
+  private final PreparedStatementBindingSerializer.ParameterValue[] parameterValues;
 
   public SnowflakePreparedStatementImpl(SnowflakeConnectionImpl connection, String sql) {
     super(connection);
     this.sql = sql;
-    // Count parameter placeholders (simple implementation)
+    // TODO: Align with snowflake-jdbc by deriving bind count from server-side describe metadata
```
sfc-gh-pfus, line 51: Do we really need this? What's the point of this? We don't have it in Go and it works fine.

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/SnowflakePreparedStatementImpl.java
```diff
@@ -33,112 +38,132 @@
...
     checkClosed();
-    setParameter(parameterIndex, x);
+    setParameter(parameterIndex, "REAL", String.valueOf(x));
   }
 
   @Override
   public void setBigDecimal(int parameterIndex, BigDecimal x) throws SQLException {
     checkClosed();
-    setParameter(parameterIndex, x);
+    if (x == null) {
```
sfc-gh-pfus: Maybe we can introduce:
```
setNullableParameter(parameterIndex, type, obj, toStringFun) {
   if (obj == null) {
     setNull(parameterIndex, type)
   } else {
      setParameter(parameterIndex, type, toStringFun(obj))
    }
}

setBigDecimal(parameterIndex, x) {
   setNullableParameter(parameterIndex, "FIXED", x, String::valueOf
}
```
WDYT?

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/SnowflakePreparedStatementImpl.java
```diff
@@ -159,27 +184,123 @@ public void setBinaryStream(int parameterIndex, InputStream x, int length) throw
...
+    }
+    if (targetSqlType == Types.TIMESTAMP) {
+      if (!(x instanceof Timestamp)) {
+        throw new SQLException(
+            "Invalid parameter type for TIMESTAMP at index "
+                + parameterIndex
+                + ": "
+                + x.getClass().getCanonicalName());
+      }
+      setTimestamp(parameterIndex, (Timestamp) x);
```
sfc-gh-pfus, line 219: How do we handle LTZ/TZ/NTZ? TODO?

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/SnowflakeStatementImpl.java
```diff
@@ -49,6 +54,14 @@ public SnowflakeStatementImpl(SnowflakeConnectionImpl connection) {
   @Override
   public ResultSet executeQuery(String sql) throws SQLException {
     checkClosed();
+    return executeQueryWithBindings(sql, null);
+  }
+
+  protected ResultSet executeQueryWithBindings(String sql, QueryBindings bindings)
+      throws SQLException {
+    boolean hasBindings = bindings != null;
```
sfc-gh-pfus, line 62: Inline?

jdbc/src/main/java/net/snowflake/client/internal/unicore/JNICoreTransport.java
```diff
@@ -31,10 +36,24 @@ public class JNICoreTransport implements CoreTransport {
   @Override
   public TransportResponse handleMessage(String serviceName, String methodName, byte[] requestBytes)
       throws TransportException {
+    logger.trace(
+        "JNI transport request: service={}, method={}, requestBytes={}",
+        serviceName,
+        methodName,
+        requestBytes == null ? 0 : requestBytes.length);
```
sfc-gh-pfus: Maybe -1?

jdbc/src/main/java/net/snowflake/client/internal/unicore/JNICoreTransport.java
```diff
@@ -31,10 +36,24 @@ public class JNICoreTransport implements CoreTransport {
...
+          "JNI transport returned null response: service={}, method={}", serviceName, methodName);
       throw new TransportException("Empty transport response");
     }
+    byte[] responseBytes = response.getResponseBytes();
+    logger.trace(
+        "JNI transport response: service={}, method={}, code={}, responseBytes={}",
+        serviceName,
+        methodName,
+        response.getCode(),
+        responseBytes == null ? 0 : responseBytes.length);
```
sfc-gh-pfus: Same.

jdbc/src/main/java/net/snowflake/client/internal/util/SnowflakeUtil.java
```diff
@@ -15,4 +15,18 @@ public class SnowflakeUtil {
...
+
+  public static String bytesToHex(byte[] value) {
```
sfc-gh-pfus: Create a dedicated class, let's not create another monster that eats everything.

jdbc/src/test/java/net/snowflake/client/api/statement/SnowflakePreparedStatementBindingTest.java
```diff
@@ -0,0 +1,472 @@
...
+        conn.prepareStatement("INSERT INTO " + tableName + " (id) VALUES (?)")) {
+      insert.setNull(1, Types.INTEGER);
+      insert.execute();
+    }
+
+    try (PreparedStatement verify =
+            conn.prepareStatement("SELECT COUNT(*), COUNT(id) FROM " + tableName);
+        ResultSet rs = verify.executeQuery()) {
+      assertTrue(rs.next(), "Verification query should return one row");
+      assertEquals(1, rs.getInt(1), "Exactly one row should be inserted");
```
sfc-gh-pfus: We could probably extract common code for all tests to a common function, there's a lot of boilerplate here.

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
sfc-gh-rkowalski: This class is used outside of `PreparedStatementBindingSerializer`

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
sfc-gh-rkowalski: This class is a simple wrapper eg. `new ParameterValue("FIXED", "42")`

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
sfc-gh-rkowalski: This too is used outside in `SnowflakePreparedStatementImpl`

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
sfc-gh-rkowalski: Yes, this is only for autocloseable

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
sfc-gh-rkowalski: Yes, this is additional guard to prevent:
> If an exception is thrown after NativeBuffer.fromBytes() allocates resources but before the SerializedBindings is returned (e.g., during QueryBindings.newBuilder().setJson(jsonPtr).build()), the nativeBuffer will not be closed, leading to a memory leak. Consider wrapping the code after NativeBuffer allocation in a try-catch block that closes nativeBuffer on exception.

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/SnowflakePreparedStatementImpl.java
sfc-gh-rkowalski: I thought about it and it depends. I don't think its required but we have it in the old driver so this could be considered behavior change

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
sfc-gh-rkowalski: Probably not and this is not the only place when we encounter big vs little endian issue. I plan to test and fix for AIX when we get at least to PrPr

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/SnowflakePreparedStatementImpl.java
sfc-gh-rkowalski: We don't yet 😄 every time related type is unsupported for now. `setTimestamp` throws `new SQLFeatureNotSupportedException("setTimestamp not supported");`

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/SnowflakeStatementImpl.java
sfc-gh-rkowalski: This value is used few times in this method, thats why it was extracted to a variable

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/SnowflakePreparedStatementImpl.java
sfc-gh-rkowalski: make sense, lets see how it works out

jdbc/src/main/java/net/snowflake/client/internal/util/SnowflakeUtil.java
sfc-gh-rkowalski: Ok, but this monster will be hungry :(

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
sfc-gh-rkowalski: Reworked this to: allocate resources inside a try, use a success flag and then close `ArrowBuf` and `RootAllocator` in a single finally block on any failure. I don't think it should be moved to upper layer since these resources are owned by this method. When `NativeBuffer` is created it will be closed properly in upper layer

jdbc/src/test/java/net/snowflake/client/api/statement/SnowflakePreparedStatementBindingTest.java
sfc-gh-rkowalski: I've refactored it a bit (around 100 lines removed)

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/SnowflakePreparedStatementImpl.java
sfc-gh-pfus: Why? This is an implementation detail only. It's transparent for customers.

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
```diff
@@ -124,19 +124,32 @@ private NativeBuffer(RootAllocator allocator, ArrowBuf arrowBuf, long address) {
...
+              "Failed to allocate native memory for binding data: payloadBytes={}", source.length);
+          throw new SQLException("Failed to allocate native memory for binding data");
+        }
+        logger.debug("Allocated native binding buffer: payloadBytes={}", source.length);
+        success = true;
+        return new NativeBuffer(allocator, arrowBuf, address);
+      } finally {
+        if (!success) {
+          if (arrowBuf != null) {
+            arrowBuf.close();
```
sfc-gh-pfus, line 146: It can throw an exception and the allocator won't be released, right?

sfc-gh-pfus (approved):

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/SnowflakePreparedStatementImpl.java
sfc-gh-rkowalski: It changes behavior a little bit. If we use describe first we can validate data when its bind to the statement not when its send to the server. I'll leave it as todo for now and we can get to this back in the future.

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
sfc-gh-rkowalski: It should not:
- NativeBuffer.fromBytes(...) is exception-safe - failed allocation path closes both ArrowBuf and RootAllocator.
- Success path returns ownership to SerializedBindings, and SerializedBindings.close() closes the native buffer.
- Callers use try-with-resources, so allocator close is deterministic on both success/failure paths.

PR #326, opened 2026-02-20
Python - Implement missing scope for execute query

sfc-gh-pczajka:
Changes:
* Added `.rownumber()` and `.sqlstate()` cursor attributes
* Properly use `StatementRelease` after executing query

### DIFF
--- a/python/tests/integ/test_cursor.py
+++ b/python/tests/integ/test_cursor.py
@@ -11,6 +11,8 @@ from snowflake.connector.errors import NotSupportedError, ProgrammingError
 from tests.e2e.types.utils import assert_sequential_values
 
 
+
+
 class TestCursorSfqid:
     """Integration tests for Cursor.sfqid property."""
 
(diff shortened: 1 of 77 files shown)


python/tests/integ/test_cursor.py
```diff
@@ -564,6 +566,237 @@ def test_rowcount_after_ddl(self, cursor, tmp_schema):
...
+        cursor.fetchone()
+
+        # Then sqlstate should still be the same
+        assert cursor.sqlstate == sqlstate_before
+
+
+class TestCursorStatementLifecycle:
+    """Integration tests for statement handle lifecycle (create/release)."""
+
+    def test_sequential_executes_do_not_leak(self, cursor):
```
sfc-gh-turbaszek, line 719: What is exactly purpose of this test?

python/tests/integ/test_cursor.py
```diff
@@ -564,6 +566,237 @@ def test_rowcount_after_ddl(self, cursor, tmp_schema):
...
+class TestCursorSqlstate:
+    """Integration tests for Cursor.sqlstate property."""
+
+    def test_sqlstate_is_none_before_execute(self, connection):
+        """Test that sqlstate returns None before any query is executed."""
+        # Given a new cursor
+        cursor = connection.cursor()
+
+        # When accessing sqlstate before execute
+        result = cursor.sqlstate
```
sfc-gh-turbaszek: Can we add some tests for values of `sqlstate`? What about old driver `sqlstate.py`, should we use it somehow?

sfc-gh-turbaszek (changes requested):

python/tests/integ/test_cursor.py
sfc-gh-pczajka: It was meant to ensure there is no memory leaks (StatementRelease was not called); I've changed it so now it contains actual checks :]

sfc-gh-turbaszek (approved):

PR #334, opened 2026-02-23
NO-SNOW: Improve Python wrapper ↔ sf_core communication

sfc-gh-turbaszek:
## Summary

- Expand `ConnectionGetInfoResponse` with `host`, `port`, `server_url`, `session_token`, `session_id` fields (replaces opaque `bytes info_data`)
- Implement `connection_get_info` in Rust core (was a not-implemented stub)
- Store `_connection_info` and `kwargs` on Python `Connection` for later use by property accessors
- Add `query` property to cursor for reading last executed SQL

Split from turbaszek-backcompat-3 — communication layer changes only.

## Test plan
- [ ] Existing integration tests pass (session parameters, query execution)
- [ ] Verify `connection_get_info` returns correct host/port/token info
- [ ] Verify `cursor.query` returns the last executed SQL text
- [ ] Verify `cursor.rowcount` works correctly for DML, SELECT, and DDL

Made with [Cursor](https://cursor.com)

PR #335, opened 2026-02-23
NO-SNOW: Add backward compatibility with old Python driver

sfc-gh-turbaszek:
## Summary

- Add `SnowflakeConnection` alias for `Connection`
- Add `execute_string()` and `execute_stream()` convenience methods with `split_statements()` SQL splitter
- Add `SnowflakeRestful` wrapper exposing `rest` property for ecosystem compatibility
- Add decorator markers (`@backward_compatibility`, `@internal_api`)
- Add stub/re-export modules: `compat.py`, `description.py`, `telemetry.py`, `time_util.py`, `util_text.py`, `auth/workload_identity.py`
- Add `ForbiddenError` to errors
- Add `CursorType`/`CursorInstance` type aliases, `_is_put_get` param on `execute()`
- Change `autocommit` from property to method for old driver compat
- Add `role`, `database`, `schema`, `account`, `warehouse` connection properties
- Add `_rewrite_private_key_password` for `private_key_file_pwd` compat

Split from turbaszek-backcompat-3 — backward compatibility changes only.

> **Note:** For full `SnowflakeRestful` functionality (session token, host info), merge PR #334 (communication improvements) first.

## Test plan
- [ ] `python/tests/integ/test_connection.py` passes (autocommit, property accessors)
- [ ] `python/tests/integ/test_cursor.py` passes
- [ ] `execute_string` and `execute_stream` work with multi-statement SQL
- [ ] `SnowflakeConnection` alias is importable from `snowflake.connector`
- [ ] Stub modules are importable without errors


python/pyproject.toml
```diff
@@ -28,6 +28,9 @@ dependencies = [
     "cryptography>=44.0.1",
     "protobuf>=6.32.1",
     "pytz",
+    # For backcompat reasons
+    "platformdirs>=2.6.0,<5.0.0",
+    "tomlkit",
```
sfc-gh-turbaszek: Those are not used and should be removed.

python/ruff.toml
```diff
@@ -3,7 +3,7 @@ extend-exclude = [
     '.hatch/',
     'src/snowflake/connector/_core',
     'src/snowflake/connector/_internal/protobuf_gen',
-    'src/snowflake/connector/_internal/arrow_context.py'
+    'src/snowflake/connector/_internal/arrow_context.py',
```
sfc-gh-turbaszek: Unnecessary change

python/src/snowflake/connector/_internal/snowflake_restful.py
```diff
@@ -0,0 +1,37 @@
...
+    def _protocol(self) -> str | None:
+        return urlparse(self._connection_info.server_url).scheme
+
+    @property
+    def _port(self) -> int | None:
+        return urlparse(self._connection_info.server_url).port or 443
+
+    @property
+    def master_token(self) -> str | None:
+        return "TODO"
```
sfc-gh-fpawlowski: Would making it return NotImplementedError cause pipelines to fail significantly more than with this wrong value? Could we add comment with TODO and JIRA number?

python/src/snowflake/connector/_internal/snowflake_restful.py
```diff
@@ -0,0 +1,37 @@
...
+
+from urllib.parse import urlparse
+
+from snowflake.connector._internal.protobuf_gen.database_driver_v1_services import (
+    ConnectionGetInfoResponse,
+)
+
+
+class SnowflakeRestful:
+    """Extend as required"""
```
sfc-gh-fpawlowski: Should we have more elaborate comment or remove this one?

python/src/snowflake/connector/connection.py
```diff
@@ -198,35 +210,20 @@ def set_autocommit(self, autocommit: bool) -> None:
 
         Args:
             autocommit (bool): True to enable autocommit, False to disable
-
-        Raises:
-            NotSupportedError: If not implemented
         """
-        raise NotSupportedError("set_autocommit is not implemented")
+        # TODO: Lacks full implementation
```
sfc-gh-fpawlowski: nit: should we have follow up jira and more elaborate comment with its number

python/src/snowflake/connector/connection.py
```diff
@@ -198,35 +210,20 @@ def set_autocommit(self, autocommit: bool) -> None:
...
-
-    @property
-    def autocommit(self) -> bool:
-        """
-        Get/set autocommit mode as a property.
-
         Returns:
             bool: Current autocommit setting
         """
+        # TODO: Lacks full implementation
```
sfc-gh-fpawlowski: nit: as above - jira or better comment whats lacking for now?

python/src/snowflake/connector/connection.py
```diff
@@ -198,35 +210,20 @@ def set_autocommit(self, autocommit: bool) -> None:
...
         return self._autocommit
 
-    @autocommit.setter
     def autocommit(self, value: bool) -> None:
```
sfc-gh-fpawlowski, line 225: should we mark this or set_../get_autocommit as backward_compatibility?

python/src/snowflake/connector/connection.py
```diff
@@ -271,3 +265,80 @@ def paramstyle(self) -> ParamStyle:
...
+        for sql, is_put_or_get in non_empty_statements:
+            cur = self.cursor(cursor_class=cursor_class)
+            cur.execute(sql, _is_put_get=is_put_or_get)
+            yield cur
+
+    @property
+    @internal_api
+    @backward_compatibility
+    def rest(self) -> SnowflakeRestful:
+        return SnowflakeRestful(connection_info=self._connection_info)
```
sfc-gh-fpawlowski: is _connection_info updated in core or do we plan for it to be? Won't we end up with outdated data in the Connection instance? Shouldn't we instead call db_api always in such cases for the newest data?

sfc-gh-fpawlowski (commented):
Left few comments

python/src/snowflake/connector/_internal/snowflake_restful.py
sfc-gh-turbaszek: Done

python/src/snowflake/connector/_internal/snowflake_restful.py
sfc-gh-turbaszek: Changed

python/src/snowflake/connector/connection.py
sfc-gh-turbaszek: Added JIRA

python/src/snowflake/connector/connection.py
sfc-gh-turbaszek: Good question, I intended `backward_compatibility` as marker of something that is to be remove in future. Not sure if `autocommit` is something we need to tackle.

python/src/snowflake/connector/connection.py
sfc-gh-turbaszek: Good question, this seems to be a bigger problem not only related to SnowflakeRestful.

sfc-gh-turbaszek: Fixed for now by refreshing details every time access is made.

python/src/snowflake/connector/_internal/snowflake_restful.py
```diff
@@ -0,0 +1,45 @@
...
+from urllib.parse import urlparse
+
+from snowflake.connector._internal.protobuf_gen.database_driver_v1_pb2 import ConnectionGetInfoResponse
+
+
+if TYPE_CHECKING:
+    from ..connection import SnowflakeConnection
+
+
+class SnowflakeRestful:
```
sfc-gh-fpawlowski, line 13: Should we actually decorate SnowflakeRestful as backward_compatibility? Would we have it here if not for the old driver

python/tests/integ/test_connection.py
```diff
@@ -68,48 +70,196 @@ def test_ping_not_implemented(self, connection):
...
+
+    def test_execute_string_remove_comments(self, connection):
+        """Test execute_string with remove_comments=True."""
+        sql = "-- leading comment\nSELECT 1; /* inline */ SELECT 2"
+        # When executing with remove_comments
+        cursors = connection.execute_string(sql, remove_comments=True)
+
+        # Then statements should still execute correctly
+        cursors = list(cursors)
+        assert len(cursors) == 2
```
sfc-gh-fpawlowski, line 184: nit: we are not checking if we are actually getting values that we expect and not some 2 cursors without actual returned values created immediately when splitting strings (without the actual execution).

python/src/snowflake/connector/connection.py
```diff
@@ -271,3 +263,86 @@ def paramstyle(self) -> ParamStyle:
...
+    @backward_compatibility
+    def _rewrite_private_key_password(self, kwargs: ConnectionParameters) -> ConnectionParameters:
+        private_key_file_pwd = kwargs.pop("private_key_file_pwd", None)
+        if private_key_file_pwd is not None:
+            kwargs = {**kwargs, "private_key_password": private_key_file_pwd}
+        return kwargs
+
+    @property
+    def role(self) -> str | None:
+        return self.kwargs.get("role")  # type: ignore[return-value]
```
sfc-gh-fpawlowski, line 324: Might be worth leaving todo and Jira for migrating this into protobuf call to be sure we return the values that reflect server-state (e.g. when someone changed the initial role with USE ROLE;)

python/src/snowflake/connector/telemetry.py
```diff
@@ -0,0 +1,19 @@
...
+        self.timestamp = timestamp
+
+
+class TelemetryField(Enum):
+    KEY_SOURCE = "source"
+    KEY_TYPE = "type"
+
+
+class TelemetryClient:
+    def try_add_log_to_batch(self, *args, **kwargs):  # type: ignore
```
sfc-gh-fpawlowski, line 18: TODO: we might want to raise NotImplementedError here

sfc-gh-fpawlowski (approved):
LGTM, left small nits

python/src/snowflake/connector/connection.py
sfc-gh-fpawlowski: Was it commited?

python/src/snowflake/connector/connection.py
sfc-gh-fpawlowski: It is deprecated in pep249 due to issues in concurrent environments so I supposed we would like to tackle this in the future to be able to avoid ambigous behaviour in asyncio. Wdyt?

python/src/snowflake/connector/_internal/snowflake_restful.py
sfc-gh-turbaszek: Imo no need, now this is only internal API that will be removed once it's not referenced in public surface

python/src/snowflake/connector/telemetry.py
sfc-gh-turbaszek: Done

python/src/snowflake/connector/connection.py
sfc-gh-turbaszek: Done

python/tests/integ/test_connection.py
sfc-gh-turbaszek: Fixed

PR #338, opened 2026-02-23
SNOW-2972472 token cache

sfc-gh-mhofman:
### TL;DR

Added a new `token_cache` module to `sf_core` that provides a trait-based interface for secure authentication token storage and retrieval.

### What changed?

- Added `token_cache.rs` module with a `TokenCache` trait defining methods for adding, removing, and retrieving tokens from secure storage
- Introduced `TokenType` enum supporting ID tokens, MFA tokens, OAuth access/refresh tokens, and DPoP bundled access tokens
- Implemented comprehensive error handling with `TokenCacheError` enum covering keystore access, storage, retrieval, and removal failures
- Added utility function `build_cache_key()` that constructs cache keys using the format `"{host};{username};{token_type}"`
- Exposed the new module in `sf_core/src/lib.rs`

### How to test?

- Implement the `TokenCache` trait for your preferred keystore backend
- Test token storage and retrieval operations with different `TokenType` variants
- Verify error handling by simulating keystore access failures
- Validate cache key generation with various host/username/token type combinations

### Why make this change?

This change establishes a foundation for secure token caching functionality, enabling authentication tokens to be stored and retrieved from platform-specific keystores in a consistent manner across different token types and authentication flows.

PR #343, opened 2026-02-23
SNOW-2972472 File based token cache fallback

sfc-gh-mhofman:
SNOW-2972472 File based token cache fallback

This change introduces file based token cache when OS-specific keyring solution is unavailable. Token cache locking mechanism is provided. Stale lock will get removed after a timeout.    
API for file based cache remains blocking.  
Location for token cache lookup in below order:

1. $SF_TEMPORARY_CREDENTIAL_CACHE_DIR/credential_cache_v2.json
2. $XDG_CACHE_HOME/snowflake/credential_cache_v1.json
3. $HOME/.cache/snowflake/credential_cache_v1.json

Tests provided.  
Test scenario descriptions in the next PR.


sfc-gh-mhofman:
* **#502** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/502?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#386** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/386?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#348** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/348?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#343** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/343?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/343?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#338** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/338?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sf_core/src/token_cache/file_cache.rs
```diff
@@ -0,0 +1,850 @@
...
+use std::time::{Duration, SystemTime};
+
+use keyring::credential::{CredentialApi, CredentialBuilderApi, CredentialPersistence};
+use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
+use snafu::Location;
+
+use super::TokenCacheError;
+
+const DEFAULT_CACHE_FILE_NAME: &str = "credential_cache_v2.json";
```
sfc-gh-pbulawa, line 20: How do we ensure for the customers using the pre-UD driver, who set the custom filename, to keep the compatibility? Shouldn't we have some kind of fallback?  
  
Also, maybe it's worth to design the v2 and onwards to have some kind of a header in the file that will define the credential cache version. This way we could implement mechanisms to fallback the read to the old implementation.

sf_core/src/token_cache/file_cache.rs
```diff
@@ -0,0 +1,850 @@
...
+    fn ensure_file_permissions(&self) -> Result<(), TokenCacheError> {
+        use std::os::unix::fs::PermissionsExt;
+        let metadata =
+            fs::metadata(&self.cache_file_path).map_err(|e| TokenCacheError::TokenRetrieval {
+                source: Box::new(e),
+                location: Location::default(),
+            })?;
+        let mode = metadata.permissions().mode() & 0o777;
+        if mode != 0o600 {
+            return Err(TokenCacheError::InsufficientPermissions {
```
sfc-gh-pbulawa: Is it Insufficient, Too broad or just wrong file permissions?

sf_core/src/token_cache/file_cache.rs
sfc-gh-jszczerbinski: IMHO Trying to keep compatibility has potential issues:
- New driver corrupting cache for old driver
- Old driver corrupting cache for new driver
The upside we get is minimal small subset of clients that use new and old drivers on the same machine will get less authentications. I think it's a very rare use case and we shouldn't worry about since most of the clients will get more authentications simply because we switch keystore implementation for Linux.

sf_core/src/token_cache/file_cache.rs
```diff
@@ -0,0 +1,850 @@
...
+            });
+        }
+        Ok(())
+    }
+
+    #[cfg(unix)]
+    fn ensure_file_ownership(&self) -> Result<(), TokenCacheError> {
+        use std::os::unix::fs::MetadataExt;
+        let metadata =
+            fs::metadata(&self.cache_file_path).map_err(|e| TokenCacheError::TokenRetrieval {
```
sfc-gh-jszczerbinski: Shouldn't write and ensure metadata on the same file descriptor?

sf_core/src/token_cache/file_cache.rs
```diff
@@ -0,0 +1,850 @@
...
+
+    fn read_cache(&self) -> Result<CacheFileContent, TokenCacheError> {
+        if !self.cache_file_path.exists() {
+            return Ok(CacheFileContent {
+                tokens: HashMap::new(),
+            });
+        }
+
+        #[cfg(unix)]
+        self.ensure_file_ownership()?;
```
sfc-gh-jszczerbinski: I believe we need to check file permissions, ownership and do the read or write on the same file descriptor

sfc-gh-jszczerbinski: Right now we are in data race between checking and reading/writing

sf_core/src/token_cache/file_cache.rs
```diff
@@ -0,0 +1,850 @@
...
+            .and_then(|m| m.modified().ok())
+            .and_then(|modified| SystemTime::now().duration_since(modified).ok())
+            .map(|age| age > stale_timeout)
+            .unwrap_or(true)
+    }
+}
+
+impl Drop for FileLock {
+    fn drop(&mut self) {
+        let _ = fs::remove_file(&self.lock_path);
```
sfc-gh-jszczerbinski: Maybe a warning?

sf_core/src/token_cache/file_cache.rs
```diff
@@ -0,0 +1,850 @@
...
+            })?;
+        }
+
+        for attempt in 0..retry_count {
+            match fs::OpenOptions::new()
+                .write(true)
+                .create_new(true)
+                .open(&lock_path)
+            {
+                Ok(_) => {
```
sfc-gh-jszczerbinski: I am pretty sure we create close the file here? Shouldn't we keep it open?

sf_core/src/token_cache/file_cache.rs
```diff
@@ -0,0 +1,850 @@
...
+    fn acquire(
+        cache_path: &Path,
+        retry_count: u32,
+        retry_delay: Duration,
+        stale_timeout: Duration,
+    ) -> Result<Self, TokenCacheError> {
+        let lock_path = cache_path.with_extension("json.lck");
+
+        if let Some(parent) = lock_path.parent() {
+            fs::create_dir_all(parent).map_err(|e| TokenCacheError::LockAcquisition {
```
sfc-gh-jszczerbinski: So a cache_path?

sfc-gh-jszczerbinski: I think we shouldn't create $HOME here for user at top dir is probably with wrong permissions

sf_core/src/token_cache/file_cache.rs
```diff
@@ -0,0 +1,850 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+struct CacheFileContent {
+    tokens: HashMap<String, String>,
+}
+
+/// Resolves the cache directory from environment variables in priority order:
+/// 1. `$SF_TEMPORARY_CREDENTIAL_CACHE_DIR`
+/// 2. `$XDG_CACHE_HOME/snowflake`
+/// 3. `$HOME/.cache/snowflake`
+fn resolve_cache_dir() -> Option<PathBuf> {
```
sfc-gh-jszczerbinski: Shouldn't we:
- Check permissions for the directories
- Create if they don't exist

sf_core/src/token_cache/file_cache.rs
```diff
@@ -0,0 +1,850 @@
...
+                Ok(_) => {
+                    return Ok(FileLock { lock_path });
+                }
+                Err(ref e) if e.kind() == std::io::ErrorKind::AlreadyExists => {
+                    if Self::is_stale(&lock_path, stale_timeout) {
+                        let _ = fs::remove_file(&lock_path);
+                        continue;
+                    }
+                    if attempt < retry_count - 1 {
+                        std::thread::sleep(retry_delay);
```
sfc-gh-jszczerbinski, line 129: Arre we running this in async code? I think we shouldn't block execution thread. Should we make this async and use https://docs.rs/tokio/latest/tokio/time/fn.sleep.html

sf_core/src/token_cache/file_cache.rs
```diff
@@ -0,0 +1,850 @@
...
+
+        serde_json::from_str(&content).map_err(|e| TokenCacheError::TokenRetrieval {
+            source: Box::new(e),
+            location: Location::default(),
+        })
+    }
+
+    fn write_cache(&self, cache: &CacheFileContent) -> Result<(), TokenCacheError> {
+        if let Some(parent) = self.cache_file_path.parent() {
+            fs::create_dir_all(parent).map_err(|e| TokenCacheError::TokenStorage {
```
sfc-gh-jszczerbinski: Same here, what permissions and why is a driver creating homedir for me?

sf_core/src/token_cache/file_cache.rs
```diff
@@ -0,0 +1,850 @@
...
+    #[cfg(not(unix))]
+    fn write_with_permissions(&self, content: &str) -> Result<(), TokenCacheError> {
+        fs::write(&self.cache_file_path, content).map_err(|e| TokenCacheError::TokenStorage {
+            source: Box::new(e),
+            location: Location::default(),
+        })
+    }
+
+    #[cfg(unix)]
+    fn ensure_file_permissions(&self) -> Result<(), TokenCacheError> {
```
sfc-gh-jszczerbinski: What about parent directory permissions? Are we requiring any?

sf_core/src/token_cache/file_cache.rs
```diff
@@ -0,0 +1,850 @@
...
+            return Err(TokenCacheError::InsufficientPermissions {
+                path: self.cache_file_path.clone(),
+                location: Location::default(),
+            });
+        }
+        Ok(())
+    }
+
+    #[cfg(unix)]
+    fn ensure_file_ownership(&self) -> Result<(), TokenCacheError> {
```
sfc-gh-jszczerbinski: What about parent dir?

sfc-gh-jszczerbinski (changes requested):

sf_core/src/token_cache/file_cache.rs
sfc-gh-mhofman: Changed to a directory based lock

sf_core/src/token_cache/file_cache.rs
```diff
@@ -0,0 +1,1233 @@
...
+                        tokio::time::sleep(retry_delay).await;
+                    }
+                }
+                Err(e) => {
+                    return Err(e).context(LockAcquisitionSnafu);
+                }
+            }
+        }
+
+        Err(std::io::Error::new(
```
sfc-gh-jszczerbinski: Why create underlying error here? Shouldn't we just have another Snafu

sf_core/src/token_cache/file_cache.rs
```diff
@@ -0,0 +1,1233 @@
...
+struct FileCredential {
+    #[allow(dead_code)]
+    service: String,
+    user: String,
+    cache: Arc<FileTokenCache>,
+}
+
+impl CredentialApi for FileCredential {
+    fn set_secret(&self, secret: &[u8]) -> keyring::Result<()> {
+        tokio::task::block_in_place(|| {
```
sfc-gh-jszczerbinski: We will acquire blocking thread here and still use executor threads.   
  
Let's revert back to the sync code and we can call all caches with tokio::task::block_in_place. This way we will acquire blocking thread, but we won't use executor threads. This way we should also handle long running blocking operations in keyring cache (like waiting for user to unlock keyring)

sfc-gh-jszczerbinski: The alternative is to expose async api to the caller

sfc-gh-jszczerbinski (changes requested):
The cache implementation looks good:  
\- we need to rethink our approach to sync  
   - Either top level interface should be async  
   - or we should keep the code sync  
\+ one minor comment

sf_core/src/token_cache/file_cache.rs
sfc-gh-mhofman: Let's keep things simple. Having timelines in mind. We can improve later if needed

sf_core/src/token_cache/file_cache.rs
sfc-gh-mhofman: Fixed

sf_core/src/token_cache/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+        host: &str,
+        username: &str,
+        token_type: TokenType,
+        token_value: &str,
+    ) -> Result<(), TokenCacheError>;
+
+    /// Removes a token from the keystore.
+    ///
+    /// # Arguments
+    /// * `host` - The host associated with the token
```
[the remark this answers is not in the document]
sfc-gh-mhofman: No, host can be any IdP machine, not necessarily Snowflake's host

sf_core/src/token_cache/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+        &self,
+        host: &str,
+        username: &str,
+        token_type: TokenType,
+    ) -> Result<(), TokenCacheError>;
+
+    /// Retrieves a token from the keystore.
+    ///
+    /// # Arguments
+    /// * `host` - The host associated with the token
```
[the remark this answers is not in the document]
sfc-gh-mhofman: No, as above

sf_core/src/token_cache/file_cache.rs
sfc-gh-mhofman: We can add async api later when needed, wdyt?

sfc-gh-jszczerbinski: Sure, but let's revert to sync version then.

sfc-gh-mhofman:
As we discussed @sfc-gh-jszczerbinski async ops from cache got removed.

sfc-gh-jszczerbinski (approved):

sfc-gh-mhofman:
### Merge activity

* **Mar 2, 1:18 PM UTC**: @sfc-gh-mhofman merged this pull request with [Graphite](https://app.graphite.com/github/pr/snowflakedb/universal-driver/343).

PR #344, opened 2026-02-23
NO-SNOW Add configuration option to load a test private key from path instead of content

sfc-gh-pfus:
This PR makes loading private key from the parameters.json easier - instead of having private key content embedded into JSON, just use `SNOWFLAKE_TEST_PRIVATE_KEY_FILE` to provide a private key path.

### DIFF
--- a/jdbc/build.gradle
+++ b/jdbc/build.gradle
@@ -58,7 +58,7 @@ dependencies {
 spotless {
     java {
         removeUnusedImports()
-        googleJavaFormat('1.15.0')
+        googleJavaFormat('1.7')
         target 'src/main/java/**/*.java', 'src/test/java/**/*.java'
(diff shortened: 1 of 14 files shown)


sfc-gh-jszczerbinski (changes requested):
IMHO not needed, also this common conf for all tests not just Core - please add support for languages too if you disagree if its needed

sfc-gh-pfus:
> IMHO not needed, also this common conf for all tests not just Core - please add support for languages too if you disagree if its needed

I think it has a few advantages:
1. You don't need to convert your file content into JSON array for every environment you want to use.
2. You don't have to undergo the same procedure for every key pair change.

sfc-gh-jszczerbinski (approved):

PR #351, opened 2026-02-24
NO-SNOW: Generate protobufs when building python

sfc-gh-turbaszek:
Remove the checked-in generated protobuf Python files and instead generate them at build time, similar to how the Rust protobuf code is already generated on the fly via build.rs.


sfc-gh-jszczerbinski (approved):
LGTM

PR #352, opened 2026-02-24
[2/7] config: introduce ParamRegistry as a centralized parameter definition table

sfc-gh-boler:
There was no single source of truth for what connection parameters the driver accepts. Canonical names, aliases, types, defaults, and sensitivity flags were scattered across `rest_parameters.rs`, `tls/config.rs`, `crl/config.rs`, and the ODBC `connection.rs` — making alias resolution, type validation, and log redaction ad-hoc in each caller.

This change introduces `sf_core/src/config/param_registry.rs`, a static table of `ParamDef` entries covering all ~30 known parameters across the Server, Auth, Session, TLS, CRL, and Client scopes. Each entry carries the canonical name, accepted aliases (e.g. `SERVER`/`HOST` → `host`, `UID` → `user`), expected `ValueType`, `Required` policy, an optional default factory, a `sensitive` flag for log redaction, and a human-readable description. At startup a `ParamRegistry` singleton builds a lowercase alias index over the table using `once_cell::sync::Lazy`, making all lookups O(1) and allocation-free after initialization. The `resolve()` method performs case-insensitive lookup by canonical name or any alias; `is_known()` provides a fast existence check.

A `param_names` submodule provides `pub const &str` constants for every canonical parameter name (e.g. `ACCOUNT`, `HOST`, `USER`). The `PARAM_DEFS` table references these constants instead of inline string literals, and downstream code uses them for compile-time typo detection and IDE navigation. The module is re-exported from `config::mod` for ergonomic access.

The registry is not yet wired into `connection_set_option()` — that happens in the next PR.

### DIFF
--- a/sf_core/src/config/param_registry.rs
+++ b/sf_core/src/config/param_registry.rs
@@ -1,0 +1,598 @@
+use std::collections::HashMap;
+use std::fmt;
+
+use std::sync::LazyLock;
+
+use crate::config::settings::Setting;
+
+/// A strongly-typed wrapper around a canonical parameter name.
+///
+/// Provides compile-time safety over bare `&str` keys while remaining
+/// zero-cost at runtime (it is `Copy` and stores a `&'static str`).
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct ParamKey(pub(crate) &'static str);
+
+impl ParamKey {
+    pub const fn as_str(&self) -> &'static str {
+        self.0
+    }
+}
+
+impl fmt::Display for ParamKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(self.0)
+    }
+}
+
+impl From<ParamKey> for String {
+    fn from(key: ParamKey) -> String {
+        key.0.to_owned()
+    }
+}
+
+impl AsRef<str> for ParamKey {
+    fn as_ref(&self) -> &str {
+        self.0
+    }
+}
+
+/// Canonical parameter name constants.
+///
+/// Use these instead of bare string literals when referencing parameter names
+/// in production code.  This gives compile-time typo detection and
+/// find-all-references support.
+pub mod param_names {
+    use super::ParamKey;
+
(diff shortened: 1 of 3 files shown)


sf_core/src/config/param_registry.rs
```diff
@@ -0,0 +1,553 @@
...
+    pub required: Required,
+
+    /// Default value factory, if any.
+    pub default: Option<fn() -> Setting>,
+
+    /// Whether the value contains secrets (for log redaction).
+    pub sensitive: bool,
+
+    /// Logical grouping (for documentation/validation).
+    pub scope: ParamScope,
```
sfc-gh-mhofman: There are session params which are at the same time client params. 
e.g. CLIENT_STORE_TEMPORARY_CREDENTIAL or CLIENT_REQUEST_MFA_TOKEN.
Also an auth parameter might be also client and session param.
Should those scopes be rather a subset here then?

sf_core/src/config/param_registry.rs
```diff
@@ -0,0 +1,553 @@
...
+    Never,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ParamScope {
+    Server,
+    Auth,
+    Session,
+    Tls,
+    Crl,
```
sfc-gh-mhofman: Ocsp? or in general cert?

sfc-gh-mhofman (commented):
Added some comments

sfc-gh-turbaszek:
Is there a plan to expose those parameters in wrappers? Currently in python there's no code that tells users what connection parameters are available. This cause issues with linters in users code bases (e.g no static validation).

Should we define `ConnectionParametrs` as protobuf message that then can be used across core and wrappers?

sfc-gh-boler:
> Is there a plan to expose those parameters in wrappers? Currently in python there's no code that tells users what connection parameters are available. This cause issues with linters in users code bases (e.g no static validation).
>
> Should we define `ConnectionParametrs` as protobuf message that then can be used across core and wrappers?

​Yes, I have a whole chain of PRs ready, including interfaces for py, odbc, jdbc. I'll upload them shortly

PR #359, opened 2026-02-25
SNOW-3111869 Add a JSON result set feature

sfc-gh-pfus:
This PR introduces a new feature that verifies that core creates a correct Arrow from JSON result set. It is based on the existing code and only formalises it a bit.

The biggest change is how we create integer Arrow arrays - based on the actual values (simulating backend behaviour).


sf_core/tests/e2e/query/json_result_set.rs
```diff
@@ -0,0 +1,61 @@
...
+
+impl ArrowDeserialize for AllTypes {
+    fn deserialize_one(
+        batch: &crate::common::arrow_deserialize::RecordBatch,
+        row_index: usize,
+    ) -> Result<Self, String> {
+        Ok(AllTypes {
+            s: extract_arrow_value::<String>(batch.column(0), row_index)
+                .map_err(|e| e.to_string())?,
+            i: extract_arrow_value::<i64>(batch.column(1), row_index).map_err(|e| e.to_string())?,
```
sfc-gh-jszczerbinski:
```suggestion
            i: extract_arrow_value::<i64>(batch.column(1), row_index).map_err(|e| format!("{:?}", e))?,
```
You don't need to implement Display for this type

sf_core/tests/e2e/query/json_result_set.rs
```diff
@@ -0,0 +1,61 @@
+use crate::common::arrow_deserialize::ArrowDeserialize;
+use crate::common::arrow_extract_value::extract_arrow_value;
+use crate::common::arrow_result_helper::ArrowResultHelper;
+use crate::common::snowflake_test_client::SnowflakeTestClient;
+
+#[derive(Debug)]
```
sfc-gh-jszczerbinski:
```suggestion
#[derive(Debug, ArrowDeserialize, PartialEq)]
```
Don't implement, derive.

sf_core/tests/e2e/query/json_result_set.rs
```diff
@@ -0,0 +1,61 @@
...
+    );
+    let result = client.execute_statement_query(&stmt);
+    assert_eq!(result.rows_affected(), 1, "Cannot force JSON result set");
+
+    // When Query "SELECT 'abc', 123" is executed
+    client.set_sql_query(&stmt, "SELECT 'abc', 123");
+    let result = client.execute_statement_query(&stmt);
+
+    // Then all values are deserialized correctly
+    let mut arrow_helper = ArrowResultHelper::from_result(result);
```
sfc-gh-jszczerbinski: So I don't think we should use result helper for those tests since it's a extra layer of abstraction that does conversion. I would suggest that the tests look like this:
- Run query two times: Once with arrow, Once with JSON
- Compare raw arrow: schema, metadata, values etc. it should be 100% the same

sfc-gh-jszczerbinski (changes requested):
Some rust comments (doing my job as Rust preacher :P), but I think we should decide on a better approach (the main point here)

sf_core/tests/e2e/query/json_result_set.rs
sfc-gh-jszczerbinski: This way we can do tests in quite automated manner just have one parametrized test and run different queries on it

sf_core/tests/e2e/query/json_result_set.rs
sfc-gh-pfus: Is it more readable? And we have to do the same in every place where we want to use it? IMHO this `Display` is more readable and easier to use.

sf_core/tests/e2e/query/json_result_set.rs
sfc-gh-pfus: This part is removed.

sf_core/tests/e2e/query/json_result_set.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    let mut arrow_result_helper = ArrowResultHelper::from_result(arrow_result);
+    let mut json_result_helper = ArrowResultHelper::from_result(json_result);
+
+    // Then Schema for both queries should match
+    let arrow_schema = arrow_result_helper.schema();
+    let json_schema = json_result_helper.schema();
+    assert_eq!(arrow_schema, json_schema, "Schemas do not match");
+
+    let arrow_records = arrow_result_helper
```
sfc-gh-jszczerbinski: Let's not convert records, but compare them directly too.

sfc-gh-jszczerbinski (changes requested):

sf_core/tests/e2e/query/json_result_set.rs
```diff
@@ -0,0 +1,49 @@
...
+    client.execute_statement_query(&stmt);
+
+    // Query "SELECT * FROM json_result_set_simple_types" is executed
+    client.set_sql_query(&stmt, "SELECT * FROM json_result_set_simple_types");
+    let arrow_result = client.execute_statement_query(&stmt);
+
+    // And Query result format is forced to JSON
+    client.set_sql_query(
+        &stmt,
+        "ALTER SESSION SET PYTHON_CONNECTOR_QUERY_RESULT_FORMAT = JSON",
```
sfc-gh-turbaszek, line 27: Do we have such param for every driver? Should we consider creating a driver-agnostic one? Or we see this as test-only configuration for now?

tests/definitions/core/query/json_result_set.feature
```diff
@@ -0,0 +1,19 @@
...
+    And Query "SELECT * FROM json_result_set_simple_types" is executed
+    And Query result format is forced to JSON
+    And Query "SELECT * FROM json_result_set_simple_types" is executed
+    Then Schema for both queries should match
+    And the result for both queries should match
+    And Statement should be released
+
+  # TODO add a test for larger result set with chunks
+
+  # TODO add a test for all possible data types
```
sfc-gh-turbaszek, line 18: Is this tracked by some other ticket? TIMESTAMP_LTZ is critical for SHOW/DESC queries.

sfc-gh-turbaszek (approved):
Two comments, otherwise looks reasonable to me

sf_core/tests/e2e/query/json_result_set.rs
sfc-gh-pfus: We have such a parameter for every driver that supports Arrow (all except Node.JS and PHP AFAIK). This is a test only param for now - UD uses python when introducing itself to client backend..

tests/definitions/core/query/json_result_set.feature
sfc-gh-pfus: It is tracked by https://snowflakecomputing.atlassian.net/browse/SNOW-3111876 and I'm currently working on this while waiting for review.

sf_core/tests/e2e/put_get/put_get_basic_operations.rs
```diff
@@ -183,7 +183,6 @@ fn check_text_field(field: &Field, name: &str) {
     assert_eq!(field.name(), name);
     let m0 = field.metadata();
     assert_eq!(m0.get("logicalType"), Some(&"TEXT".to_string()));
-    assert_eq!(m0.get("physicalType"), Some(&"LOB".to_string()));
```
sfc-gh-jszczerbinski, line 186: Why do we remove those assertions?

sf_core/tests/common/arrow_result_helper.rs
```diff
@@ -113,3 +115,114 @@ impl ArrowResultHelper {
...
+            .iter()
+            .map(|(k, v)| (k.clone(), v.clone()))
+            .collect();
+        let json_meta: BTreeMap<String, String> = json_field
+            .metadata()
+            .iter()
+            .map(|(k, v)| (k.clone(), v.clone()))
+            .collect();
+
+        let filtered_arrow = filter_metadata(&arrow_meta);
```
sfc-gh-jszczerbinski, line 187: Why do we exclude here?

sf_core/src/arrow_utils.rs
```diff
@@ -77,40 +67,112 @@ pub fn create_field(row_type: &RowType) -> Field {
...
-                        value: v.to_string(),
-                    })
-                })
-                .collect();
-            Ok(Arc::new(Int64Array::from(int_values?)))
+
+            let decimal_values = decimal_values?;
+            if decimal_values.is_empty() {
+                return Ok((
+                    create_field_with_type(row_type, DataType::Int64), // TODO is it correct? We have to assume something, but it probably doesn't matter.
```
sfc-gh-jszczerbinski, line 132: Shouldn't this be DataType::Decimal128? This the only type where we can guarantee that all values fit in

sfc-gh-jszczerbinski: Can we test it with SELECT 1::NUMBER WHERE 0=1;

sfc-gh-jszczerbinski (changes requested):
Will we do more extensive testing at this level?

sf_core/tests/common/arrow_result_helper.rs
sfc-gh-pfus: Surprisingly, some metadata differ depending on the actual query. Example: when I have a query that returns only strings and integers, the backend returns only the relevant metadata for all columns. But when I add NUMBER(38, 2) to the same query, the backend returns irrelevant metadata: it returns `finalType: T` for all fields, precision and scale for strings, and so on. These are not relevant, so we should focus only on the required fields.

sf_core/src/arrow_utils.rs
sfc-gh-pfus: This is a case where we have empty result set, so actually this shouldn't matter. What should matter is that we return any integer type.

sf_core/tests/e2e/put_get/put_get_basic_operations.rs
sfc-gh-pfus: Because `physicalType` is not needed by wrappers and this is not easy to get this accurate. We can always add it later if it is really needed.

sfc-gh-pfus:
> Will we do more extensive testing at this level?

Yes, but this PR only introduces tests for what was already implemented. My plan is to focus on timestamps now, because they are needed by snowfort, and then focus on more thorough implementation of all types and tests.

sfc-gh-jszczerbinski (approved):

PR #361, opened 2026-02-25
NO-SNOW: Add Connection and Cursor API stubs

sfc-gh-turbaszek:


python/src/snowflake/connector/connection.py
```diff
@@ -321,28 +339,210 @@ def _rewrite_private_key_password(self, kwargs: ConnectionParameters) -> Connect
...
+        """The login timeout in seconds."""
+        raise NotImplementedError("login_timeout is not yet implemented")
+
+    @property
+    def network_timeout(self) -> int | None:
+        """The network timeout in seconds for all other operations."""
+        raise NotImplementedError("network_timeout is not yet implemented")
+
+    @property
+    def socket_timeout(self) -> int | None:
```
sfc-gh-fpawlowski, line 409: nit: Were the networking decisions taken already? Do we know if this kind of mapping will be possible with core's timeout params or should some of those be marked as backward_compat? 

I suppose we can fix that reactively in the future - just asking out of curiosity

scripts/python_api_state.py
```diff
@@ -0,0 +1,253 @@
+#!/usr/bin/env python3
+"""Analyze the public API of Connection and Cursor classes and report implementation status.
+
+The script inspects the source code via the ``ast`` module so it does not
+need to import or instantiate any driver classes (no native libraries
+required).
+
+Usage:
+    python scripts/python_api_state.py
```
sfc-gh-fpawlowski, line 9: Do we intend to call this script in CI or is it simply helper script?

sfc-gh-fpawlowski (approved):
Overall looks good - we may want to have some tests or CI checks for the API we just have introduced (especially the pep249 api), but I am not a big fan of those tests that check if class hasattr etc. so it's not a must have imo

scripts/python_api_state.py
sfc-gh-turbaszek: Simply a helper

python/src/snowflake/connector/connection.py
sfc-gh-turbaszek: Let's see, imo removal of any such property will be a BCR and needs to be documented

PR #363, opened 2026-02-25
SNOW-2467780: Add float feature tests

sfc-gh-rkowalski:
Adds JDBC E2E coverage for FLOAT type behavior and updates shared float Gherkin definitions to ensure scenario coverage validation is aligned.
Includes tests for casting, literals, special values (NaN, inf, -inf), boundary values, NULL handling, and large result sets for JDBC float scenarios.
Does not include parameter bindings as they are not supported yet


sfc-gh-merbel (approved):
As arctic owl is saying we should consider later also the negative test case scenarios.

sfc-gh-pfus (approved):

PR #364, opened 2026-02-25
Rust config api proposal

sfc-gh-jszczerbinski:
Focal points:
- ConfigSource trait - represents a config source. We can use param definitions to fetch values from the source along with all the required metadata that will allows us to build good errors.
- FromConfigSource trait - does the actual validation, gets the parameter values and throws informative errors if we fail to validate the parameter values. Each config struct should implement one.


sf_core/src/config/config_source.rs
```diff
@@ -0,0 +1,201 @@
...
+
+use snafu::{Location, Snafu};
+
+use crate::config::{
+    ConfigError, InvalidParameterValueSnafu, MissingParameterSnafu,
+    param_registry::{ParamDef, ValueType},
+    settings::Setting,
+};
+
+enum ConfigSettingMeta {
```
sfc-gh-jszczerbinski, line 11: Metadata about one setting Key -> Value

sf_core/src/config/config_source.rs
```diff
@@ -0,0 +1,201 @@
...
+        name: String,
+        value: toml::Value,
+    },
+    PythonKwargs {
+        name: String,
+        value: ValueType,
+    },
+}
+
+enum ConfigSourceMeta {
```
sfc-gh-jszczerbinski, line 34: Metadata about whole data source. Might be useful for missing parameters, debug logging config source etc.

sf_core/src/config/config_source.rs
```diff
@@ -0,0 +1,201 @@
...
+    EnvironmentVariables,
+    OdbcConnectionStrings { connection_string: String },
+    TomlFiles { path: String },
+}
+
+pub struct ConfigSetting {
```
sfc-gh-jszczerbinski, line 40: Information about retrieved param: should have everything that would allow us to construct an error

sf_core/src/config/config_source.rs
```diff
@@ -0,0 +1,201 @@
...
+            .flat_map(|source| source.meta())
+            .collect()
+    }
+}
+
+/*
+FromConfigSource, allows us to read a ConfigSource and return a specific type.
+*/
+pub trait FromConfigSource: Sized {
+    fn from_config_source(source: &Box<dyn ConfigSource>) -> Result<Self, Vec<ConfigError>>;
```
sfc-gh-jszczerbinski, line 200: Returns vec of errors since validation returns all errors at once if possible and not the earliest validation failures

sf_core/src/config/config_source.rs
```diff
@@ -0,0 +1,201 @@
...
+pub trait ConfigSource {
+    fn get(&self, parameter: &'static ParamDef) -> Option<ConfigSetting>;
+    fn get_required(
+        &self,
+        parameter: &'static ParamDef,
+    ) -> Result<ConfigSetting, Vec<ConfigError>> {
+        self.get(parameter).ok_or_else(|| {
+            vec![
+                MissingParameterSnafu {
+                    parameter: parameter.canonical_name.to_string(),
```
sfc-gh-jszczerbinski, line 95: Should probably be alias name for some of the sources, this is proposal after all

sf_core/src/config/connection_config.rs
```diff
@@ -35,6 +39,33 @@ pub struct ServerConfig {
     pub server_url: String,
 }
 
+/*
+FromConfigSource implementation for ServerConfig.
+Note this is naive implementation that fails early.
+We can do better by collecting all errors and returning them together.
+This is harder to implement but returning Vec<ConfigError> is more flexible and allows us to do that implementation.
+
+In theory this could automatically derived
```
sfc-gh-jszczerbinski, line 48: Not sure if this isn't an overkill, but it would allow us cut down on boilerplate

sf_core/src/config/param_registry.rs
```diff
@@ -50,6 +13,8 @@ pub struct ParamDef {
     /// e.g. `&["SERVER", "HOST"]` all resolve to `"host"`.
     pub aliases: &'static [&'static str],
 
+    pub odbc_alias: Option<&'static str>,
```
sfc-gh-mhofman, line 16: a bit strange to have odbc-specific field in core

sfc-gh-jszczerbinski: This was PR for discussion with @sfc-gh-boler . But as a general note: It makes a lot of sense - you need this information since you want to have driver specific errors in core (So that driver says "Required parameter PWD not set in SnowflakeDSII DSN" not "Required parameter password not set" to ODBC user)

PR #368, opened 2026-02-26
NO-SNOW: Pin virtualenv <21.0 due to breaking changes

sfc-gh-turbaszek:


sfc-gh-fpawlowski (approved):
lgtm

.github/workflows/test-python.yml
sfc-gh-turbaszek:
```suggestion
        run: uv run --python ${{ matrix.py }} --with hatch --with cython --with setuptools --with "virtualenv<21.0.0" hatch build
```

.github/workflows/pre-commit.yml
```diff
@@ -48,17 +48,15 @@ jobs:
...
         uses: actions/setup-java@v4
         with:
           java-version: '17'
           distribution: 'temurin'
 
-      - name: Install pre-commit
-        run: pip install pre-commit
+      - name: Setup Python
+        run: |
+          pip install pre-commit hatch "virtualenv<21.0.0"
```
sfc-gh-fpawlowski, line 59: Do we intentionally use _pip_ and not _uv pip_? Why is it so?

sfc-gh-turbaszek: I don't know. I think we need to rethink how we setup the projects because single line change needs to be propagated multiple times

PR #374, opened 2026-02-26
SNOW-3074918 : Fix S3 file transfer for stage subdirectory uploads

sfc-gh-turbaszek:
Fix S3 file transfer for stage subdirectory uploads

The S3 file transfer was failing for PUT commands targeting stage
subdirectories (e.g. @stage/nested/subdir) because:

1. The endPoint from Snowflake's StageInfo was ignored, causing the S3
   client to derive its endpoint solely from the region. The old Python
   connector honored this field for FIPS and regional routing.

2. The key_prefix parsed from stageInfo.location was not normalized with
   a trailing slash, producing malformed S3 keys like
   "prefix/nested/subdirfile.csv" instead of
   "prefix/nested/subdir/file.csv".

3. The HeadObject existence check (used when overwrite=false) treated
   any non-404 error as a hard failure, including 403 Forbidden from
   limited temporary credentials.

4. The overwrite field was required in the Snowflake response, while the
   old connector defaulted to false when absent.

Also adds a subdirectory upload E2E test and the BackCompat BadRequest
error class.

### DIFF
--- a/sf_core/src/file_manager/mod.rs
+++ b/sf_core/src/file_manager/mod.rs
@@ -183,7 +183,10 @@ pub async fn download_single_file(
     // Create the full output path: local_location/src_location
     let output_path = Path::new(&data.local_location).join(&data.src_location);
 
-    // Save the compressed data to the constructed path
+    if let Some(parent) = output_path.parent() {
+        std::fs::create_dir_all(parent).context(IoSnafu)?;
+    }
+
     let mut output_file = File::create(&output_path).context(IoSnafu)?;
     output_file.write_all(&compressed_data).context(IoSnafu)?;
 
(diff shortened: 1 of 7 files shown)


sf_core/src/file_manager/mod.rs
```diff
@@ -180,10 +180,14 @@ pub async fn download_single_file(
         decrypt_file_data(&encrypted_data, &file_metadata, &data.encryption_material)
             .context(DecryptionSnafu)?;
 
-    // Create the full output path: local_location/src_location
-    let output_path = Path::new(&data.local_location).join(&data.src_location);
+    // Use only the filename (basename) from src_location to match the old connector
+    // behavior, which downloads files flat into the local directory regardless of
+    // any subdirectory structure in the stage.
```
sfc-gh-pczajka, line 185: Could you add shared test for this behavior? Easy to detect via reference whether it was python-specific or shared bug.

sfc-gh-pczajka (commented):
LGTM, let's add this test

sfc-gh-pczajka (approved):
Test scenario will be added in separate PR, as discussed

PR #376, opened 2026-02-27
Decouple protobuf from rest of the core

sfc-gh-jszczerbinski:
The goal here is make sure that protobuf layer is not tightly coupled from the rest of the core implementation.

We don't want tight coupling, because:  
\- Rust based wrappers don't need to use protobuf - we should have an option to expose rust only interface without extra dependencies  
\- We might pivot from protobuf to another interface layer at some point

How we avoid coupling:  
\- We make the protobuf interface an optional feature in sf_core crate  
\- We create CI check where we compile without this feature


sfc-gh-jszczerbinski:
* **#376** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/376?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/376?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/test-rust-core.yml
```diff
@@ -45,6 +45,17 @@ jobs:
       - name: Run clippy
         run: cargo clippy --workspace --all-targets --all-features -- -D warnings
 
+  build-without-protobuf:
```
sfc-gh-turbaszek, line 48: Why we build without protobuf?

sfc-gh-jszczerbinski: I updated the description

sfc-gh-boler (approved):
Neat approach. LGTM.

PR #259, opened 2026-02-09
INIT and GET basic session parameters

sfc-gh-pczajka:
Add connection cache (in rust) for session parameters. Behavior:
* add session parameters argument to connection parameters
* set up cache based on session parameters returned by server after authentication
* ConnectionGetParameterRequest always read from the cache
* cache is updated:
   * every time server returns parameter list
   * on ALTER SESSION queries (detected via regex) in case server does not return parameter list in query (based on experiments and python connector code, other drivers did not had this case covered)

PR #377, opened 2026-02-27
SNOW-3155976: Implement autocommit in python

sfc-gh-turbaszek:
Replaces the stub-only autocommit support with real SQL execution:
- commit() executes COMMIT, rollback() executes ROLLBACK
- set_autocommit() issues ALTER SESSION SET autocommit=<bool>
- autocommit kwarg in __init__ injected into session_parameters
- __exit__ uses try/finally to ensure close() always runs and
  skips commit/rollback when autocommit is ON

### DIFF
--- a/python/src/snowflake/connector/connection.py
+++ b/python/src/snowflake/connector/connection.py
@@ -48,12 +48,19 @@ ConnectionParameters = dict[str, ConnectionParamValue]
 class Connection:
     """Connection objects represent a database connection."""
 
-    def __init__(self, *, paramstyle: str | None = None, **kwargs: ConnectionParamValue) -> None:
+    def __init__(
+        self,
+        *,
+        paramstyle: str | None = None,
+        autocommit: bool | None = None,
+        **kwargs: ConnectionParamValue,
+    ) -> None:
         """
         Initialize a new connection object.
 
         Args:
             paramstyle: Binding style – ``"pyformat"`` (default), ``"format"``, ``"qmark"`` or ``"numeric"``
+            autocommit: Optional bool to enable/disable autocommit at connection time
             database: Database name
             user: Username
             password: Password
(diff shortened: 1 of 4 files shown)


python/src/snowflake/connector/connection.py
```diff
@@ -74,6 +78,19 @@ def __init__(self, *, paramstyle: str | None = None, **kwargs: ConnectionParamVa
         # Extract session_parameters before processing other kwargs
         session_params: SessionParameters | None = kwargs.pop("session_parameters", None)  # type: ignore
 
+        # TODO: _session_parameters should be read from sf_core instead of being tracked locally
+        self._session_parameters: dict[str, Any] = {}
```
sfc-gh-pczajka: It was introduced here: https://github.com/snowflakedb/universal-driver/pull/259
Let's use dedicated get_session_parameter instead of reimplementing that in PY

python/src/snowflake/connector/connection.py
```diff
@@ -74,6 +78,19 @@ def __init__(self, *, paramstyle: str | None = None, **kwargs: ConnectionParamVa
...
+
+        # Pop autocommit before the type-dispatch loop (bool is a subclass of int)
+        autocommit = kwargs.pop("autocommit", None)
```
sfc-gh-pczajka: Could we use named arg instead of popping a kwarg? It procides good differentiation between core-defined params (kwars) and params used by wrapper logic (named params)

python/src/snowflake/connector/connection.py
```diff
@@ -218,15 +224,27 @@ def ping(self) -> bool:
         """
         raise NotSupportedError("ping is not implemented")
 
-    def set_autocommit(self, autocommit: bool) -> None:
-        """
-        Set the autocommit mode.
+    @property
+    def _autocommit(self) -> bool:
+        return bool(self._session_parameters.get("AUTOCOMMIT", False))
```
sfc-gh-pczajka: Let's use get_session_parameter

sfc-gh-pczajka (changes requested):

sfc-gh-pczajka (approved):

PR #378, opened 2026-02-27
SNOW-3104121: Redact sensitive data in Debug output and logging

sfc-gh-makowalski:
Introduce a lightweight Redacted helper struct and redact() function to consistently mask secrets (passwords, tokens, SAML responses, private keys) in Debug implementations and log statements.

Changes:
- Add sf_core::redact module with Redacted struct and redact() helper
- Replace #[derive(Debug)] with custom Debug impls for NativeOktaConfig, LoginMethod, and AuthRequestData to mask sensitive fields
- Redact connection string logging in ODBC driver_connect (PWD, TOKEN, PRIV_KEY_FILE_PWD, PRIV_KEY_PWD, PRIV_KEY_BASE64)
- Replace login_request JSON dump with structured log showing only authenticator and login_name

### DIFF
--- a/odbc/src/api/connection.rs
+++ b/odbc/src/api/connection.rs
@@ -55,7 +55,10 @@ pub fn driver_connect(
                 let v = if REDACTED_KEYS.contains(&k.to_uppercase().as_str()) {
                     "****"
                 } else {
-                    connection_string_map.get(k).map(|s| s.as_str()).unwrap_or("")
+                    connection_string_map
+                        .get(k)
+                        .map(|s| s.as_str())
+                        .unwrap_or("")
                 };
                 (k, v)
             })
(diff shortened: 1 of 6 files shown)


sfc-gh-jszczerbinski (changes requested):
What about the approach with Secret<T> type? It seems more foolproof

sfc-gh-makowalski:
> What about the approach with Secret type? It seems more foolproof

That was another option, the problem with both is that everyone has to remember to use the dedicated type or to implement the debug and logging in a secure way. Last time I proposed a PR with the secret type we did not merge it because there was not enough benefit to it. I'm all for the secret type, should I refactor? @sfc-gh-jszczerbinski

sfc-gh-pbulawa:
I agree with Jakub. It seems that it is easier to use the Secret type and for me it's more elegant.

PR #325, opened 2026-02-20
Snow 2997744 fix empty result handling

sfc-gh-pczajka:
Fix missing schema error for empty results

Follow up to #323

PR #383, opened 2026-03-02
Unskip empty result tests in PY after SNOW-2997744 got fixed

sfc-gh-pczajka:
Follow up of #294 (instead of #325)


sfc-gh-fpawlowski (approved):
lgtm

PR #388, opened 2026-03-03
SNOW-2467847: Add missing parameter binding tests

sfc-gh-rkowalski:
## Summary
- Adds missing JDBC E2E parameter-binding coverage for `DECFLOAT`, `FLOAT`, `NUMBER`, and `STRING` type suites.
- Extends JDBC tests to validate both `SELECT` and `INSERT` paths with `PreparedStatement`, including explicit `NULL` bindings and type casts.
- Covers precision and boundary scenarios:
  - `DECFLOAT`: normal + extreme exponent values (`1E+16384`, `1E-16383`, `-1.234E+8000`)
  - `NUMBER`: standard precision and high-precision `NUMBER(38,0)/(38,2)` round-trip values
  - `STRING`: literal binding, Unicode/Japanese, whitespace/control chars, quotes, backslash, combining/surrogate edge cases
  - `FLOAT`: numeric bindings and null handling for float and synonyms
- Updates shared Gherkin definitions to include `@jdbc_e2e` tags on existing parameter-binding scenarios so JDBC now executes the same cross-driver scenarios already covered elsewhere.

### DIFF
--- a/jdbc/src/test/java/net/snowflake/jdbc/e2e/types/DecfloatTests.java
+++ b/jdbc/src/test/java/net/snowflake/jdbc/e2e/types/DecfloatTests.java
@@ -8,8 +8,10 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.math.BigDecimal;
 import java.sql.Connection;
+import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.Statement;
+import java.sql.Types;
 import java.util.Arrays;
 import java.util.List;
 import net.snowflake.client.SnowflakeIntegrationTestBase;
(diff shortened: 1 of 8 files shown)


jdbc/src/test/java/net/snowflake/jdbc/e2e/types/DecfloatTests.java
```diff
@@ -297,6 +299,128 @@ public void shouldDownloadLargeResultSetWithMultipleChunksFromTable() throws Exc
     }
   }
 
+  @Test
+  public void shouldSelectDecfloatUsingParameterBinding() throws Exception {
+    // Given Snowflake client is logged in
+    // When Query "SELECT ?::DECFLOAT, ?::DECFLOAT, ?::DECFLOAT" is executed with bound DECFLOAT
+    // values [123.456, -789.012, 42.0]
+    // Then Result should contain [123.456, -789.012, 42.0]
+    // When Query "SELECT ?::DECFLOAT" is executed with bound NULL value
```
sfc-gh-pfus, line 308: I think I already mentioned it somewhere, but it would be better to split it and assign to particular lines IMHO, unless you have a good reasoning to not do that?

jdbc/src/test/java/net/snowflake/jdbc/e2e/types/DecfloatTests.java
```diff
@@ -297,6 +299,128 @@ public void shouldDownloadLargeResultSetWithMultipleChunksFromTable() throws Exc
...
+        assertFalse(resultSet.next(), "Expected exactly one row for type: DECFLOAT");
+      }
+    }
+
+    try (PreparedStatement preparedStatement = connection.prepareStatement("SELECT ?::DECFLOAT")) {
+      preparedStatement.setNull(1, Types.DECIMAL);
+      try (ResultSet resultSet = preparedStatement.executeQuery()) {
+        assertTrue(resultSet.next(), "Expected one row for type: DECFLOAT");
+        assertNull(resultSet.getObject(1), "Column 1 should be NULL for DECFLOAT");
+        assertTrue(resultSet.wasNull(), "Column 1 should set wasNull() for DECFLOAT");
```
sfc-gh-pfus, line 333: Maybe it's worth teseting this at least sometimes when it returns false?

sfc-gh-pfus (approved):

sfc-gh-merbel (approved):

PR #389, opened 2026-03-03
SNOW-2467761: Add boolean feature tests

sfc-gh-rkowalski:
Introduces comprehensive `Boolean` E2E tests based on `boolean.feature`

### DIFF
--- a/jdbc/src/test/java/net/snowflake/jdbc/e2e/types/BooleanTests.java
+++ b/jdbc/src/test/java/net/snowflake/jdbc/e2e/types/BooleanTests.java
@@ -1,0 +1,247 @@
+package net.snowflake.jdbc.e2e.types;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.Statement;
+import java.sql.Types;
+import java.util.Arrays;
+import java.util.List;
+import net.snowflake.client.SnowflakeIntegrationTestBase;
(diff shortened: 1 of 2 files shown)


jdbc/src/test/java/net/snowflake/jdbc/e2e/types/BooleanTests.java
```diff
@@ -0,0 +1,247 @@
...
+        createTempTable(
+            connection,
+            "ud_boolean_",
+            "c1 " + BOOLEAN_TYPE + ", c2 " + BOOLEAN_TYPE + ", c3 " + BOOLEAN_TYPE);
+    execute(connection, "INSERT INTO " + tableName + " VALUES (TRUE, FALSE, TRUE)");
+    assertSingleRow(connection, "SELECT * FROM " + tableName, Arrays.asList(true, false, true));
+  }
+
+  @Test
+  public void shouldHandleNULLValuesFromTable() throws Exception {
```
sfc-gh-pfus, line 101: For me, again, such a test is not needed, as we already have the same thing tested in the above query. Having such a test, we test Snowflake itself, and not the contract between drivers and GS.

jdbc/src/test/java/net/snowflake/jdbc/e2e/types/BooleanTests.java
```diff
@@ -0,0 +1,247 @@
...
+    // When Query "SELECT * FROM <table>" is executed
+    // Then Result should contain [NULL, TRUE, FALSE] in any order
+    Connection connection = getDefaultConnection();
+    String tableName = createTempTable(connection, "ud_boolean_", "col " + BOOLEAN_TYPE);
+    execute(connection, "INSERT INTO " + tableName + " VALUES (NULL), (TRUE), (FALSE)");
+    assertBooleanCounts(connection, "SELECT col FROM " + tableName, 1, 1, 1);
+  }
+
+  @Test
+  public void shouldDownloadLargeResultSetWithMultipleChunksFromTable() throws Exception {
```
sfc-gh-pfus, line 114: If we want to test chunks, I think that such tests don't help much and introduce slowness - it may be cheaper to create a large query with all types returned at once.

sfc-gh-merbel (approved):

PR #392, opened 2026-03-04
NO-SNOW: Add build.rs to jdbc_bridge to emit loader rpaths via shared build_common.rs

sfc-gh-rkowalski:
- JNI loading depended on a macOS-only Gradle workaround that hardcoded an AWS-LC FIPS filename pattern and searched a specific build path.
- `jdbc_bridge` now embeds runtime search paths (`@loader_path-based`) for transitive native libraries.
- This makes loading more deterministic, version-agnostic, and easier to maintain across dependency updates.
- Security posture improves by avoiding process-wide `DYLD_LIBRARY_PATH` overrides and relying on explicit linker configuration.
- Verified by building `jdbc_bridge` (`cargo check -p jdbc_bridge`) after the change.

### DIFF
--- a/jdbc_bridge/build.rs
+++ b/jdbc_bridge/build.rs
@@ -1,0 +1,8 @@
+include!(concat!(env!("CARGO_MANIFEST_DIR"), "/../build_common.rs"));
+
+fn main() {
+    #[cfg(not(target_os = "windows"))]
+    {
+        emit_loader_rpaths();
+    }
+}
--- a/jdbc/build.gradle
+++ b/jdbc/build.gradle
@@ -165,13 +165,6 @@ def configureCommonTestTask = { t ->
                   'libjdbc_bridge.so'
     t.environment 'CORE_PATH', file("${libPath}/${libName}").absolutePath
     t.environment 'PARAMETER_PATH', file('../parameters.json').absolutePath
-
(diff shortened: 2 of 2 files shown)


jdbc_bridge/build.rs
```diff
@@ -0,0 +1,8 @@
+include!(concat!(env!("CARGO_MANIFEST_DIR"), "/../build_common.rs"));
+
+fn main() {
+    #[cfg(not(target_os = "windows"))]
```
sfc-gh-jszczerbinski, line 4: Why not:     if !cfg!(target_os = "windows") ? Are sure the semantics are the same with cross compilation?

sfc-gh-jszczerbinski (approved):

jdbc_bridge/build.rs
sfc-gh-rkowalski: We (me + llm) intentionally use `#[cfg(...)] instead of if cfg!(...)` because this must be conditional compilation, not a runtime-style branch. With `cfg!`, both branches are still type-checked, which breaks target-specific code. For cross-compilation, both forms use target cfg values, so behavior remains target-based.

sfc-gh-jszczerbinski: Ok LGTM

PR #395, opened 2026-03-04
Fix rownumber with fetchall behavior

sfc-gh-pczajka:
Fetchall did not update rownumber. This PR fixes that

### DIFF
--- a/python/src/snowflake/connector/cursor.py
+++ b/python/src/snowflake/connector/cursor.py
@@ -577,7 +577,9 @@ class SnowflakeCursorBase(abc.ABC):
         self._check_not_closed()
         if self._iterator is None:
             self._iterator = self._get_iterator()
-        return list(self._iterator)
+        rows = list(self._iterator)
+        self._rownumber += len(rows)
+        return rows
 
     # ------------------------------------------------------------------
     # PEP 249 optional / no-op methods
--- a/python/tests/unit/test_cursor.py
+++ b/python/tests/unit/test_cursor.py
@@ -644,3 +644,89 @@ class TestFetchmanyArraysizeAttribute:
             result = cursor.fetchmany()
 
         assert len(result) == 5
+
+
+class TestRownumber:
+    """Unit tests for Cursor.rownumber property."""
+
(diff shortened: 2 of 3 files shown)


python/src/snowflake/connector/cursor.py
```diff
@@ -577,7 +577,9 @@ def fetchall(self) -> list[Any]:
         self._check_not_closed()
         if self._iterator is None:
             self._iterator = self._get_iterator()
-        return list(self._iterator)
+        rows = list(self._iterator)
+        self._rownumber += len(rows)
```
sfc-gh-fpawlowski, line 581: Out of curiosity - what is our final goal here? Do we want _rownumber to be kept on the wrapper side or on the core side? If so, then we would be always just calling core for the most fresh value and update it there

python/tests/unit/test_cursor.py
```diff
@@ -644,3 +644,89 @@ def test_fetchmany_uses_instance_arraysize(self, cursor):
...
+
+    def test_rownumber_fetchall_on_empty_result(self, cursor):
+        """rownumber stays None when fetchall returns no rows."""
+        cursor._iterator = iter([])
+
+        with patch.object(cursor, "_get_iterator"):
+            cursor.fetchall()
+            assert cursor.rownumber is None
+
+    def test_rownumber_none_after_execute_resets(self, cursor):
```
sfc-gh-fpawlowski, line 723: should we have such integ test as well?

sfc-gh-fpawlowski (approved):
Overall looks good - left 2 comments

python/src/snowflake/connector/cursor.py
sfc-gh-pczajka: No, core responsibility ends on creating the arrow stream. Wrapper should be responsible for how it is being consumed by user

python/tests/unit/test_cursor.py
sfc-gh-pczajka: There is :)
`test_rownumber_fetchall_on_empty_result`

PR #405, opened 2026-03-05
2467760: Logical type support in ODBC

sfc-gh-ddas:


tests/definitions/odbc/types/boolean_conversion_to_c_real.feature
```diff
@@ -0,0 +1,48 @@
...
+
+  # ============================================================================
+  # SUCCESSFUL CONVERSIONS - Boolean to SQL_C_BINARY
+  # ============================================================================
+
+  @odbc_e2e
+  Scenario: should convert boolean to SQL_C_BINARY
+    Given Snowflake client is logged in
+    When Query "SELECT TRUE::BOOLEAN, FALSE::BOOLEAN" is executed
+    Then SQL_C_BINARY should return byte 0x01 for TRUE and 0x00 for FALSE
```
sfc-gh-makowalski: Should this test be here? Maybe we should move it to some other file, as well as the respective test case from `boolean_conversion_to_c_real.cpp`? I understand it's only one test so I'm not pushing for a new file, but allocating it here seems mesleading.

sfc-gh-jszczerbinski (approved):
LGTM

sfc-gh-makowalski (changes requested):

tests/definitions/odbc/types/boolean_conversion_to_c_real.feature
sfc-gh-ddas: I segregated it to another file. the reason I had it in the same one before is because string to binary conversion tests do the same, but I agree that this is more clear. I'll clean that one up in a future PR as it is unrelated to logical types.

sfc-gh-makowalski (approved):

PR #413, opened 2026-03-06
SNOW-2433865 separate driver images in perf tests, simplify recorded http tests cert handling

sfc-gh-akolodziejczyk:
1. Create separate images for UD/Old drivers (common base with only a difference in the driver installation) to avoid potential conflicts between drivers (there were concerns for Python)
2. The current approach for disabling cert validation in recorded HTTP tests was not working correctly for the OLD ODBC driver. Instead, wiremock generated cert will be added to the system CA bundle, as that solution works for all drivers

### DIFF
--- a/tests/performance/drivers/odbc/app/config.cpp
+++ b/tests/performance/drivers/odbc/app/config.cpp
@@ -57,15 +57,10 @@ std::map<std::string, std::string> parse_parameters_json() {
   std::string search_str = json_str;
 
   std::vector<std::pair<std::string, std::vector<std::string>>> key_mappings = {
-      {"account", {"SNOWFLAKE_TEST_ACCOUNT", "account"}},
-      {"host", {"SNOWFLAKE_TEST_HOST", "host"}},
-      {"user", {"SNOWFLAKE_TEST_USER", "user"}},
-      {"database", {"SNOWFLAKE_TEST_DATABASE", "database"}},
-      {"schema", {"SNOWFLAKE_TEST_SCHEMA", "schema"}},
-      {"warehouse", {"SNOWFLAKE_TEST_WAREHOUSE", "warehouse"}},
-      {"role", {"SNOWFLAKE_TEST_ROLE", "role"}},
-      {"verify_certificates", {"verify_certificates"}},
-      {"verify_hostname", {"verify_hostname"}}};
+      {"account", {"SNOWFLAKE_TEST_ACCOUNT", "account"}}, {"host", {"SNOWFLAKE_TEST_HOST", "host"}},
+      {"user", {"SNOWFLAKE_TEST_USER", "user"}},          {"database", {"SNOWFLAKE_TEST_DATABASE", "database"}},
+      {"schema", {"SNOWFLAKE_TEST_SCHEMA", "schema"}},    {"warehouse", {"SNOWFLAKE_TEST_WAREHOUSE", "warehouse"}},
+      {"role", {"SNOWFLAKE_TEST_ROLE", "role"}}};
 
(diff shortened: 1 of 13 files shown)


tests/performance/drivers/odbc/app/config.cpp
```diff
@@ -57,15 +57,10 @@ std::map<std::string, std::string> parse_parameters_json() {
...
-      {"user", {"SNOWFLAKE_TEST_USER", "user"}},
-      {"database", {"SNOWFLAKE_TEST_DATABASE", "database"}},
-      {"schema", {"SNOWFLAKE_TEST_SCHEMA", "schema"}},
-      {"warehouse", {"SNOWFLAKE_TEST_WAREHOUSE", "warehouse"}},
-      {"role", {"SNOWFLAKE_TEST_ROLE", "role"}},
-      {"verify_certificates", {"verify_certificates"}},
-      {"verify_hostname", {"verify_hostname"}}};
+      {"account", {"SNOWFLAKE_TEST_ACCOUNT", "account"}}, {"host", {"SNOWFLAKE_TEST_HOST", "host"}},
+      {"user", {"SNOWFLAKE_TEST_USER", "user"}},          {"database", {"SNOWFLAKE_TEST_DATABASE", "database"}},
+      {"schema", {"SNOWFLAKE_TEST_SCHEMA", "schema"}},    {"warehouse", {"SNOWFLAKE_TEST_WAREHOUSE", "warehouse"}},
```
sfc-gh-pcyrek, line 62: do we intentionally format like this?

tests/performance/drivers/odbc/build.sh
```diff
@@ -57,12 +57,23 @@ echo "${RUST_VERSION}" > tests/performance/.tmp/rust_version
...
+echo "→ Building old driver image..."
 docker build -f tests/performance/drivers/odbc/Dockerfile \
   --build-arg BUILDPLATFORM="${BUILDPLATFORM}" \
-  -t odbc-perf-driver:latest .
+  --target old \
+  -t odbc-perf-driver-old:latest .
 
 echo ""
-echo "✓ Build complete: odbc-perf-driver:latest"
+echo "✓ Built: odbc-perf-driver-old:latest"
```
sfc-gh-pcyrek, line 79: typo?

tests/performance/drivers/odbc/app/config.cpp
sfc-gh-akolodziejczyk: formatter changed it like this

tests/performance/drivers/odbc/build.sh
sfc-gh-akolodziejczyk: I think it's ok, Built is also correct

tests/performance/wiremock/extensions/ResponseTimeExtension.java
```diff
@@ -34,94 +45,77 @@ public String getName() {
...
+            return;
         }
         
         if (responseTime == 0) {
-            // This happens when request timestamp equals current time (same millisecond)
-            // Use 1ms as minimum to avoid division by zero and indicate request was processed
             responseTime = 1;
         }
         
         if (responseTime > MAX_RESPONSE_TIME_MS) {
```
sfc-gh-pcyrek: else if? minor

sfc-gh-pcyrek (approved):
lgtm

tests/performance/wiremock/extensions/ResponseTimeExtension.java
sfc-gh-akolodziejczyk: I simplified the whole function

PR #424, opened 2026-03-06
[core] PUT operation on GCP

sfc-gh-pczajka:
Add Google Cloud Storage (GCS) upload and download support to the universal   
  driver's Rust core (sf_core).                                                 
                                                                                
  This extends the existing S3-only file transfer to support multiple cloud     
  providers via a dispatch pattern. The implementation was designed by analyzing
   how JDBC, ODBC, and Python Snowflake drivers handle GCS uploads, then
  converging on a unified approach that preserves cross-driver behavioral
  compatibility.

  Changes

  New files:
  - sf_core/src/file_manager/gcs_transfer.rs — GCS upload, download, and file
  existence check via reqwest direct HTTP
  - sf_core/src/file_manager/s3_transfer.rs — S3 code extracted from
  file_transfer.rs (no logic changes)

  Modified files:
  - sf_core/src/file_manager/types.rs — StageLocationType enum, CloudCredentials
   enum (replaces S3-only Credentials), CredentialRefreshFn callback type
  - sf_core/src/file_manager/file_transfer.rs — refactored to cloud-agnostic
  dispatch layer
  - sf_core/src/file_manager/mod.rs — wires up new modules, calls dispatch
  functions
  - sf_core/src/rest/snowflake/query_response.rs — activates previously unused
  locationType, GCS_ACCESS_TOKEN, useRegionalUrl, useVirtualUrl fields

  Design docs:
  - doc/gcp-upload/ — 20+ documents including driver analyses, cross-driver
  comparison, and implementation plan

  Key design decisions

  - Direct HTTP via reqwest instead of a GCS SDK — matches ODBC/Python approach,
   avoids new dependencies
  - No multipart upload for GCS — all three existing drivers use single PUT
  regardless of file size
  - Empty Content-Encoding header on all GCS uploads — prevents GCS transparent
  decompression which would break the encrypt-then-compress pipeline
  - Enum-based CloudCredentials — compile-time guarantee that the right
  credential type is used per provider
  - Retry via existing execute_with_retry — 7 attempts, 1s base, 16s cap,
  decorrelated jitter
  - Token refresh as error propagation — TokenExpired error propagates to
  caller; CredentialRefreshFn callback type defined for future wiring from
  connection layer
  - Token-only auth for MVP — presigned URL mode deferred (clear error if
  GCS_ACCESS_TOKEN absent)
  - Download extracts metadata from HTTP headers — GCS returns encryption data
  in x-goog-meta-* headers (vs S3 object metadata). Headers cloned before
  consuming response body.

  Test plan

  - All 227 existing sf_core unit tests pass (no S3 regressions)
  - cargo clippy -D clippy::all clean
  - 6 new unit tests: URL construction (standard/virtual/regional), encryption
  JSON parsing (valid/missing key/missing IV)
  - Integration tests with wiremock (follow-up)
  - E2E tests with real GCS-backed Snowflake stage (follow-up)

### DIFF
--- a/sf_core/src/file_manager/file_transfer.rs
+++ b/sf_core/src/file_manager/file_transfer.rs
@@ -1,217 +1,74 @@
-use super::types::{EncryptedFileMetadata, EncryptionResult, MaterialDescription, StageInfo};
-use snafu::{Location, OptionExt, ResultExt, Snafu};
+// Cloud-agnostic dispatch layer for file transfers
+// This module routes upload/download operations to the appropriate cloud provider
 
-// AWS SDK imports
-use aws_config::{BehaviorVersion, Region};
-use aws_credential_types::Credentials;
-use aws_sdk_s3::error::SdkError;
-use aws_sdk_s3::{Client as S3Client, primitives::ByteStream};
+use super::gcs_transfer;
+use super::s3_transfer;
+use super::types::{EncryptedFileMetadata, EncryptionResult, StageInfo, StageLocationType};
+use snafu::{Location, ResultExt, Snafu};
 
-const SNOWFLAKE_UPLOAD_PROVIDER: &str = "snowflake-upload";
-const SNOWFLAKE_DOWNLOAD_PROVIDER: &str = "snowflake-download";
-const CONTENT_TYPE_OCTET_STREAM: &str = "application/octet-stream";
-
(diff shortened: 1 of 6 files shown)


sfc-gh-pczajka:
Duplicated in #580

PR #435, opened 2026-03-09
SNOW-3111876 Add timestamp NTZ conversion from JSON to Arrow

sfc-gh-pfus:
This PR introduces timestamp NTZ conversion from JSON to Arrow. It works for both variants: short scale when timestamp is returned as single int64 array and for larger scale when the struct (int64 epoch + int32 fraction) is used.


sfc-gh-pfus:
* **#482** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/482?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#435** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/435?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/435?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sf_core/src/arrow_utils.rs
```diff
@@ -114,10 +138,7 @@ fn create_column_array(
     row_type: &RowType,
 ) -> Result<(Field, Arc<dyn Array>), ArrowUtilsError> {
     match row_type {
-        RowType::Text { .. } => Ok((
-            create_field_with_type(row_type, DataType::Utf8),
-            Arc::new(StringArray::from(values)),
-        )),
+        RowType::Text { .. } => Ok((create_field(row_type), Arc::new(StringArray::from(values)))),
```
sfc-gh-jszczerbinski: Let's keep it explicit?

sf_core/src/arrow_utils.rs
```diff
@@ -215,37 +236,78 @@ fn create_column_array(
                 })
                 .collect();
             Ok((
-                create_field_with_type(row_type, DataType::Date32),
+                create_field(row_type),
```
sfc-gh-jszczerbinski: Should we keep the type explicit? Easier to read IMHO

sf_core/tests/integration/query/json_result_set.rs
sfc-gh-jszczerbinski, line 1: Why did we move this to integration?

sf_core/src/arrow_utils.rs
```diff
@@ -215,37 +236,78 @@ fn create_column_array(
...
+                        Arc::new(arrow::array::PrimitiveArray::<Int32Type>::from(
+                            normalized_fraction_values,
+                        ));
+                    let values = vec![
+                        (fields[0].clone(), epoch_array),
+                        (fields[1].clone(), fraction_array),
+                    ];
+                    Ok((field, Arc::new(arrow::array::StructArray::from(values))))
+                }
+                _ => panic!("Unexpected data type for TIMESTAMP_NTZ column"),
```
sfc-gh-jszczerbinski: We should never panic in sf_core

sf_core/src/arrow_utils.rs
```diff
@@ -33,22 +38,30 @@ pub fn create_field_with_type(row_type: &RowType, data_type: DataType) -> Field
             metadata.insert("logicalType".to_string(), "FIXED".to_string());
             metadata.insert("scale".to_string(), scale.to_string());
             metadata.insert("precision".to_string(), precision.to_string());
-            Field::new(name, data_type, *nullable).with_metadata(metadata)
+            Field::new(
+                name,
+                data_type.expect("data type is required for FIXED column"),
```
sfc-gh-jszczerbinski: Expect panics, let's not use it?

sf_core/src/arrow_utils.rs
```diff
@@ -7,9 +7,14 @@ use snafu::{Location, ResultExt, Snafu};
...
 
+/// Creates an Arrow Field from a RowType, embedding Snowflake-like metadata
+pub fn create_field(row_type: &RowType) -> Field {
+    create_field_with_type(row_type, None)
+}
+
 /// Creates an Arrow Field from a RowType, embedding Snowflake-like metadata
 /// Takes specific_data_type to allow overriding the default type inference for FIXED types based on scale/precision
-pub fn create_field_with_type(row_type: &RowType, data_type: DataType) -> Field {
+pub fn create_field_with_type(row_type: &RowType, data_type: Option<DataType>) -> Field {
```
sfc-gh-jszczerbinski: When do we need this to be optional?

sfc-gh-jszczerbinski (changes requested):
In general looks good to me, but:
- Let's revert to required non-optional data type for create_field_with_type, it's much more readable IMHO
- Don't panic in core

sf_core/src/arrow_utils.rs
sfc-gh-pfus: You mean data type? I don't think it is needed. This is only a hint if the type is ambiguous. For strings, booleans and so on, the type is obvious. However, it is unknown for FIXED columns, where having a `row_type` only is not enough, because we have to decide the final type based on values in a particular batch.
Moreover, this data type is simple here, for strings or booleans. But for complex types like timestamps, it's IMHO the responsibility of `create_field` function to decide the data type.

sf_core/tests/integration/query/json_result_set.rs
sfc-gh-pfus: To be honest, I don't see a point in keeping e2e test, because it requires  constant syncing with feature file. And this is a strictly core thing. WDYT?

sf_core/src/arrow_utils.rs
sfc-gh-pfus: I can change it, but it should never happen. This is more like unreachable block, only Rust doesn't know that 🤷

sf_core/src/arrow_utils.rs
sfc-gh-pfus: As mentioned earlier - because for now only FIXED columns data type cannot be inferred without knowledge about data inside.

sfc-gh-jszczerbinski (approved):

sfc-gh-pfus:
### Merge activity

* **Mar 12, 9:58 PM UTC**: @sfc-gh-pfus merged this pull request with [Graphite](https://app.graphite.com/github/pr/snowflakedb/universal-driver/435).

PR #437, opened 2026-03-09
SNOW-3223792: Implementation of SQLFreeStmt

sfc-gh-ddas:


odbc/src/c_api.rs
```diff
@@ -87,7 +87,10 @@ pub unsafe extern "C" fn SQLFreeStmt(
     statement_handle: sql::Handle,
     option: sql::FreeStmtOption,
```
sfc-gh-jszczerbinski, line 88: We are assuming that this correct enum value, let's accept integer here and implement TryFrom for FreeStmtOption so we can validate the value. Let's do it c_api module and pass enum to api module. I am making changes so we do that for every ODBC call

odbc_tests/tests/odbc-api/terminating_statement/free_stmt_tests.cpp
```diff
@@ -6,7 +6,6 @@
```
sfc-gh-jszczerbinski, line 6: Is enabling test enough here, I think we should test more interactions?

This what cursor suggested to me:

# Missing SQLFreeStmt Tests

## High Priority — Spec-Required Behavior

### 1. SQL_CLOSE preserves column bindings
The spec says SQL_CLOSE only closes the cursor; it does not unbind columns. Bind a column, execute a query, call SQL_CLOSE, re-execute, fetch, and verify the column binding is still populated with the new result.

### 2. SQL_CLOSE preserves parameter bindings
SQL_CLOSE should not reset parameters. Prepare a parameterized query, bind a parameter, execute, SQL_CLOSE, execute again without re-binding, and verify the parameter is still used.

### 3. SQL_UNBIND does not close cursor
Per the ODBC state transition table, SQL_UNBIND is a no-op in cursor states (S5–S7). Execute a query, fetch a row, call SQL_UNBIND, then continue fetching with SQLGetData — the cursor should still be open and functional.

### 4. SQL_RESET_PARAMS does not close cursor
Same as above but for SQL_RESET_PARAMS. Execute, fetch, call SQL_RESET_PARAMS, then continue fetching — cursor should remain open.

### 5. SQL_CLOSE on prepared-but-not-executed statement (S2/S3 no-op)
The spec says SQL_CLOSE has no effect when called in the prepared state (no cursor is open). Prepare a statement, call SQL_CLOSE, then execute — the prepared statement should still be valid.

## Medium Priority — Edge Cases & Robustness

### 6. SQL_CLOSE called multiple times consecutively
Verify idempotency. Execute a query, then call SQL_CLOSE three times in a row — all calls should return SQL_SUCCESS.

### 7. SQL_CLOSE after DML statement (no result set / S4 state)
Current tests only use SELECT statements. Execute a DML statement (INSERT/UPDATE), then call SQL_CLOSE. Verify it succeeds and the statement handle can be reused for another query.

### 8. SQL_UNBIND with multiple bound columns
Current test only binds a single column. Bind three or more columns, call SQL_UNBIND, execute, fetch, and verify none of the bindings are populated.

### 9. SQL_RESET_PARAMS with multiple parameters
Current test only binds a single parameter. Bind multiple parameters, call SQL_RESET_PARAMS, then execute — should fail with SQLSTATE 07002 (missing parameters).

## Lower Priority — Interaction & Metadata Tests

### 10. SQLNumResultCols after SQL_CLOSE on a prepared statement
Prepare a statement, execute, SQL_CLOSE, then call SQLNumResultCols. Per the spec, column metadata should still be available because the statement remains in the prepared state (S3).

### 11. SQL_UNBIND followed by re-bind
Unbind all columns, then re-bind to different columns or different buffers. Execute, fetch, and verify the new bindings work correctly.

### 12. Diagnostic record after SQLFreeStmt error
Now that diagnostics are wired up in the C API layer, trigger an error condition and verify that SQLGetDiagRec returns a meaningful SQLSTATE and message.

sfc-gh-jszczerbinski (changes requested):
Take a look at: https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/statement-transitions?view=sql-server-ver17#sqlfreestmt

We cannot transition to created state every time.

odbc/src/api/statement.rs
sfc-gh-pbulawa, line 381: It's not directly related to this PR, but per ODBC spec, `SQL_CLOSE` on a statement in state S2 (prepared, no open cursor) should remain in S2. The current implementation always resets to Created.

Can you add this as a part of the PR?

odbc_tests/tests/odbc-api/terminating_statement/free_stmt_tests.cpp
sfc-gh-pbulawa: This. Also,
SQL_CLOSE on a prepared-but-not-executed statement: The existing test "SQL_CLOSE preserves prepared statement" only tests Prepare → Execute → Close → Execute. There's no test for Prepare → Close → Execute (closing without ever executing).

sfc-gh-pbulawa (changes requested):

PR #462, opened 2026-03-11
[1/7] config: add native Bool variant to Setting enum and wire through proto/RPC layer

sfc-gh-boler:
Boolean config values (TLS flags, CRL toggles) were previously stored
and transmitted as the strings `"true"` / `"false"`, making
boolean settings stringly typed throughout the stack and requiring
consumers to parse them manually. This change adds a `Bool(bool)`
variant to `Setting`, extends the `ConfigSetting` protobuf oneof with
`bool_value`, adds `SetOptionBool` RPCs for database, connection, and
statement handles, and updates the TOML loader to preserve boolean
values instead of stringifying them.

The wire format now matches the logical type of these settings, while
later changes migrate higher-level config handling and wrapper call
sites onto the new representation.


sfc-gh-jszczerbinski (approved):

sfc-gh-boler:
The failing odbc_reference_tests uses SnowflakeDSII.dll, so that's the old driver. I'm going to ignore that failure.

PR #463, opened 2026-03-11
[2/7] config: centralize parameter definitions and validate batched options

sfc-gh-boler:
The driver previously had no single source of truth for supported
connection parameters, and configuring a handle required one
`SetOption*` RPC per setting. This change introduces
`param_registry.rs` as the shared table of canonical names, aliases,
types, defaults, sensitivity, and deprecation metadata, then builds the
new batched `SetOptions` path for connection, database, and statement
handles on top of that registry.

A new `validation.rs` module resolves submitted keys through the
registry, canonicalizes aliases, coerces unambiguous string values,
returns warnings for deprecated or unknown parameters, and rejects
invalid batches atomically. It also fails alias collisions that collapse
to the same canonical parameter so wrapper behavior no longer depends on
map iteration order. The legacy per-type setters remain for compatibility
until the final cleanup change removes them.

### DIFF
--- a/sf_core/src/config/param_registry.rs
+++ b/sf_core/src/config/param_registry.rs
@@ -1,0 +1,600 @@
+use std::collections::HashMap;
+use std::fmt;
+
+use std::sync::LazyLock;
+
+use crate::config::settings::Setting;
+
+/// A strongly-typed wrapper around a canonical parameter name.
+///
+/// Provides compile-time safety over bare `&str` keys while remaining
+/// zero-cost at runtime (it is `Copy` and stores a `&'static str`).
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct ParamKey(pub(crate) &'static str);
+
+impl ParamKey {
+    pub const fn as_str(&self) -> &'static str {
+        self.0
+    }
+}
+
+impl fmt::Display for ParamKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(self.0)
+    }
+}
+
+impl From<ParamKey> for String {
+    fn from(key: ParamKey) -> String {
+        key.0.to_owned()
+    }
+}
+
+impl AsRef<str> for ParamKey {
+    fn as_ref(&self) -> &str {
+        self.0
+    }
+}
+
+/// Canonical parameter name constants.
+///
+/// Use these instead of bare string literals when referencing parameter names
+/// in production code.  This gives compile-time typo detection and
+/// find-all-references support.
+pub mod param_names {
+    use super::ParamKey;
+
+    pub const ACCOUNT: ParamKey = ParamKey("account");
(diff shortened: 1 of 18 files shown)


sf_core/src/config/param_registry.rs
```diff
@@ -0,0 +1,598 @@
...
+
+/// Canonical parameter name constants.
+///
+/// Use these instead of bare string literals when referencing parameter names
+/// in production code.  This gives compile-time typo detection and
+/// find-all-references support.
+pub mod param_names {
+    use super::ParamKey;
+
+    pub const ACCOUNT: ParamKey = ParamKey("account");
```
sfc-gh-jszczerbinski, line 47: So can we get rid of ParamKey and just use ParamDef?

sf_core/src/config/param_registry.rs
```diff
@@ -0,0 +1,598 @@
...
+}
+
+/// Defines a single supported configuration parameter.
+pub struct ParamDef {
+    /// The canonical key name used internally (e.g. `"host"`).
+    pub canonical_name: &'static str,
+
+    /// Alternative names accepted from wrappers (case-insensitive lookup).
+    /// e.g. `&["SERVER", "HOST"]` all resolve to `"host"`.
+    pub aliases: &'static [&'static str],
```
sfc-gh-jszczerbinski, line 88: We shouldn't accept odbc aliases in Python. How will we do that?

sf_core/src/config/param_registry.rs
```diff
@@ -0,0 +1,598 @@
...
+    Bool,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Required {
+    /// Always required (e.g. `account`).
+    Always,
+    /// Required only when the authenticator matches (e.g. `password` for
+    /// `SNOWFLAKE_PASSWORD`).
+    WhenAuthMethod(&'static str),
```
sfc-gh-jszczerbinski, line 126: I like that we do some validation at early stage, but I don't think we should express more complex constraints at this level

sfc-gh-jszczerbinski (approved):

sf_core/src/config/param_registry.rs
sfc-gh-boler: Discussed offline: ParamDef stores a function reference, so it can't be easily stored in a vector. We may drop the validation function in the future (depending how validation patterns will look like across various core components) which will enable us to replace ParamKey with ParamDef.

sf_core/src/config/param_registry.rs
sfc-gh-boler: Added a TODO. Once we have wrapper identity (needed for ClientInfo), we can extend this.

sf_core/src/config/param_registry.rs
sfc-gh-boler: Discussed offline. We will revisit once we develop core more and gain experience. That will enable elimination of ParamKey.

PR #464, opened 2026-03-11
[3/7] config: type and validate resolved connection settings

sfc-gh-boler:
Connection setup was still driven by loosely merged settings and ad hoc
string reads, which made it hard to apply connect-time defaults
consistently, validate them in one place, and honor scope and mutability
rules after connect. This change introduces a typed `ConnectionConfig`
built from a layered `resolver`, extends parameter metadata with scope
and connect-time mutability, and routes connection and statement option
handling through that model so connect-time validation and runtime
behavior agree.

To support that flow, the protobuf/API layer now exposes structured
validation results, maps typed config validation failures to the right
driver error/status codes, and adds `connection_validate_options`. The
change also fixes Arrow chunk-result routing for large LOB queries so
chunked results still produce rows when headers are omitted, updates the
CRL test to set `crl_mode` before connect because it is a
connection-scoped immutable-after-connect setting, and adds blocking
protobuf-client adapters used only by synchronous Rust test helpers and
small smoke tests. The ODBC `string_lob` execution-order stabilization
needed for the full suite is intentionally split into the parent change.

### DIFF
--- a/sf_core/src/apis/database_driver_v1/connection.rs
+++ b/sf_core/src/apis/database_driver_v1/connection.rs
@@ -8,50 +8,21 @@ use tokio::sync::RwLock as AsyncRwLock;
 use super::Setting;
 use super::error::*;
 use super::global_state::DatabaseDriverV1;
-use super::validation::{ValidationIssue, normalize_host_underscores, resolve_and_apply_options};
+use super::validation::{
+    ValidationIssue, ValidationSeverity, canonicalize_setting_key, normalize_host_underscores,
+    resolve_options, validate_connection_seed_write, validate_session_override_write,
+};
 use crate::config::ParamStore;
-use crate::config::config_manager;
-use crate::config::param_registry::{ParamKey, param_names};
-use crate::config::path_resolver::ConfigPaths;
-use crate::config::rest_parameters::{ClientInfo, LoginMethod, LoginParameters};
+use crate::config::connection_config::ConnectionConfig;
+use crate::config::param_registry::{ParamKey, ParamScope, param_names};
+use crate::config::resolver;
+use crate::config::rest_parameters::{ClientInfo, LoginMethod, LoginParameters, QueryParameters};
 use crate::config::retry::RetryPolicy;
 use crate::handle_manager::Handle;
 use crate::rest::snowflake::{self, RestError, SessionTokens, SnowflakeResponseError};
 use crate::sensitive::SensitiveString;
 use crate::tls::client::create_tls_client_with_config;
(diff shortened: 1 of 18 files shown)


sf_core/src/apis/database_driver_v1/connection.rs
```diff
@@ -60,34 +28,25 @@ impl DatabaseDriverV1 {
...
-                    let login_parameters = LoginParameters::from_settings(&conn.settings)
-                        .context(ConfigurationSnafu)?;
+                let (config, host, port, client_info, init_params) = {
+                    let conn = conn_ptr.lock().await;
+                    // TODO(sfc-gh-boler): Clone the mutable connection inputs under the mutex,
+                    // then drop the lock before calling resolve/build. Those paths can do
+                    // synchronous disk I/O and private-key parsing, which currently extends the
+                    // connection mutex critical section and can block the async runtime thread.
+                    let resolved = conn.resolved_settings().context(ConfigurationSnafu)?;
+                    let config = ConnectionConfig::build(&resolved).context(ConfigurationSnafu)?;
```
sfc-gh-jszczerbinski, line 43: This make strong assumption that the config doesn't change over the lifetime of Connection.   
This is not always true:  
\- https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/connection.py#L905-L1116  
\- For standard ODBC connection attributes (see the table): https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlsetconnectattr-function?view=sql-server-ver17#comments

sfc-gh-jszczerbinski (approved):

sf_core/src/protobuf/apis/database_driver_v1.rs
```diff
@@ -1377,3 +1456,197 @@ pub type DatabaseDriverClient =
 pub fn database_driver_client() -> DatabaseDriverClient {
     DatabaseDriverClient::new(crate::protobuf::apis::RustTransport::new())
 }
+
+// Synchronous convenience wrappers used by Rust test helpers and small
+// in-process smoke tests. Production callers should prefer the async
+// client methods directly.
+static BLOCKING_CLIENT_RUNTIME: LazyLock<tokio::runtime::Runtime> = LazyLock::new(|| {
+    tokio::runtime::Builder::new_current_thread()
```
sfc-gh-jszczerbinski: Use multi_threaded_runtime instead

sfc-gh-jszczerbinski (approved):
Looks good to me, not sure about the session scope

PR #465, opened 2026-03-11
[4/7] python: migrate connection setup to batch SetOptions RPC

sfc-gh-boler:
Opening a connection previously made one `connection_set_option_*` RPC
call per parameter and sent boolean values as strings rather than using
the native wire type. This change replaces that per-key dispatch loop
with a single pass that builds a `dict[str, ConfigSetting]` using the
appropriate protobuf field for each Python type and submits the whole map
through `connection_set_options`.

Any `ValidationIssue` warnings returned by the core are forwarded via
`warnings.warn`, and the regenerated Python protobuf bindings expose the
batched option-setting and validation RPCs introduced earlier in the
stack. The result is fewer FFI round-trips during connection setup and a
wrapper that matches the core's typed config surface.


sfc-gh-turbaszek (approved):
LGTM 👌

PR #466, opened 2026-03-11
[5/7] jdbc: migrate connection setup to batch SetOptions RPC

sfc-gh-boler:
The JDBC wrapper previously looped over connection properties and made
one `connectionSetOption*` RPC call per entry, while also stringifying
some non-string values on the way across the FFI boundary. This change
replaces that flow with a single pass that builds a
`Map<String, ConfigSetting>` and sends the entire option set in one
`connectionSetOptions` call.

`Boolean`, `Integer`, `String`, and `Double` values are now mapped to
their matching `ConfigSetting` builder methods, and
`ConnectionOptionsResolver` gains a helper for storing integer defaults
without boxing them through `String`. The regenerated Java protobuf
bindings are updated to expose the batched option-setting and validation
RPCs added earlier in the stack.

### DIFF
--- a/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/SnowflakeConnectionImpl.java
+++ b/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/SnowflakeConnectionImpl.java
@@ -19,6 +19,7 @@ import java.sql.Savepoint;
 import java.sql.Statement;
 import java.sql.Struct;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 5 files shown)


sfc-gh-rkowalski (approved):

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/SnowflakeConnectionImpl.java
```diff
@@ -70,41 +71,34 @@ public SnowflakeConnectionImpl(String url, Properties properties) throws SQLExce
...
-              }
+              optionsMap.put(
+                  keyStr, ConfigSetting.newBuilder().setStringValue((String) value).build());
+            } else if (value instanceof Integer) {
+              optionsMap.put(
+                  keyStr, ConfigSetting.newBuilder().setIntValue((Integer) value).build());
+            } else if (value instanceof Boolean) {
+              optionsMap.put(
+                  keyStr, ConfigSetting.newBuilder().setBoolValue((Boolean) value).build());
+            } else if (value instanceof Double) {
```
sfc-gh-dheyman: we can add handling of `bytes_value` for parity and consistence

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/SnowflakeConnectionImpl.java
```diff
@@ -70,41 +71,34 @@ public SnowflakeConnectionImpl(String url, Properties properties) throws SQLExce
...
+              optionsMap.put(
+                  keyStr, ConfigSetting.newBuilder().setDoubleValue((Double) value).build());
             }
           });
+      if (!optionsMap.isEmpty()) {
+        ProtobufApis.databaseDriverV1.connectionSetOptions(
```
sfc-gh-dheyman: `ConnectionSetOptionsResponse ` contains `warnings` field. Let's log them accordingly

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/SnowflakeConnectionImpl.java
```diff
@@ -70,41 +71,34 @@ public SnowflakeConnectionImpl(String url, Properties properties) throws SQLExce
...
-                      .setValue((Integer) value)
-                      .build();
-              try {
-                ProtobufApis.databaseDriverV1.connectionSetOptionInt(request);
-              } catch (DatabaseDriverService.ServiceException e) {
-                throw new RuntimeException(e);
-              }
+              optionsMap.put(
+                  keyStr, ConfigSetting.newBuilder().setStringValue((String) value).build());
+            } else if (value instanceof Integer) {
```
sfc-gh-dheyman: `Long` values will fail this check

sfc-gh-dheyman: `int_value` is defined as int64 in .proto file

sfc-gh-dheyman (approved):

PR #489, opened 2026-03-12
ADR-28: Async Rust exposed at API level


sfc-gh-jszczerbinski:
## **ADR-28: Async Rust exposed at API level**

- Converts the protobuf API layer from synchronous to async/await.
- Replaced synchronization primitives: std::sync -> tokio::sync
- Refactors JDBC bridge and ODBC to manage their own tokio runtimes for bridging to synchronous C APIs. 
- Removes the centralized `async_bridge` module in favor of component-owned runtimes.

### Next up:

- Use common runtime for CRL
- Implement chunk prefetching

### DIFF
--- a/jdbc_bridge/src/lib.rs
+++ b/jdbc_bridge/src/lib.rs
@@ -14,6 +14,9 @@ struct JdbcBridge {
 impl JdbcBridge {
     pub fn new() -> Self {
         Self {
+            // Single worker thread is intentional: keeps contention minimal and
+            // makes deadlocks easier to detect. Will be increased during
+            // performance optimization.
             runtime: tokio::runtime::Builder::new_multi_thread()
                 .worker_threads(1)
                 .enable_all()
--- a/jdbc_bridge/src/lib.rs
+++ b/jdbc_bridge/src/lib.rs
@@ -1,8 +1,43 @@
+use std::sync::LazyLock;
+
 use jni::JNIEnv;
 use jni::objects::{JByteArray, JClass, JObject, JString, JValue};
 use jni::sys::{jint, jobject};
-use proto_utils::ProtoError;
-use sf_core::protobuf::apis::call_proto;
+use proto_utils::{ProtoError, Transport};
+use sf_core::protobuf::apis::RustTransport;
+
+struct JdbcBridge {
+    runtime: tokio::runtime::Runtime,
+    transport: RustTransport,
+}
+
+impl JdbcBridge {
+    pub fn new() -> Self {
+        Self {
+            runtime: tokio::runtime::Builder::new_multi_thread()
+                .worker_threads(1)
+                .enable_all()
+                .build()
+                .expect("Failed to create tokio runtime"),
+            transport: RustTransport::new(),
+        }
+    }
+
+    pub fn handle_message_sync(
+        &self,
+        service_name: &str,
+        method_name: &str,
+        request_bytes: Vec<u8>,
+    ) -> Result<Vec<u8>, ProtoError<Vec<u8>>> {
+        self.runtime.block_on(self.transport.handle_message(
+            service_name,
+            method_name,
+            request_bytes,
+        ))
+    }
+}
+
+static JDBC_BRIDGE: LazyLock<JdbcBridge> = LazyLock::new(JdbcBridge::new);
 
 mod slf4j_layer;
 
@@ -62,11 +97,10 @@ pub unsafe extern "system" fn Java_net_snowflake_client_internal_unicore_JNICore
         Err(_) => return std::ptr::null_mut(),
     };
 
-    // Call the protobuf API
-    let result = call_proto(
+    let result = JDBC_BRIDGE.handle_message_sync(
         &service_name_str.to_string_lossy(),
         &method_name_str.to_string_lossy(),
-        request_bytes_vec.as_slice(),
+        request_bytes_vec,
     );
 
     // Find the TransportResponse class
--- a/odbc/src/api/runtime.rs
+++ b/odbc/src/api/runtime.rs
@@ -4,6 +4,17 @@ use std::sync::RwLock;
 use sf_core::protobuf::apis::database_driver_v1::{DatabaseDriverClient, database_driver_client};
 use snafu::{Location, ResultExt, Snafu};
 
+/// Holds the shared tokio runtime and driver client used by all ODBC
+/// environments in this process.
+///
+/// ODBC requires an Environment handle before any Connection or Statement can
+/// be created. An application may allocate multiple Environments (e.g. for
+/// different global settings), but they all share the same underlying driver
+/// state. We therefore keep a single `OdbcGlobals` instance behind a
+/// reference-counted latch (`env_count`): the first `SQLAllocHandle(SQL_HANDLE_ENV)`
+/// creates it, and the last `SQLFreeHandle(SQL_HANDLE_ENV)` tears it down.
+/// On Windows the ODBC Driver Manager unloads the driver DLL after the last
+/// environment is freed, so we must shut down before that happens.
 pub struct OdbcGlobals {
     runtime: tokio::runtime::Runtime,
     client: DatabaseDriverClient,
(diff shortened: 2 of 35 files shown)


jdbc_bridge/src/lib.rs
```diff
@@ -62,11 +97,10 @@ pub unsafe extern "system" fn Java_net_snowflake_client_internal_unicore_JNICore
...
     };
 
-    // Call the protobuf API
-    let result = call_proto(
+    let result = JDBC_BRIDGE.handle_message_sync(
         &service_name_str.to_string_lossy(),
         &method_name_str.to_string_lossy(),
-        request_bytes_vec.as_slice(),
+        request_bytes_vec,
     );
```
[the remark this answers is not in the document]
sfc-gh-mhofman: this? :arrow_up:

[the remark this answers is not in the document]
sfc-gh-jszczerbinski: Valid concern, but out of scope for this PR - I am trying to minimize changes, the concern was also valid before the changes - for panics generated by core

odbc/src/api/runtime.rs
```diff
@@ -0,0 +1,92 @@
...
+            .context(RuntimeCreationSnafu)?;
+        let client = database_driver_client();
+        guard.globals = Some(OdbcGlobals { runtime, client });
+    }
+    Ok(())
+}
+
+pub fn env_freed() -> Result<(), OdbcRuntimeError> {
+    let mut guard = STATE.write().map_err(|_| LockPoisonedSnafu.build())?;
+    guard.env_count = guard.env_count.saturating_sub(1);
```
sfc-gh-mhofman, line 98: no logging

odbc/src/api/runtime.rs
```diff
@@ -0,0 +1,92 @@
...
+    let mut guard = STATE.write().map_err(|_| LockPoisonedSnafu.build())?;
+    guard.env_count += 1;
+    if guard.globals.is_none() {
+        let runtime = tokio::runtime::Builder::new_multi_thread()
+            .worker_threads(1)
+            .enable_all()
+            .build()
+            .context(RuntimeCreationSnafu)?;
+        let client = database_driver_client();
+        guard.globals = Some(OdbcGlobals { runtime, client });
```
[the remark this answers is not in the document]
sfc-gh-mhofman: might be valid @sfc-gh-jszczerbinski

odbc/src/api/runtime.rs
sfc-gh-mhofman: @sfc-gh-jszczerbinski but odbc layer does this logging, however if we had it here as well wouldn't hurt as serve for all wrappers

sf_core/src/protobuf/apis/mod.rs
```diff
@@ -4,21 +4,34 @@ use proto_utils::*;
...
-    fn handle_message(
+    async fn handle_message(
+        &self,
         service: &str,
         method: &str,
         message: Vec<u8>,
     ) -> Result<Vec<u8>, ProtoError<Vec<u8>>> {
-        call_proto(service, method, &message)
+        match service {
+            "DatabaseDriver" => self.driver.handle_message(method, message).await,
```
sfc-gh-mhofman, line 33: shouldn't version here be involved? but we can back to that later

sfc-gh-mhofman (approved):

sf_core/src/protobuf/apis/mod.rs
sfc-gh-jszczerbinski: Let's add it later if needed

odbc/src/api/runtime.rs
```diff
@@ -0,0 +1,92 @@
...
+
+pub fn global() -> Result<GlobalsGuard, OdbcRuntimeError> {
+    let guard = STATE.read().map_err(|_| LockPoisonedSnafu.build())?;
+    if guard.globals.is_none() {
+        return NotInitializedSnafu.fail();
+    }
+    Ok(GlobalsGuard(guard))
+}
+
+pub fn env_allocated() -> Result<(), OdbcRuntimeError> {
```
sfc-gh-boler, line 81: Why are we keeping this state as statics instead of a reference of Environment?



Nit: env_allocated and env_freed sound a bit like callbacks responding to some events, but those are more of new/free functions.

sf_core/src/protobuf/c_api.rs
```diff
@@ -50,7 +65,11 @@ pub unsafe extern "C" fn sf_core_api_call_proto(
             .to_string_lossy()
             .to_string();
         let message = std::slice::from_raw_parts(request, request_len);
-        call_proto(&api, &method, message)
+        STATE.runtime.block_on(
```
sfc-gh-boler, line 71: Do you envision changing this to support python async? At the moment we still block the thread. We would need some sort of python callback passing to Tokyo.

jdbc_bridge/src/lib.rs
```diff
@@ -1,8 +1,43 @@
...
+struct JdbcBridge {
+    runtime: tokio::runtime::Runtime,
+    transport: RustTransport,
+}
+
+impl JdbcBridge {
+    pub fn new() -> Self {
+        Self {
+            runtime: tokio::runtime::Builder::new_multi_thread()
+                .worker_threads(1)
```
sfc-gh-boler, line 21: How did you pick the number? A comment would be useufl.

sf_core/src/protobuf/c_api.rs
```diff
@@ -1,7 +1,22 @@
...
+use proto_utils::{ProtoError, Transport};
+
+struct CApiState {
+    runtime: tokio::runtime::Runtime,
+    transport: RustTransport,
+}
+
+static STATE: LazyLock<CApiState> = LazyLock::new(|| CApiState {
+    runtime: tokio::runtime::Builder::new_multi_thread()
+        .worker_threads(1)
```
sfc-gh-boler, line 17: Why 1? It would be helpful for the future us to have a brief explanation in code here.

sfc-gh-boler (approved):
LGTM, but I think it will need a few followups to truly take advantage of having the fully async core.

jdbc_bridge/src/lib.rs
sfc-gh-jszczerbinski: Deadlocks are most likely to happen with one thread, I want to keep it on one thread for now then I will increase amount of threads during perfomance optimization

sfc-gh-jszczerbinski: I'll add a comment

sf_core/src/protobuf/c_api.rs
sfc-gh-jszczerbinski: Part of motivation for those changes are discussions with Python team about asyncio. I expect us to use callback and spawn when implementing asyncio support. For now I am keeping changes minimal and try not to change the behavior.

odbc/src/api/runtime.rs
sfc-gh-jszczerbinski: I want to have shared Global State for all ODBC environments. env_allocated and env_freed are events that happen in AllocHandle or FreeHandle function and they correspond to ODBC Environment being freed or allocated. Freeing last env will cause Windows manager to unload the driver - I wanna catch this event and peform deinitialization before that happens.

sfc-gh-boler: Is the nature of ODBC that there is always only one query in flight? Are there logical connections, i.e. what if I connect to two different accounts from a single app? How will tha be modeled here?

sfc-gh-boler (approved):
(still LGTM)

odbc/src/api/runtime.rs
sfc-gh-jszczerbinski: So in odbc we have logical connections, but the concept of environments is more important here. Environment is a container for all "global" setting for a driver, before you create connection in ODBC you need to create environment. Usually you have one environment per application, but you can multiple in one application if for example you need different environment settings for parts of your app. For this code what's important is that allocating environment handle is the first odbc interaction for any application and deallocating is the last.

So having multiple connection to different accounts is irellevant here. We can have those within one environment or one environment per connection. What we care about is that first env allocation can be treated as init for our global state and any other calls will be made after that and last deallocation can be treated a as shutdown since no calls will be made after that.

sfc-gh-boler: Ok, so you want to have this global state (instead of per Environment) to share the underlying instance of the client which can multiplex logical connections.  
  
I think this is fine, but it would be great to document in the code why it's shared state, especiailly for folks less familiar with ODBC.

sfc-gh-boler (approved):

PR #516, opened 2026-03-16
NO-SNOW: Adjust the Linux-only ownership so it exits early when the test process is running as root

sfc-gh-rkowalski:


sf_core/src/token_cache/file_cache.rs
```diff
@@ -814,6 +814,11 @@ mod tests {
         #[cfg(target_os = "linux")]
         #[test]
         fn validate_file_fd_rejects_file_not_owned_by_current_user() {
+            let current_uid = unsafe { libc::getuid() };
```
sfc-gh-boler, line 820: Shouldn't this test just fail if run by root? Otherwise if somehow this test always runs as root, it will never really validate anything.

sfc-gh-rkowalski: This will led to false negatives and unstable builds. Maybe we should create different setup when running as root? Creating file with ownership different than root and using it instead of `/etc/hosts`?

sfc-gh-boler: What if we never run the tests as root? We can make them explicitly fail if we attempt to do so.

sfc-gh-rkowalski: Actually I'm looking for a work around for `Universal Driver Code Coverage` failures but maybe we shouldn't run our tests as root in the first place? https://github.com/snowflakedb/universal-driver/runs/67337939137

PR #620, opened 2026-03-19
SNOW-3286956: Add caching to workflows

sfc-gh-pbulawa:
###

### DIFF
--- a/.github/workflows/test-odbc.yml
+++ b/.github/workflows/test-odbc.yml
@@ -139,15 +139,6 @@ jobs:
           PARAMETERS_SECRET: ${{ secrets.PARAMETERS_SECRET }}
         shell: bash
 
-      - name: Cache C++ ODBC test build (reference)
-        uses: actions/cache@v5
-        with:
-
-          path: odbc_tests/cmake-build
-          key: ${{ runner.os }}-cpp-odbc-tests-old-${{ hashFiles('odbc_tests/**/*.cpp', 'odbc_tests/**/*.hpp', 'odbc_tests/**/*.h', 'odbc_tests/**/CMakeLists.txt') }}
-          restore-keys: |
-            ${{ runner.os }}-cpp-odbc-tests-old-
-
       - name: Run Old ODBC tests (Linux)
         if: matrix.os == 'ubuntu-latest'
         id: run_odbc_ref_tests_linux
@@ -303,15 +294,6 @@ jobs:
         run: cargo build
         env:
           RUSTFLAGS: "-C linker=rust-lld"
-
-      - name: Cache C++ ODBC test build
-        uses: actions/cache@v5
-        with:
-
-          path: odbc_tests/cmake-build
-          key: ${{ runner.os }}-cpp-odbc-tests-${{ hashFiles('odbc_tests/**/*.cpp', 'odbc_tests/**/*.hpp', 'odbc_tests/**/*.h', 'odbc_tests/**/CMakeLists.txt') }}
-          restore-keys: |
-            ${{ runner.os }}-cpp-odbc-tests-
 
       - name: Build and run ODBC tests (Unix)
         if: runner.os != 'Windows'
--- a/.github/workflows/test-odbc.yml
+++ b/.github/workflows/test-odbc.yml
@@ -36,7 +36,7 @@ jobs:
           path: |
             ~/.cargo/registry
             ~/.cargo/git
-          key: ${{ runner.os }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
+          key: ${{ runner.os }}-cargo-registry-odbc-unit-${{ hashFiles('**/Cargo.lock') }}
           restore-keys: |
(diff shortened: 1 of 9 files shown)


sfc-gh-pbulawa:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/620?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#620** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/620?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/620?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#626** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/626?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sfc-gh-jszczerbinski (approved):

.github/workflows/test-odbc.yml
```diff
@@ -111,6 +131,15 @@ jobs:
...
         shell: bash
 
+      - name: Cache C++ ODBC test build (reference)
+        uses: actions/cache@v5
+        with:
+
+          path: odbc_tests/cmake-build
+          key: ${{ runner.os }}-cpp-odbc-tests-old-${{ hashFiles('odbc_tests/**/*.cpp', 'odbc_tests/**/*.hpp', 'odbc_tests/**/*.h', 'odbc_tests/**/CMakeLists.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-cpp-odbc-tests-old-
```
sfc-gh-jszczerbinski: Why do we restore from any key? CMake is not very cacheable

sfc-gh-jszczerbinski (changes requested):
One comment

sfc-gh-jszczerbinski:
For any build: Why do we even append the hash at all, when restore keys never have a key

.github/workflows/test-python.yml
```diff
@@ -33,12 +33,91 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Cache Cargo registry
+        uses: actions/cache@v5
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+          key: ${{ runner.os }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
```
sfc-gh-turbaszek, line 60: How does the Cargo.lock version sf_core?

.github/workflows/test-python.yml
```diff
@@ -60,10 +139,17 @@ jobs:
     env:
       # Configure CMake for ARM64 on Windows ARM64 runners
       CMAKE_GENERATOR_PLATFORM: ${{ matrix.os == 'windows-11-arm' && 'ARM64' || '' }}
+      SKIP_CORE_BUILD: "true"
```
sfc-gh-turbaszek, line 139: Given that, we need to add wheel building test to pakaging tests. Previously it was always tested when building the lib.

sfc-gh-turbaszek (changes requested):
Please make sure the E2E wheel building path is covered in python packaging tests.

sfc-gh-pbulawa:
> For any build: Why do we even append the hash at all, when restore keys never have a key

​The `hashFiles('**/Cargo.lock')` in the primary `key` enables **exact cache hits**. Without it, every run would be a partial match via `restore-keys`, and the post step would always try to save a new cache. With the hash, once the cache is saved for a given `Cargo.lock`, subsequent runs get an exact hit and skip the save entirely. The `restore-keys` fallback only kicks in when `Cargo.lock` changes, giving a warm-ish starting point instead of a cold start.

.github/workflows/test-python.yml
sfc-gh-pbulawa: `Cargo.lock` pins all transitive dependency versions for the workspace. When any dependency of `sf_core` changes, the lock file changes, invalidating the cache and triggering a fresh build. If only `sf_core` source code changes (but not dependencies), the cache key stays the same and cargo handles incremental recompilation of just the changed crate.

.github/workflows/test-python.yml
sfc-gh-pbulawa: Added

.github/workflows/test-odbc.yml
sfc-gh-pbulawa: You're right. I've removed it for now. If it will be too slow, we can use ccache.

sfc-gh-turbaszek (approved):
Thanks for adding the test!

sfc-gh-jszczerbinski (approved):

PR #665, opened 2026-03-20
Change ODBC test authentication to keypair

sfc-gh-pczajka:
Change ODBC test suite authentication method to keypair (aligned with other wrappers) instead of user + password

### DIFF
--- a/odbc_tests/common/include/test_setup.hpp
+++ b/odbc_tests/common/include/test_setup.hpp
@@ -9,6 +9,14 @@
 #include <set>
 #include <sstream>
 
+#ifdef _WIN32
+#include <process.h>
+inline int current_pid() { return _getpid(); }
+#else
+#include <unistd.h>
+inline int current_pid() { return getpid(); }
+#endif
+
 #include <catch2/catch_test_macros.hpp>
 
 #include "ODBCConfig.hpp"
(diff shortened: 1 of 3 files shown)


sfc-gh-pbulawa (approved):

odbc_tests/common/include/test_setup.hpp
```diff
@@ -98,38 +133,35 @@ inline void configure_driver_string(std::stringstream& ss) {
...
+  add_param_optional<std::string>(ss, params, "SNOWFLAKE_TEST_PROTOCOL", "PROTOCOL");
 }
 
 inline std::string get_connection_string() {
   auto params = get_test_parameters("testconnection");
   std::stringstream ss;
   read_default_params(ss, params);
-  add_param_required<std::string>(ss, params, "SNOWFLAKE_TEST_PASSWORD", "PWD");
+  ss << "AUTHENTICATOR=SNOWFLAKE_JWT;";
+#ifdef SNOWFLAKE_OLD_DRIVER
```
sfc-gh-jszczerbinski: Why ifdef? Can't we do the same for both?

sfc-gh-jszczerbinski (commented):
One commment? Otherwise LGTM?

odbc_tests/common/include/test_setup.hpp
sfc-gh-pczajka: There is a bug in old ODBC, @sfc-gh-pbulawa will create a bug report for it

PR #669, opened 2026-03-23
SNOW-3258878: SQLCancel scaffolding

sfc-gh-ddas:
### TL;DR

Implements SQLCancel functionality for the ODBC driver with proper state management, cross-thread cancellation support, and comprehensive test coverage.

### What changed?

- **Added SQLCancel API**: Exported `SQLCancel` function in `exports.def` and implemented the C API entry point in `c_api.rs`
- **Enhanced statement state management**: Split the previous `Executed` and `NoResultSet` states into more specific states (`QueryExecuted`, `DdlExecuted`, `DmlExecuted`) to better track statement types and cursor status
- **Implemented cancellation infrastructure**: Added `CancellationToken` to `Statement` struct and wired it through execution paths for future async cancellation support
- **Added cursor state validation**: SQLExecDirect and SQLExecute now return error 24000 (Invalid cursor state) when called with an open cursor, requiring explicit cursor closure
- **Enhanced error handling**: Added new error types `CursorAlreadyOpen`, `DataNotFetched`, and `OperationCanceled` with appropriate SQL state mappings
- **Improved DML handling**: Zero-row DML operations now return `SQL_NO_DATA` instead of opening a cursor
- **Added comprehensive test suite**: 30+ test cases covering basic cancellation, cross-thread scenarios, state preservation, and edge cases

### How to test?

Run the new test suite in `odbc_tests/tests/odbc-api/terminating_statement/cancel_tests.cpp` which covers:
- Basic cancellation scenarios (idle statements, after execution, after fetch)
- Cross-thread cancellation with long-running queries
- State preservation (bindings, attributes, cursor position)
- Error conditions and edge cases
- Platform-specific behavior differences between Windows and Unix ODBC Driver Managers

### Why make this change?

- **ODBC compliance**: SQLCancel is a required ODBC API function that was missing from the driver
- **Application compatibility**: Many ODBC applications expect SQLCancel to be available for query interruption and cursor management
- **Better state management**: The refined state machine provides clearer semantics for different statement types (DDL vs DML vs queries)
- **Foundation for async support**: The cancellation token infrastructure prepares for future asynchronous execution and server-side query cancellation
- **Improved error handling**: Proper cursor state validation prevents application bugs and aligns with ODBC 3.x specifications

### DIFF
--- a/odbc/src/api/statement.rs
+++ b/odbc/src/api/statement.rs
@@ -1,8 +1,8 @@
 use crate::api::CDataType;
 use crate::api::encoding::OdbcEncoding;
 use crate::api::error::{
-    ArrowArrayStreamReaderCreationSnafu, DisconnectedSnafu, InvalidBufferLengthSnafu,
-    InvalidCursorStateSnafu, InvalidHandleSnafu, InvalidParameterNumberSnafu,
+    ArrowArrayStreamReaderCreationSnafu, CursorAlreadyOpenSnafu, DisconnectedSnafu,
+    InvalidBufferLengthSnafu, InvalidHandleSnafu, InvalidParameterNumberSnafu,
     InvalidPrecisionOrScaleSnafu, JsonBindingSnafu, NoMoreDataSnafu, NullPointerSnafu,
     OdbcRuntimeSnafu, ReadOnlyAttributeSnafu, Required, StatementNotExecutedSnafu,
 };
@@ -22,6 +22,7 @@ use sf_core::protobuf::generated::database_driver_v1::{
     StatementPrepareRequest, StatementSetSqlQueryRequest, query_bindings,
 };
 use snafu::ResultExt;
+use tokio_util::sync::CancellationToken;
 use tracing;
 
 /// Execute a SQL statement directly (SQLExecDirect / SQLExecDirectW).
@@ -40,12 +41,12 @@ fn exec_direct_impl(statement_handle: sql::Handle, statement_text: &str) -> Odbc
 
     if matches!(
         stmt.state.as_ref(),
-        StatementState::Executed { .. }
+        StatementState::QueryExecuted { .. }
             | StatementState::Fetching { .. }
             | StatementState::Done { .. }
     ) {
         tracing::error!("exec_direct: cursor is already open");
-        return InvalidCursorStateSnafu.fail();
+        return CursorAlreadyOpenSnafu.fail();
     }
 
     match &mut stmt.conn.state {
@@ -56,6 +57,10 @@ fn exec_direct_impl(statement_handle: sql::Handle, statement_text: &str) -> Odbc
             let (bindings, _json_owner) = apply_parameter_bindings(&stmt.apd, &stmt.ipd, false)?;
             let stmt_handle = stmt.stmt_handle;
 
+            stmt.cancel_token = CancellationToken::new();
+            let _cancel_token = stmt.cancel_token.clone();
+            // TODO(SNOW-3258922): Wrap RPC in tokio::select! with
+            // _cancel_token.cancelled() to support cross-thread SQLCancel.
             let response = global().context(OdbcRuntimeSnafu)?.block_on(async |c| {
                 c.statement_set_sql_query(StatementSetSqlQueryRequest {
                     stmt_handle: Some(stmt_handle),
@@ -75,8 +80,19 @@ fn exec_direct_impl(statement_handle: sql::Handle, statement_text: &str) -> Odbc
 
             let query_id = response.result.as_ref().map(|r| r.query_id.clone());
             update_numeric_settings(conn_handle, &mut stmt.conn.numeric_settings)?;
-            set_state(stmt, create_execute_state(response, false)?);
+            let execute_state = create_execute_state(response, false)?;
+            let is_zero_dml = matches!(
+                &execute_state,
+                StatementState::DmlExecuted {
+                    rows_affected: 0,
+                    ..
+                }
+            );
+            set_state(stmt, execute_state);
             stmt.last_query_id = query_id.filter(|s| !s.is_empty());
+            if is_zero_dml {
+                return NoMoreDataSnafu.fail();
+            }
             Ok(())
         }
         ConnectionState::Disconnected => {
@@ -136,8 +152,6 @@ fn update_numeric_settings(
     Ok(())
 }
 
-/// Read the query text from an ODBC buffer and prepare
-/// (SQLPrepare / SQLPrepareW).
 /// Prepare a SQL statement (SQLPrepare / SQLPrepareW).
 pub fn prepare<E: OdbcEncoding>(
     statement_handle: sql::Handle,
@@ -168,12 +182,12 @@ fn prepare_impl(statement_handle: sql::Handle, query: &str) -> OdbcResult<()> {
 
     if matches!(
         stmt.state.as_ref(),
-        StatementState::Executed { .. }
+        StatementState::QueryExecuted { .. }
             | StatementState::Fetching { .. }
             | StatementState::Done { .. }
     ) {
         tracing::error!("prepare: cursor is already open");
-        return InvalidCursorStateSnafu.fail();
+        return CursorAlreadyOpenSnafu.fail();
     }
 
     match &mut stmt.conn.state {
@@ -184,6 +198,10 @@ fn prepare_impl(statement_handle: sql::Handle, query: &str) -> OdbcResult<()> {
             tracing::debug!("prepare: query = {query}");
 
             let stmt_handle = stmt.stmt_handle;
+            stmt.cancel_token = CancellationToken::new();
+            let _cancel_token = stmt.cancel_token.clone();
+            // TODO(SNOW-3258922): Wire _cancel_token into tokio::select!
+            // alongside the RPC future to support cancellation.
             let prepare_result = global().context(OdbcRuntimeSnafu)?.block_on(async |c| {
                 c.statement_set_sql_query(StatementSetSqlQueryRequest {
                     stmt_handle: Some(stmt_handle),
@@ -234,17 +252,18 @@ pub fn execute(statement_handle: sql::Handle) -> OdbcResult<()> {
 
     if matches!(
         stmt.state.as_ref(),
-        StatementState::Executed { .. }
+        StatementState::QueryExecuted { .. }
             | StatementState::Fetching { .. }
             | StatementState::Done { .. }
     ) {
         tracing::error!("execute: cursor is already open");
-        return InvalidCursorStateSnafu.fail();
+        return CursorAlreadyOpenSnafu.fail();
     }
 
     let prepared = match stmt.state.as_ref() {
         StatementState::Prepared { .. } => true,
-        StatementState::NoResultSet { prepared, .. } => *prepared,
+        StatementState::DdlExecuted { prepared, .. }
+        | StatementState::DmlExecuted { prepared, .. } => *prepared,
         _ => false,
     };
 
@@ -255,6 +274,10 @@ pub fn execute(statement_handle: sql::Handle) -> OdbcResult<()> {
         } => {
             let (bindings, _json_owner) = apply_parameter_bindings(&stmt.apd, &stmt.ipd, prepared)?;
 
+            stmt.cancel_token = CancellationToken::new();
+            let _cancel_token = stmt.cancel_token.clone();
+            // TODO(SNOW-3258922): Wrap RPC in tokio::select! with
+            // _cancel_token.cancelled() to support cross-thread SQLCancel.
             let response = global().context(OdbcRuntimeSnafu)?.block_on(async |c| {
                 c.statement_execute_query(StatementExecuteQueryRequest {
                     stmt_handle: Some(stmt.stmt_handle),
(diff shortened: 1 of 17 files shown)


sfc-gh-ddas:
* **#752** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/752?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#769](https://github.com/snowflakedb/universal-driver/pull/769) <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/769?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#857](https://github.com/snowflakedb/universal-driver/pull/857) <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/857?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#751** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/751?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#669** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/669?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/669?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

odbc_tests/tests/odbc-api/terminating_statement/cancel_tests.cpp
```diff
@@ -336,8 +357,350 @@ TEST_CASE_METHOD(StmtDefaultDSNFixture, "SQLCancel: Diagnostics after cancel err
...
+
+  SQLRETURN cancel_ret = SQLCancel(stmt_handle());
+  REQUIRE(cancel_ret == SQL_SUCCESS);
+
+  int polls = 0;
+  SQLRETURN poll_ret = SQL_STILL_EXECUTING;
+  while (poll_ret == SQL_STILL_EXECUTING && ++polls < kMaxPollIterations) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    poll_ret = SQLExecDirect(stmt_handle(), sqlchar("SELECT COUNT(*) FROM TABLE(GENERATOR(TIMELIMIT => 30))"), SQL_NTS);
+  }
```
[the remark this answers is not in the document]
sfc-gh-ddas: This is necessary to ensure the query is in-flight when cancel fires.

odbc_tests/tests/odbc-api/terminating_statement/cancel_tests.cpp
sfc-gh-pbulawa, line 1: Please add tests:

- SQLCancel after error recovery and re-execution. 
- SQLCancel on an allocated-but-never-used statement, then free. Tests cover cancel-on-idle, but don't verify that SQLFreeHandle(SQL_HANDLE_STMT) succeeds after cancel on a fresh statement. This confirms cancel doesn't corrupt the handle state.
- Parameter binding preservation after cancel with open cursor. The existing "Preserves bound parameters" test cancels on a prepared-but-not-executed statement. A variant should cancel after execution with an open cursor, close the cursor, re-execute with the same bindings, and verify values.

sfc-gh-pbulawa (changes requested):

odbc_tests/tests/odbc-api/terminating_statement/cancel_tests.cpp
sfc-gh-ddas: Added tests for these.

odbc_tests/BehaviorDifferences.yaml
```diff
@@ -312,3 +311,28 @@ behavior_differences:
...
+    description: |
+      OLD driver: SQLCancel unconditionally closes/invalidates the cursor,
+      causing subsequent SQLFetch to return HY010 and SQLExecDirect to return
+      24000. This is a known deviation from the ODBC 3.5 specification.
+      NEW driver: SQLCancel on a synchronous (already-completed) operation is
+      a no-op per the ODBC 3.5 spec. The cursor remains open and usable.
+      Spec reference: https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlcancel-function
+      Impact: Applications that rely on SQLCancel closing cursors will need
+      to use SQLFreeStmt(SQL_CLOSE) or SQLCloseCursor instead.
+  31:
```
sfc-gh-jszczerbinski, line 324: Are you sure about this behavior difference? I think you are applying the documentation incorrectly here. It says we need call close cursor before reexecuting, but in the tests you are applying this BD where there is no cursor open (for example INSERT returns no result set -> therefore doesn't open cursor)

odbc_tests/BehaviorDifferences.yaml
```diff
@@ -312,3 +311,28 @@ behavior_differences:
       field incorrect), caught locally by the driver.
       NEW driver: Sends the query to the server, which returns SQL_ERROR with
       SQLSTATE 42000 (Syntax error: Bind variable ? not set).
+  30:
```
sfc-gh-jszczerbinski, line 315: Looks like DM behavior difference between Linux and Windows. Good catch

odbc_tests/tests/odbc-api/terminating_statement/cancel_tests.cpp
```diff
@@ -127,30 +133,38 @@ TEST_CASE_METHOD(StmtDefaultDSNFixture, "SQLCancel: After cancel on executed pre
...
-
   SQLRETURN ret = SQLExecDirect(stmt_handle(), sqlchar("SELECT 1"), SQL_NTS);
   REQUIRE(ret == SQL_SUCCESS);
 
   ret = SQLCancel(stmt_handle());
   REQUIRE(ret == SQL_SUCCESS);
 
+  // Both drivers return 24000 on re-execution after cancel, but for
+  // different reasons:
+  // - Old driver (BD#30): cancel closes the cursor on Unix, but leaves the
```
sfc-gh-jszczerbinski: Can we disambiguate here somehow in code?

odbc_tests/tests/odbc-api/terminating_statement/cancel_tests.cpp
```diff
@@ -164,37 +178,99 @@ TEST_CASE_METHOD(StmtDefaultDSNFixture, "SQLCancel: Statement recoverable via SQ
...
                  "[odbc-api][cancel][terminating_statement]") {
-  SKIP_NEW_DRIVER_NOT_IMPLEMENTED();
-
   SQLRETURN ret = SQLExecDirect(stmt_handle(), sqlchar("SELECT 1"), SQL_NTS);
   REQUIRE(ret == SQL_SUCCESS);
 
   ret = SQLCancel(stmt_handle());
   REQUIRE(ret == SQL_SUCCESS);
 
   WINDOWS_ONLY {
```
sfc-gh-jszczerbinski, line 210: Nit: in general could swap PLATFORM_ONLY statements be inside DRIVER_ONLY? So we can highlight that new driver behave the same way on both oses and old driver doesn't? Similiar for other tests like that

odbc_tests/tests/odbc-api/terminating_statement/cancel_tests.cpp
```diff
@@ -307,27 +423,36 @@ TEST_CASE_METHOD(StmtDefaultDSNFixture, "SQLCancel: Cancels data-at-execution op
...
-// ============================================================================
-
-TEST_CASE_METHOD(StmtDefaultDSNFixture, "SQLCancel: Diagnostics after cancel error state",
+TEST_CASE_METHOD(StmtDefaultDSNFixture, "SQLCancel: Re-execute immediately after canceling data-at-execution",
                  "[odbc-api][cancel][terminating_statement]") {
   SKIP_NEW_DRIVER_NOT_IMPLEMENTED();
 
-  SQLRETURN ret = SQLExecDirect(stmt_handle(), sqlchar("SELECT 1"), SQL_NTS);
+  SQLRETURN ret = SQLPrepare(stmt_handle(), sqlchar("SELECT ?"), SQL_NTS);
   REQUIRE(ret == SQL_SUCCESS);
```
sfc-gh-jszczerbinski: You might want to migrate to matchers in those tests: REQUIRE_THAT(OdbcResult(stmt, ret), IsSuccess) gives much better debug info

odbc_tests/tests/odbc-api/terminating_statement/cancel_tests.cpp
```diff
@@ -336,8 +461,350 @@ TEST_CASE_METHOD(StmtDefaultDSNFixture, "SQLCancel: Diagnostics after cancel err
...
+}
+
+// ============================================================================
+// SQLCancel - Async Cancel
+// ============================================================================
+
+TEST_CASE_METHOD(StmtDefaultDSNFixture, "SQLCancel: Async cancel interrupts execution with HY008",
+                 "[odbc-api][cancel][terminating_statement][async]") {
+  SKIP_NEW_DRIVER_NOT_IMPLEMENTED();
+  SKIP_OLD_DRIVER("BD#30", "Reference driver async cancel does not interrupt in-progress operations");
```
sfc-gh-jszczerbinski: Is this the same behavior difference that is descibed in YAML?

odbc_tests/tests/odbc-api/terminating_statement/cancel_tests.cpp
```diff
@@ -336,8 +461,350 @@ TEST_CASE_METHOD(StmtDefaultDSNFixture, "SQLCancel: Diagnostics after cancel err
...
+  // The HY008 from the canceled SQLExecDirect should be the only record.
+  auto records = get_diag_rec(SQL_HANDLE_STMT, stmt);
+  REQUIRE(records.size() == 1);
+  REQUIRE(records[0].sqlState == "HY008");
+}
+
+TEST_CASE_METHOD(StmtDefaultDSNFixture, "SQLCancel: HY018 when server declines cancel request",
+                 "[odbc-api][cancel][terminating_statement][cross_thread]") {
+  SKIP_NEW_DRIVER_NOT_IMPLEMENTED();
+  SKIP_OLD_DRIVER("BD#30", "Snowflake server does not decline cancel requests");
```
sfc-gh-jszczerbinski: Is this the same behavior difference?

sfc-gh-jszczerbinski (changes requested):
In general looks good, two main points:  
\- I disagree with some of the behavior differences  
\- Let's discuss the overall approach for cancellation, I don't think this extra atomic field is the right choice here, but we can tackle that in the next PR

odbc_tests/BehaviorDifferences.yaml
sfc-gh-ddas: Added new state for `Statement` and renamed existing states to make it more clear.

odbc_tests/tests/odbc-api/terminating_statement/cancel_tests.cpp
sfc-gh-ddas: This is a different BD, Old driver doesn't seem to do the Async cancel correctly.

odbc/src/api/statement.rs
```diff
@@ -22,8 +22,60 @@ use sf_core::protobuf::generated::database_driver_v1::{
...
+use tokio_util::sync::CancellationToken;
 use tracing;
 
+// ---------------------------------------------------------------------------
+// Cancel registry: stores CancellationTokens keyed by statement handle address.
+// SQLCancel looks up the token here without dereferencing the Statement,
+// eliminating aliasing UB when called from another thread.
+// ---------------------------------------------------------------------------
+
+static CANCEL_REGISTRY: LazyLock<Mutex<HashMap<usize, CancellationToken>>> =
```
sfc-gh-jszczerbinski: We shouldn't rely on global state in ODBC, as Driver manager can load and unload the binary and statics do not get dropped in Rust (see how we handle [runtime.rs](http://runtime.rs))

Also we have a global mutex here that could receive a lot of contention, instead can we store CancelationToken as RwLock\<Arc\<CancellationToken>> in stmt/conn/env structs?

SQLCancel would take current cancellation token, call cancel on it and swap it with a new token for subsequent operations to use?

Other operations would read the cancelation token at the beginning of the operation and await cancels where needed.

sfc-gh-jszczerbinski (changes requested):
Please avoid globals in ODBC

odbc/src/api/statement.rs
```diff
@@ -38,6 +90,18 @@ fn exec_direct_impl(statement_handle: sql::Handle, statement_text: &str) -> Odbc
...
+        stmt.state.as_ref(),
+        StatementState::QueryExecuted { .. }
+            | StatementState::Fetching { .. }
+            | StatementState::Done { .. }
+    ) {
+        tracing::error!("exec_direct: cursor is already open");
+        return CursorAlreadyOpenSnafu.fail();
+    }
+
+    let _cancel_token = reset_cancel_token(statement_handle as usize);
```
sfc-gh-pbulawa: If the connection is `Disconnected`, a new token is allocated in the registry (replacing the existing one) and then immediately discarded when the function returns an error. Let's move it inside the `Connected` branch.

odbc/src/api/statement.rs
```diff
@@ -197,17 +275,20 @@ pub fn execute(statement_handle: sql::Handle) -> OdbcResult<()> {
...
+        StatementState::QueryExecuted { .. }
             | StatementState::Fetching { .. }
             | StatementState::Done { .. }
     ) {
         tracing::error!("execute: cursor is already open");
-        return InvalidCursorStateSnafu.fail();
+        return CursorAlreadyOpenSnafu.fail();
     }
 
+    let _cancel_token = reset_cancel_token(statement_handle as usize);
```
sfc-gh-pbulawa: If the connection is `Disconnected`, a new token is allocated in the registry (replacing the existing one) and then immediately discarded when the function returns an error. Let's move it inside the `Connected` branch.

sfc-gh-pbulawa (changes requested):

odbc/src/api/statement.rs
sfc-gh-ddas: When `SQLExecDirect` runs, it dereferences the statement handle to get an exclusive mutable reference to the entire struct. If `SQLCancel` is called from another thread while `SQLExecDirect` is in-flight, and `cancel()` also accesses a field on `Statement` we'd have two concurrent references to the same struct. I see the downsides of the global state, however the only way we can still have the cancel token on `Statement` is by using a pointer to it and using `addr_of!` to get to the cancel token and create a `&RwLock<>` scoped to just that field. This is very hacky imo, what do you think?

sfc-gh-jszczerbinski: Statements not being thread safe is separate issue and we should deal with it in time, it shouldn't stop us from having a single field that is thread safe. This solution is not acceptable to me

sfc-gh-jszczerbinski (changes requested):

odbc/src/api/statement.rs
sfc-gh-jszczerbinski: > however the only way we can still have the cancel token on `Statement` is by using a pointer to it and using `addr_of!`

Why is that?

sfc-gh-ddas: Removed the global registry and moved the `CancellationToken` in the `Statement` struct. `cancel()` now accesses it via `stmt_from_handle()`. The pre-existing aliasing issue from `stmt_from_handle` creating `&mut Statement` concurrently will be addressed by the future handle manager

sfc-gh-jszczerbinski (approved):

sfc-gh-pbulawa (approved):

PR #687, opened 2026-03-23
NO-SNOW: Improve JDBC logging bridge

sfc-gh-rkowalski:
Make jdbc_bridge use SFLogger instead of SLF4J directly

### DIFF
--- a/jdbc_bridge/src/sflogger_layer.rs
+++ b/jdbc_bridge/src/sflogger_layer.rs
@@ -1,28 +1,24 @@
 use jni::JavaVM;
+use jni::objects::JValue;
 use std::fmt::Debug;
 use tracing::{Event, Level, Subscriber, field::Field};
 use tracing_subscriber::Layer;
 use tracing_subscriber::layer::Context;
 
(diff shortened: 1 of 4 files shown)


sfc-gh-merbel (approved):

jdbc_bridge/src/sflogger_layer.rs
```diff
@@ -79,22 +73,15 @@ where
...
+            level_str,
+            "(Ljava/lang/String;Z)V",
+            &[(&java_log_msg).into(), JValue::Bool(1)],
         )
         .unwrap();
     }
 }
 
-/*
-TODO: Not sure if this will work,
```
sfc-gh-pfus, line 94: This TODO is not relevant anymore?

sfc-gh-pfus (approved):

jdbc_bridge/src/sflogger_layer.rs
sfc-gh-rkowalski: I think so, we did not encounter any errors related to this jdbc_bridge implementation so I'm removing it

PR #708, opened 2026-03-24
Add Python 3.12, protobuf, and uv to shell.nix dev environment

sfc-gh-turbaszek:
Add protobuf_33, python312, and uv to nativeBuildInputs. Add shellHook to automatically create and activate a Python venv in python/.venv.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -5,8 +5,8 @@ pkgs.mkShell {
     pkgs.pkg-config
     pkgs.cmake
     pkgs.rustup
-    pkgs.protobuf_33
-    pkgs.python312
+    pkgs.protobuf_32
+    pkgs.python313
     pkgs.uv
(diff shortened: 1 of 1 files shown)


shell.nix
```diff
@@ -5,11 +5,21 @@ pkgs.mkShell {
     pkgs.pkg-config
     pkgs.cmake
     pkgs.rustup
+    pkgs.protobuf_33
+    pkgs.python312
```
sfc-gh-fpawlowski: nit: Why not the newest version we use?

sfc-gh-fpawlowski (approved):
LGTM

shell.nix
sfc-gh-turbaszek: No strong opinions, I just picked up 3.12

PR #713, opened 2026-03-24
Quick and dirty SQLConnect implementation to enable EA

sfc-gh-jszczerbinski:


odbc/src/api/connection.rs
sfc-gh-rkowalski, line 141: `PARAM_MAPPINGS` uses uppercase keys ("ACCOUNT", "SERVER", etc.) and the lookup in `connect_with_params` does an exact match but, `parse_ini_section` preserves the original casing from the INI file. But odbc.ini files can use mixed case (e.g., Server, Account, Uid).

sfc-gh-rkowalski: How about we normalize keys to uppercase in `parse_ini_section`?

sfc-gh-rkowalski: `driver_connect` -> `connect_with_params`

sfc-gh-rkowalski (approved):

PR #716, opened 2026-03-24
odbc_tests: run 128 MB string LOB case before 16 MB

sfc-gh-boler:
The 128 MB ODBC `string_lob` scenario passes reliably when run on its
own, but it has been load-sensitive in the full parallel CTest run.
Because Catch2 discovery exposes each `TEST_CASE` as its own CTest
test, reordering the definitions changes which case CTest starts
first. This change keeps the coverage and assertions the same, but
moves the 128 MB case ahead of the 16 MB case so the heaviest
scenario starts earlier and the full suite stays stable in practice.

This is split out as a small test-only parent change so the config
lifecycle work can land independently of the ODBC suite
stabilization.

### DIFF
--- a/odbc_tests/tests/e2e/types/string_lob.cpp
+++ b/odbc_tests/tests/e2e/types/string_lob.cpp
@@ -41,62 +41,6 @@ static std::string generate_random_ascii_string(std::mt19937& gen, size_t length
 // LOB (LARGE OBJECT) STRING HANDLING
 // ============================================================================
 
-TEST_CASE("should handle LOB string at historical 16 MB limit", "[datatype][string][lob]") {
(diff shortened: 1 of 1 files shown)


sfc-gh-jszczerbinski (approved):
Looks good, but are you sure that we are running in the same process? The catch ctests integration runs each test case in a separate process AFAIK

sfc-gh-boler:
> Looks good, but are you sure that we are running in the same process? The catch ctests integration runs each test case in a separate process AFAIK

You are right, it does run in a separate process. I suspect the gain will be minimal then. I'll submit this and keep digging if it's still flaky.

PR #717, opened 2026-03-25
NO-SNOW: Dynamically build ClientInfo from real hardware and wrapper identity

sfc-gh-turbaszek:
Replace hardcoded Python-specific values in ClientInfo with real OS detection in sf_core and wrapper-provided identity via settings. Each wrapper (Python, ODBC, JDBC) now passes its own client_app_id, client_app_version, and runtime metadata. OS and OS version are auto-detected using std::env::consts::OS and libc::uname.

Also fixes duplicate User-Agent construction in snowflake_login that bypassed the user_agent() function.


sf_core/src/rest/snowflake/auth.rs
```diff
@@ -28,11 +32,11 @@ pub struct AuthRequestClientEnvironment {
     #[serde(rename = "OCSP_MODE", skip_serializing_if = "Option::is_none")]
     pub ocsp_mode: Option<String>,
     #[serde(rename = "PYTHON_VERSION", skip_serializing_if = "Option::is_none")]
-    pub python_version: Option<String>,
+    pub runtime_version: Option<String>,
```
sfc-gh-pfus, line 38: Should we change serde as well?

sfc-gh-turbaszek: Done — renamed serde keys from `PYTHON_VERSION`/`PYTHON_RUNTIME`/`PYTHON_COMPILER` to `RUNTIME_VERSION`/`RUNTIME_NAME`/`COMPILER` in 5a3e0b18. Also updated all test assertions to match.

sfc-gh-rsavenok (approved):

python/src/snowflake/connector/connection.py
```diff
@@ -67,7 +67,6 @@
 # uncast.
 DEFAULT_CONFIGURATION: dict[str, tuple[Any, tuple[type, ...]]] = {}
 
-CLIENT_NAME = "snowflake-connector-python"
```
sfc-gh-asolarski, line 71: wasn't this constant added intentionally to allow importing?

sfc-gh-asolarski: I mean - we should identify as `PythonConnector` indeed, but I'm wondering about this public constant.

sf_core/src/bin/collect_chunk_data.rs
```diff
@@ -117,11 +117,14 @@ fn load_parameters(path: &PathBuf) -> Result<Parameters, Box<dyn std::error::Err
 
 fn default_client_info() -> ClientInfo {
     ClientInfo {
-        application: "PythonConnector".to_string(),
-        version: "3.15.0".to_string(),
-        os: "Darwin".to_string(),
-        os_version: "macOS-15.5-arm64-arm-64bit".to_string(),
+        application: "drivers_core".to_string(),
+        version: "0.0.1dev".to_string(),
```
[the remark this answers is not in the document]
sfc-gh-asolarski: this is valid concern, isn't it?

sf_core/src/config/rest_parameters.rs
```diff
@@ -87,16 +93,27 @@ impl ClientInfo {
...
-        // the wrapper client type
         let client_info = ClientInfo {
             application: settings
                 .get_string("client_app_id")
-                .unwrap_or_else(|| "PythonConnector".to_string()),
-            version: "3.15.0".to_string(),
-            os: "Darwin".to_string(),
-            os_version: "macOS-15.5-arm64-arm-64bit".to_string(),
+                .filter(|s| !s.trim().is_empty())
+                .unwrap_or_else(|| "drivers_core".to_string()),
```
sfc-gh-asolarski: we have those constants defined here and in collect_chunk_data.rs - could we extract them (in case they are really needed as constant)?

sf_core/src/config/rest_parameters.rs
```diff
@@ -87,16 +93,27 @@ impl ClientInfo {
...
+                .unwrap_or_else(|| "0.0.1dev".to_string()),
+            os: std::env::consts::OS.to_string(),
+            os_version: crate::telemetry::environment::detect_os_version(),
             ocsp_mode: Some("FAIL_OPEN".to_string()),
+            runtime_name: settings
+                .get_string("client_runtime_name")
+                .and_then(|s| if s.trim().is_empty() { None } else { Some(s) }),
+            runtime_version: settings
+                .get_string("client_runtime_version")
+                .and_then(|s| if s.trim().is_empty() { None } else { Some(s) }),
```
sfc-gh-asolarski, line 113: `.trim().is_empty()`
we could have a helper for that, it's everywhere 😅

odbc/src/api/connection.rs
```diff
@@ -268,31 +268,35 @@ fn connect_with_params(
...
         apply_pre_connection_runtime_attrs_async(c, &pre_connection_attrs, conn_handle).await?;
 
         c.connection_init(ConnectionInitRequest {
             conn_handle: Some(conn_handle),
             db_handle: Some(db_handle),
-            ..Default::default()
+            wrapper_identity: Some(WrapperIdentity {
+                driver_name: Some("ODBC".to_string()),
+                driver_version: Some(env!("CARGO_PKG_VERSION").to_string()),
+                language_runtime: None,
```
sfc-gh-jszczerbinski, line 299: Can we extract those too? Should be available in compile time

sfc-gh-jszczerbinski (approved):

python/src/snowflake/connector/connection.py
sfc-gh-turbaszek: Good point, this constant was not defined in `connection.py` but in `description.py`. Let's renae it to `PythonConnector` and keep for back compat?

sf_core/src/config/rest_parameters.rs
sfc-gh-turbaszek: Yup, but let's leave it for further PRs

sf_core/src/bin/collect_chunk_data.rs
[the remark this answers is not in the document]
sfc-gh-turbaszek: fixed

odbc/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "odbc"
-version = "0.1.0"
+version = "3.0.0"
```
sfc-gh-pbulawa: It should be `4.0.0`

PR #743, opened 2026-03-27
connection background task for heartbeat

sfc-gh-mhofman:

PR #742, opened 2026-03-27
heartbeat sending service

sfc-gh-mhofman:
Add heatbeat sending service to refresh idle connections

PR #745, opened 2026-03-27
SNOW-3480688 integration tests for heartbeat

sfc-gh-mhofman:


sfc-gh-mhofman:
* **#745** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/745?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/745?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#744** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/744?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#743** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/743?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#742** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/742?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sf_core/tests/integration/session/heartbeat.rs
```diff
@@ -0,0 +1,483 @@
...
+async fn heartbeat_uses_short_interval_for_testing() {
+    // Given a master token validity of 16 seconds
+    let interval = compute_heartbeat_interval(Some(Duration::from_secs(16)));
+
+    // Then the computed interval (16s / 4 = 4s) should be clamped to the minimum of 900s
+    assert_eq!(
+        interval,
+        Duration::from_secs(900),
+        "16s validity / 4 = 4s, but clamped to min 900s"
+    );
```
sfc-gh-makowalski: This test is currently the one red dot on the `test (...)` jobs:

```
test session::heartbeat::heartbeat_uses_short_interval_for_testing ... FAILED

thread '…' panicked at sf_core/tests/integration/session/heartbeat.rs:243:5:
assertion `left == right` failed: 16s validity / 4 = 4s, but clamped to min 900s
```

`compute_heartbeat_interval` has no minimum clamp today — `16s / 4 = 4s`, and `.min(MAX_HEARTBEAT_INTERVAL)` only caps at 3600s. The unit test `compute_interval_no_bottom_clamp` in `heartbeat.rs` actually asserts there's *no* bottom clamp.

Two ways out:

1. Drop this assertion (the rest of the test — the wiremock loop — is still useful; the test name/comment would need reconciling).
2. Introduce a `MIN_HEARTBEAT_INTERVAL` constant and clamp it in `compute_heartbeat_interval`, then update the existing `compute_interval_no_bottom_clamp` unit test. Heads-up though: `heartbeat_started_when_keep_alive_true` / `connection_close_cancels_heartbeat` rely on no lower clamp (they use `masterValidityInSeconds: 1` to get a 250 ms interval), so a 900 s minimum would break them too — pick a smaller minimum or special-case the test setup.

sf_core/src/apis/database_driver_v1/mod.rs
```diff
@@ -5,7 +5,7 @@ pub mod connection;
 mod database;
 pub(crate) mod error;
 mod global_state;
-pub(crate) mod heartbeat;
+pub mod heartbeat;
```
sfc-gh-makowalski, line 12: Was widening this from `pub(crate)` to `pub` deliberate? Same question for `HeartbeatHandle`, `spawn_heartbeat_task`, and `compute_heartbeat_interval` in `heartbeat.rs`.

It looks like the only reason is so the new integration tests under `tests/integration/session/heartbeat.rs` can reach those items. The codebase already has a convention for that case — the `test-utils` Cargo feature:

```toml
[features]
test-utils = []

[dev-dependencies]
sf_core = { path = ".", default-features = false, features = ["test-utils"] }
```

and modules expose test-only internals via `#[cfg(any(test, feature = "test-utils"))] pub mod test_fixtures { ... }` (see `config/rest_parameters.rs`, `telemetry/platform_detection/mod.rs`, `fs_adapter.rs`).

Gating this module (or just a small `pub mod test_fixtures { pub use … }` re-export) behind `#[cfg(any(test, feature = "test-utils"))]` would keep the runtime public surface of `sf_core` unchanged. Happy to be overruled if you went `pub` on purpose.

sf_core/tests/integration/session/heartbeat.rs
```diff
@@ -0,0 +1,483 @@
...
+        ocsp_mode: None,
+        crl_config: CrlConfig::default(),
+        tls_config: TlsConfig::default(),
+        platforms: vec![],
+        os_details: None,
+        compiler: None,
+        runtime_name: None,
+        runtime_version: None,
+    }
+}
```
sfc-gh-makowalski: Were you aware of `sf_core::config::rest_parameters::test_fixtures::test_client_info`? It's the shared `ClientInfo` fixture (gated behind the `test-utils` feature) used by every other integration test in this repo — `telemetry/rest_endpoint.rs`, `telemetry/exporter_pipeline.rs`, `session/session_refresh.rs`, `http/sync_retry.rs`, `http/session_refresh.rs`, `e2e/session/session_refresh.rs` — *and* by the heartbeat unit tests inside `apis/database_driver_v1/heartbeat.rs` itself.

The two differ slightly (shared uses `TlsConfig::insecure()` and `std::env::consts::OS`; local uses `TlsConfig::default()` and `"test-os"`), but functionally equivalent for the wiremock HTTP tests here. Reusing it would also let you drop the `ClientInfo` / `CrlConfig` / `TlsConfig` / `SensitiveString` imports at the top of this file.

For `test_tokens` there's no shared fixture — multiple integration tests redefine their own — so that part is fine.

sfc-gh-makowalski (commented):
A few inline comments — mostly questions about whether some choices were deliberate. The only one I'd call a real blocker is the failing test on CI; the rest are nits or stylistic / consistency points.

Also left a separate general comment about the hardcoded `CLIENT_SESSION_KEEP_ALIVE` lookup — that change technically lives in a sibling PR in this Graphite stack, so I couldn't pin it inline here.

sf_core/tests/integration/session/heartbeat.rs
```diff
@@ -0,0 +1,483 @@
...
+    }
+    let count_at_drop = heartbeat_count.load(Ordering::SeqCst);
+
+    // Then no more heartbeat requests should arrive after the handle is dropped
+    tokio::time::sleep(Duration::from_millis(200)).await;
+    let count_after = heartbeat_count.load(Ordering::SeqCst);
+    assert_eq!(
+        count_at_drop, count_after,
+        "No heartbeats should be sent after HeartbeatHandle is dropped"
+    );
```
sfc-gh-makowalski, line 319: Minor flake risk on slow CI agents: between the handle being dropped and `count_at_drop = heartbeat_count.load(...)` running, an in-flight heartbeat request that crossed the wire just before the cancel could still be processed by wiremock and tick the counter. `send_heartbeat` is `select!`ed against `cancel_token.cancelled()`, so this is mostly safe, but it's not impossible for the counter to tick once between the two reads — turning the later `assert_eq!(count_at_drop, count_after)` into a flake.

A short grace sleep before snapshotting `count_at_drop` (e.g. `tokio::time::sleep(Duration::from_millis(20)).await;`) would let any in-flight request settle before the equality assertion. Same shape of risk lives around lines 158–167 in `heartbeat_stops_on_cancellation`, and to a lesser extent in `connection_close_cancels_heartbeat` (430–437) — though the connection-close path has a `cancel_and_wait` inside cleanup so it's the safest of the three.

sfc-gh-makowalski:
One more comment that I couldn't pin inline because it lives in a sibling PR in this Graphite stack (it touches `sf_core/src/apis/database_driver_v1/connection.rs`, which isn't part of #745's own diff):

The new `CLIENT_SESSION_KEEP_ALIVE` lookup is a bare string literal:

```rust
let keep_alive = merged_params
    .get("CLIENT_SESSION_KEEP_ALIVE")
    .map(|v| v.eq_ignore_ascii_case("true"))
    .unwrap_or(false);
```

A few lines below, the existing telemetry check uses a typed constant — `param_names::CLIENT_TELEMETRY_ENABLED.as_str()` — and `CLIENT_TELEMETRY_ENABLED` lives in `config/param_registry.rs`. There's no `CLIENT_SESSION_KEEP_ALIVE` constant in `param_registry.rs` yet; adding one next to `CLIENT_TELEMETRY_ENABLED` and using it here would match the existing pattern and remove a typo footgun. Not blocking — and probably better addressed on whichever stacked PR introduced this lookup.

sfc-gh-makowalski (changes requested):
PTAL at the comments I left

sfc-gh-makowalski (approved):
Approving up front, if all the comments are resolved then go ahead and merge

sf_core/tests/integration/session/heartbeat.rs
sfc-gh-mhofman: Dropping the assertion

sf_core/tests/integration/session/heartbeat.rs
sfc-gh-mhofman: ~50ms grace sleeps before snapshotting the `count_at_drop`/`count_at_close` baselines in `heartbeat_drop_cancels_task`, `heartbeat_stops_on_cancellation`, and `connection_close_cancels_heartbeat`  
added

sf_core/tests/integration/session/heartbeat.rs
sfc-gh-mhofman: Replacing this local ClientInfo for the pre-existing test fixtures

sf_core/src/apis/database_driver_v1/mod.rs
sfc-gh-mhofman: Restoring `pub(crate)` semantics by cfg-gating `pub mod heartbeat` behind `any(test, feature = "test-utils")` in `database_driver_v1/mod.rs`

sfc-gh-mhofman:
> One more comment that I couldn't pin inline because it lives in a sibling PR in this Graphite stack (it touches sf_core/src/apis/database_driver_v1/connection.rs, which isn't part of #745's own diff):
>
> The new CLIENT_SESSION_KEEP_ALIVE lookup is a bare string literal:
>
> let keep_alive = merged_params
> .get("CLIENT_SESSION_KEEP_ALIVE")
> .map(|v| v.eq_ignore_ascii_case("true"))
> .unwrap_or(false);
> A few lines below, the existing telemetry check uses a typed constant — param_names::CLIENT_TELEMETRY_ENABLED.as_str() — and CLIENT_TELEMETRY_ENABLED lives in config/param_registry.rs. There's no CLIENT_SESSION_KEEP_ALIVE constant in param_registry.rs yet; adding one next to CLIENT_TELEMETRY_ENABLED and using it here would match the existing pattern and remove a typo footgun. Not blocking — and probably better addressed on whichever stacked PR introduced this lookup.

​will add in a separate PR on top of this one

PR #723, opened 2026-03-25
SNOW-3264599: Python - implement cursor reset method

sfc-gh-asolarski:
### TL;DR

Implemented proper cursor state management by adding a `reset()` method and updating `close()` to properly clean up resources and return status indicators.

### What changed?

- **Added** **`reset()`** **method**: Clears all result-related state including `execute_result`, `_iterator`, `_description`, `_sqlstate`, `_fetch_mode`, `_binding_data`, and `_rownumber`. The `closing` parameter preserves `_rowcount` when set to `True`.
- **Enhanced** **`close()`** **method**: Now calls `reset(closing=True)`, clears the messages list, and returns a boolean indicating success (`True`), already closed (`False`), or exception (`None`).
- **Updated** **`execute()`** **method**: Added `_do_reset` parameter (defaults to `True`) and moved the reset logic to call `reset()` at the beginning instead of manually clearing state afterward.
- **Modified** **`executemany()`** **method**: Now calls `reset()` once before the loop for client-side binding scenarios and passes `_do_reset=False` to individual `execute()` calls to avoid redundant resets.

### Why make this change?

This change provides proper cursor lifecycle management by centralizing state cleanup logic in the `reset()` method. It ensures consistent behavior across different execution patterns, prevents state leakage between queries, and provides better error handling and status reporting for cursor operations. The `close()` method now properly indicates its success status, and `executemany()` avoids unnecessary resets between individual executions while maintaining clean state.

PR #696, opened 2026-03-24
SNOW-3243341: Python decorators for required conditions in cursor

sfc-gh-asolarski:

PR #653, opened 2026-03-20
SNOW-3243341: python wrapper - support pandas and pyarrow fetching

sfc-gh-asolarski:
### TL;DR

Added support for Arrow/Pandas fetch methods (`fetch_arrow_all`, `fetch_arrow_batches`, `fetch_pandas_all`, `fetch_pandas_batches`) to the Snowflake Python connector by implementing a new stream-based Arrow table iterator with Snowflake-specific type conversions.

### What changed?

- **New C++ Components**: Added `ArrowTableConverter` class to handle Snowflake-specific type conversions (scaled numbers, timestamps, etc.) and `CArrowStreamTableIterator` for stream-based Arrow batch processing
- **Python Interface**: Implemented `fetch_arrow_all`, `fetch_arrow_batches`, `fetch_pandas_all`, and `fetch_pandas_batches` methods in the cursor class
- **Type Safety**: Added fetch mode validation to prevent mixing row-by-row and Arrow/Pandas fetching within the same result set
- **Connection Properties**: Added `arrow_number_to_decimal` property to control numeric type conversion behavior
- **Build System**: Updated build configuration to include new C++ source files

PR #639, opened 2026-03-19
SNOW-3243341: optional dependencies in cursor (pyarrow, pandas)

sfc-gh-asolarski:

PR #750, opened 2026-03-27
SNOW-3264599: core + python - abort query method

sfc-gh-asolarski:


sfc-gh-asolarski:
* **#750** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/750?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/750?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#747** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/747?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#735** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/735?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#723** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/723?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#653** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/653?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#696** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/696?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#639** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/639?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#629** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/629?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#625** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/625?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

python/src/snowflake/connector/cursor.py
```diff
@@ -1031,9 +1032,16 @@ def query_result(self, qid: str) -> SnowflakeCursorBase:
 
         return self
 
+    @_requires_not_closed
+    @_requires_open_connection
```
sfc-gh-turbaszek: When looking at that I don't feel the difference. If connection is not closed then is open. Why we need both decorators?

python/tests/integ/test_cursor.py
```diff
@@ -1982,38 +1986,64 @@ def test_query_result_retrieves_data_and_metadata(self, connection):
...
-            cursor.query_result("00000000-0000-0000-0000-000000000000")
 
-    def test_query_result_raises_when_cursor_closed(self, connection):
-        """query_result raises InterfaceError on a closed cursor."""
-        cur = connection.cursor()
-        cur.close()
-        with pytest.raises(InterfaceError):
-            cur.query_result("some-qid")
+def _start_long_running_query(cur, connection) -> str:
+    """Workaround till new driver supports execute_async."""
```
sfc-gh-turbaszek: Let's move this logic to `execute_async` or let's make sure this test starts to fail once `execute_async` is implemented. This way we will be sure it will get adjusted once implementation is added

python/tests/unit/test_cursor.py
```diff
@@ -1822,3 +1822,64 @@ def test_query_result_propagates_rpc_error(self, cursor, mock_connection):
...
+        assert request.conn_handle == ConnectionHandle(id=1)
+        assert request.query_id == "01234567-abcd-ef01-0000-000000000001"
+
+    def test_abort_query_returns_false_on_failure(self, cursor, mock_connection):
+        """abort_query returns False when the server reports failure."""
+        mock_connection.db_api.connection_abort_query.return_value.success = False
+
+        result = cursor.abort_query("some-qid")
+
+        assert result is False
```
sfc-gh-turbaszek, line 2127: Does those tests really test the behavior? Imo we only test that `connection_abort_query` is called and proper value is returned

sfc-gh-turbaszek (changes requested):

python/src/snowflake/connector/cursor.py
sfc-gh-asolarski: check_not_closed is about cursor being closed or not, while check_connection_open about underlying connection.

When those are separated the following scenario is valid:

1. open connection
2. execute
3. connection is closed
4. cursor is not closed yet, so results can still be fetched

So, fetching requires cursor to be open, executing requires connection to be open.  
I assume that naming is a bit confusing, so I would rename to `_requires_not_closed` to `_requires_cursor_open`.  
But first, maybe this whole thing is overcomplicated, and we should just simplify "openess" of cursor. When either a connection or cursor is closed, then we consider the cursor closed and don't allow executing/fetching, wdyt?

sfc-gh-turbaszek: Oh, got it. I think having two different `_requires_not_closed` is viable. One for connection, one for cursor. The cursor one should invoke connection `_requires_not_closed`. As you pointed out, cursor is open iff connection is open.

python/tests/unit/test_cursor.py
sfc-gh-asolarski: The method is a thin delegation wrapper — constructing the request, forwarding it, and returning response.success is all it does. The unit test verifies that wiring (correct args, correct return value). Actual abort behavior is the responsibility of the backend/RPC layer and would be covered by integration tests.

sfc-gh-turbaszek (approved):

PR #754, opened 2026-03-30
SNOW-2872430: Chunk prefetch

sfc-gh-jszczerbinski:
This PR replaces the monolithic chunks.rs with a modular, trait-based prefetch pipeline. 

## Key changes

- New chunks/ module — splits responsibilities into prefetch.rs (generic pipeline), json_parser.rs (SIMD-accelerated JSON→Arrow), arrow_parser.rs, http_downloader.rs, and error.rs.

- JSON parsing rewrite — replaces the old serde_json-based approach with sonic-rs (SIMD) and streaming column builders, eliminating intermediate Vec<Vec<Option<String>>> allocations.

- Gzip offloaded to reqwest — the gzip feature on reqwest handles decompression transparently; the manual decode_chunk_body loop is gone.

- ChunkDownloadData enriched — now carries row_count, uncompressed_size, and compressed_size from the server response.

- Benchmarking infrastructure — a collect_chunk_data binary and prefetch_bench criterion benchmark with pprof flamegraph support.

## Highlights

- This solution reduces JSON parsing and conversion benchmark result from 2.5s to 800ms
  - Explored hand rolled SIMD parser for this use case and got down to 300-400 ms range. Decided to use generic SIMD enabled parser instead, due to maintanance and correctness concerns
  - Serde parsing was much slower than SIMD enabled Sonic with 1.6s benchmark result.
  - Tested also with simd_json crate, but it was slower than Sonic
  - Apart of JSON parsing saving the biggest improvement was evading string allocations between parsed JSON and Arrow array bulders

- The heavy compute for JSON use case might cause the driver to have async worker pool to be blocked for short periods of time when prefetching multiple result sets. With the optimizations it should be acceptable, but running chunk parsing on separate thread pool might be good improvement, but needs to be thought through.


sf_core/benches/prefetch_bench.rs
```diff
@@ -0,0 +1,346 @@
...
+            "FIXED" => RowType::fixed(
+                &self.name,
+                self.nullable,
+                self.precision.unwrap_or(38),
+                self.scale.unwrap_or(0),
+            ),
+            "TEXT" => RowType::text(
+                &self.name,
+                self.nullable,
+                self.length.unwrap_or(16777216),
```
sfc-gh-pfus: May be worth naming this magic number. Is it 128M?

sf_core/benches/prefetch_bench.rs
```diff
@@ -0,0 +1,346 @@
...
+                row_count: chunk_meta.map_or(0, |c| c.row_count),
+                uncompressed_size: chunk_meta.map_or(0, |c| c.uncompressed_size),
+                compressed_size: chunk_meta.map_or(0, |c| c.compressed_size),
+                headers: HashMap::new(),
+            }
+        })
+        .collect()
+}
+
+fn drain_reader(reader: &mut dyn RecordBatchReader) -> usize {
```
sfc-gh-pfus, line 146: This name is strange, maybe just `total_rows_from_reader`?

sf_core/src/bin/collect_chunk_data.rs
```diff
@@ -0,0 +1,436 @@
...
+    #[serde(rename = "SNOWFLAKE_TEST_HOST")]
+    host: Option<String>,
+    #[serde(rename = "SNOWFLAKE_TEST_DATABASE")]
+    database: Option<String>,
+    #[serde(rename = "SNOWFLAKE_TEST_SCHEMA")]
+    schema: Option<String>,
+    #[serde(rename = "SNOWFLAKE_TEST_WAREHOUSE")]
+    warehouse: Option<String>,
+    #[serde(rename = "SNOWFLAKE_TEST_WAREHOUSE_CORE")]
+    warehouse_core: Option<String>,
```
sfc-gh-pfus, line 59: What is the difference between this and regular warehouse?

sf_core/src/bin/collect_chunk_data.rs
```diff
@@ -0,0 +1,436 @@
+use std::path::{Path, PathBuf};
```
sfc-gh-pfus, line 1: Can we have a readme what is this for and how to use it?

sf_core/src/bin/collect_chunk_data.rs
```diff
@@ -0,0 +1,436 @@
...
+            save_arrow_data(&response.data, &cli.output_dir, &tls_client).await?;
+        }
+        sf_core::rest::snowflake::query_response::RowsetData::JsonMultiChunk { .. }
+        | sf_core::rest::snowflake::query_response::RowsetData::JsonRowset { .. } => {
+            save_json_data(&response.data, &cli.output_dir, &tls_client).await?;
+        }
+        sf_core::rest::snowflake::query_response::RowsetData::SchemaOnly { .. } => {
+            return Err("Query returned schema-only (no data)".into());
+        }
+        sf_core::rest::snowflake::query_response::RowsetData::NoData => {
```
sfc-gh-pfus, line 429: Is empty rowset an error? Is this the best handling of empty resultsets?

sf_core/src/chunks/error.rs
```diff
@@ -0,0 +1,56 @@
...
+        location: Location,
+    },
+    #[snafu(display("Failed to communicate with Snowflake to get chunk data"))]
+    Communication {
+        source: reqwest::Error,
+        #[snafu(implicit)]
+        location: Location,
+    },
+    #[snafu(display("Snowflake responded with non-successful HTTP status"))]
+    UnsuccessfulResponseHTTP {
```
sfc-gh-pfus: `UnsuccessfulHttpStatusCode`?

sf_core/src/chunks/json_parser.rs
```diff
@@ -0,0 +1,1157 @@
...
+    pub(super) fn push_value(&mut self, cell: &[u8]) -> Result<(), ArrowError> {
+        match self {
+            ColumnBuilder::Fixed { scale, storage, .. } => match storage {
+                FixedStorage::I64 { builder } => {
+                    if cell.len() <= 18 {
+                        let v = parse_i64_fixed_unchecked(cell, *scale);
+                        builder.append_value(v);
+                    } else {
+                        let v128 = parse_i128_from_bytes(cell, *scale)?;
+                        if let Ok(v) = i64::try_from(v128) {
```
sfc-gh-pfus, line 392: What if we already rescaled to decimals? It will fall into the other arm already, correct? So we use I64 only until the first I128, correct?

sf_core/src/chunks/json_parser.rs
```diff
@@ -0,0 +1,1157 @@
...
+            ColumnBuilder::TimestampStruct {
+                scale,
+                epoch_builder,
+                frac_builder,
+                nulls,
+            } => {
+                let (epoch, frac) = parse_epoch_fraction(cell, *scale)?;
+                epoch_builder.append_value(epoch);
+                frac_builder.append_value(frac * 10i32.pow((9 - *scale) as u32));
+                nulls.append(true);
```
sfc-gh-pfus, line 457: Null true or false?

sf_core/src/arrow_utils.rs
```diff
@@ -917,6 +297,7 @@ pub enum ArrowUtilsError {
 #[cfg(test)]
 mod tests {
```
sfc-gh-pfus, line 298: Should we move tests to the new module?

sf_core/benches/prefetch_bench.rs
sfc-gh-jszczerbinski: Our goal here is to exhaust the stream so the name is idiomatic

sf_core/src/bin/collect_chunk_data.rs
sfc-gh-jszczerbinski: This was added to spread the load in CI

sf_core/src/bin/collect_chunk_data.rs
sfc-gh-jszczerbinski: This a program that helps us to generate data for tests, I will add readme

sf_core/src/arrow_utils.rs
sfc-gh-jszczerbinski: That's the follow up, didn't want to blow up diff and wanted make sure that the same tests pass

sf_core/src/chunks/json_parser.rs
sfc-gh-jszczerbinski: Yes we use i64 column builder until we encounter first i128 value. Then we switch to i128 column builder and we parse every int as i128.

sf_core/src/bin/collect_chunk_data.rs
sfc-gh-jszczerbinski: We want to collect many chunks here for testing. Getting no chunks is an error for us

sf_core/src/chunks/json_parser.rs
sfc-gh-jszczerbinski: NullBuffer buffer docs
```
/// A [`BooleanBuffer`] used to encode validity for Arrow arrays
///
/// In the [Arrow specification], array validity is encoded in a packed bitmask with a
/// `true` value indicating the corresponding slot is not null, and `false` indicating
/// that it is null.
///
/// `NullBuffer`s can be creating using [`NullBufferBuilder`]
///
/// [Arrow specification]: https://arrow.apache.org/docs/format/Columnar.html#validity-bitmaps
/// [`NullBufferBuilder`]: crate::NullBufferBuilder
```

sfc-gh-pfus (approved):

PR #776, opened 2026-04-01
SNOW-3243354: Implement Connection.application property

sfc-gh-dheyman:
### TL;DR

Implemented the `application` property for Snowflake connections with a default value of "PythonConnector" and proper mapping to the underlying client configuration.

### What changed?

- Replaced the `NotImplementedError` in the `application` property getter with actual implementation
- Added logic to extract the `application` parameter from connection kwargs with "PythonConnector" as the default value
- The `application` value is now mapped to `client_app_id` in the connection options sent to the database driver
- The `application` parameter is removed from stored kwargs after processing

### How to test?

- Verify that `connection.application` returns "PythonConnector" by default
- Test setting a custom application name via the `application` parameter in connection initialization
- Confirm that the `client_app_id` option is properly set in the database driver configuration
- Ensure that `application` does not appear in the connection's stored kwargs

### Why make this change?

This implements a previously unimplemented feature that allows users to identify their applications when connecting to Snowflake, which is useful for monitoring and debugging purposes. The implementation ensures proper integration with the underlying database driver while maintaining clean separation of concerns.

### DIFF
--- a/python/src/snowflake/connector/connection.py
+++ b/python/src/snowflake/connector/connection.py
@@ -7,6 +7,7 @@ This module defines the Connection class as specified in PEP 249.
 from __future__ import annotations
 
 import logging
+import re
 
 from collections.abc import Generator, Iterable
 from io import StringIO
@@ -43,6 +44,9 @@ from .telemetry import TelemetryClient
 
 
 logger = logging.getLogger(__name__)
+
+CLIENT_NAME = "PythonConnector"
+APPLICATION_RE = re.compile(r"^[\w\d_]+$")
 
 SessionParameters = dict[str, Any]
 ConnectionParamValue = Union[int, str, float, bytes, bool, SessionParameters]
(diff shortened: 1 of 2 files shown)


sfc-gh-dheyman:
* **#776** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/776?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/776?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python/src/snowflake/connector/connection.py
```diff
@@ -103,6 +103,9 @@ def __init__(
         kwargs = self._rewrite_private_key_password(kwargs)
         kwargs = self._rewrite_mfa_params(kwargs)
 
+        self._application = str(kwargs.pop("application", "PythonConnector"))
```
sfc-gh-fpawlowski: PythonConnector could be some constant - it used to be called Client_name and was used in more places

sfc-gh-fpawlowski: we used to validate application name using APPLICATION_RE - should we keep that?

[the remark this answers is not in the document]
sfc-gh-fpawlowski: This is correct observation - I would be careful with explicit string conversion

sfc-gh-fpawlowski (commented):
Left few nits

sfc-gh-fpawlowski (approved):
Please take a look at the comments - most of them are nits, so leaving an approve :+1:

PR #782, opened 2026-04-02
SNOW-2872401: Make authenticator matching case-insensitive for old driver compatibility

sfc-gh-makowalski:
The old ODBC driver (snowflake-odbc) uses strcasecmp for all authenticator value comparisons, accepting any casing. The universal-driver used exact string matching, which broke backward compatibility for users passing lowercase values like "snowflake" or "programmatic_access_token".

Changes:
- Normalize authenticator values to uppercase before matching in both connection_config.rs (typed config path) and rest_parameters.rs (legacy path)
- Accept "snowflake" as an alias for password auth (in addition to "snowflake_password" and empty), matching the old driver constant S_AUTHENTICATOR_SNOWFLAKE = "snowflake"
- Stop sending AUTHENTICATOR field in the login JSON payload for plain password auth, matching old driver behavior where only PASSWORD is set
- Add unit tests for case-insensitive authenticator matching across all auth methods (password, PAT, MFA, JWT)
- Add e2e tests for lowercase PAT authenticator in core and ODBC
- Add Gherkin scenario for lowercase PAT authenticator compatibility

### DIFF
--- a/odbc_tests/tests/e2e/authentication/pat.cpp
+++ b/odbc_tests/tests/e2e/authentication/pat.cpp
@@ -210,6 +210,43 @@ TEST_CASE("should authenticate using PAT as token", "[pat]") {
   }
 }
 
+TEST_CASE("should authenticate using PAT as token with lowercase authenticator", "[pat]") {
+  // Given Authentication is set to lowercase programmatic_access_token and valid PAT token is provided
+  PatSetup pat_setup;
+  PatResult pat = pat_setup.acquire();
+
+  OLD_DRIVER_ONLY("BD#7") {
+    CHECK(pat.fetch_ret == SQL_ERROR);
+    REQUIRE(pat.diag_records.size() == 1);
+    CHECK(pat.diag_records[0].sqlState == "24000");
+    CHECK(pat.diag_records[0].nativeError == 10510);
+    CHECK_THAT(pat.diag_records[0].messageText, ContainsSubstring("Invalid cursor state"));
+  }
+
+  NEW_DRIVER_ONLY("BD#7") {
+    REQUIRE(pat.fetch_ret == SQL_SUCCESS);
+
+    std::string connection_string = get_pat_as_token_connection_string(pat.token_secret);
+    const std::string upper_auth = "AUTHENTICATOR=PROGRAMMATIC_ACCESS_TOKEN;";
+    const std::string lower_auth = "AUTHENTICATOR=programmatic_access_token;";
+    auto auth_pos = connection_string.find(upper_auth);
+    if (auth_pos != std::string::npos) {
+      connection_string.replace(auth_pos, upper_auth.size(), lower_auth);
+    }
+
+    auto env = setup_pat_environment();
+    auto dbc = get_pat_connection_handle(env);
+
+    // When Trying to Connect
+    attempt_pat_connection(dbc, connection_string);
+
+    // Then Login is successful and simple query can be executed
+    verify_pat_simple_query_execution(dbc);
+
+    SQLDisconnect(dbc.getHandle());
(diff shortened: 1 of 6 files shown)


sfc-gh-makowalski:
* **#783** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/783?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#782** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/782?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/782?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sf_core/src/config/connection_config.rs
```diff
@@ -561,22 +562,23 @@ pub fn validate_settings(settings: &ParamStore) -> Vec<ValidationIssue> {
...
         "" if has_private_key_params(settings) => {
             // Empty authenticator + private key params → auto-JWT, no password needed
         }
-        "SNOWFLAKE_PASSWORD" if has_private_key_params(settings) => {
+        "SNOWFLAKE" | "SNOWFLAKE_PASSWORD" if has_private_key_params(settings) => {
             issues.push(ValidationIssue {
                 severity: ValidationSeverity::Error,
                 parameter: AUTHENTICATOR.into(),
-                message: "Cannot specify 'SNOWFLAKE_PASSWORD' authenticator together with \
+                message: "Cannot specify password authenticator together with \
```
sfc-gh-mhofman: would it be more clear to say 'basic' authenticator?

odbc_tests/tests/e2e/authentication/pat.cpp
```diff
@@ -210,6 +210,43 @@ TEST_CASE("should authenticate using PAT as token", "[pat]") {
   }
 }
 
+TEST_CASE("should authenticate using PAT as token with lowercase authenticator", "[pat]") {
```
sfc-gh-mhofman, line 213: maybe worth considering such tests on integration level? would be faster...

sf_core/src/config/connection_config.rs
```diff
@@ -561,22 +562,23 @@ pub fn validate_settings(settings: &ParamStore) -> Vec<ValidationIssue> {
...
                           private key parameters; use 'SNOWFLAKE_JWT' or remove the private \
                           key parameters"
                     .into(),
                 code: ValidationCode::ConflictingParameters,
             });
         }
-        "SNOWFLAKE_PASSWORD" | "" => {
+        "SNOWFLAKE" | "SNOWFLAKE_PASSWORD" | "" => {
```
sfc-gh-mhofman, line 581: do we actually have something like SNOWFLAKE_PASSWORD value for auth value anywere?

sfc-gh-mhofman (approved):
couple of doubts only

sf_core/src/config/connection_config.rs
sfc-gh-makowalski: Yeah but only in the universal driver, it seems like it wasn't used before.

sf_core/src/config/connection_config.rs
sfc-gh-makowalski: Makes sense, I'll change that.

odbc_tests/tests/e2e/authentication/pat.cpp
sfc-gh-makowalski: TBF we have unit tests for this, but I think e2e tests are safer here, to make sure that snowflake responds to both.

PR #787, opened 2026-04-02
SNOW-2399385 performance tests instrumentation for core, python and odbc

sfc-gh-akolodziejczyk:
Adds core-level instrumentation to performance tests, breaking down `fetch_s` into three sub-metrics reported per iteration across all three drivers (Core, Python, ODBC):

- **`core_batch_wait_s`** — time the consumer blocks waiting for the next batch from the prefetch channel
- **`core_chunk_download_s`** — cumulative wall-clock time of HTTP chunk downloads
- **`core_arrow_decode_s`** — cumulative time spent decoding Arrow IPC / JSON into RecordBatches
- **`wrapper_time_s`** (Python/ODBC only) — `fetch_s - core_batch_wait_s`, isolating driver wrapper overhead

`wrapper_type_conversion_s` was removed. It was nearly identical to `wrapper_time_s` in practice, and with the current approach the wrapper overhead is derived from core metrics (`fetch_s - core_batch_wait_s`) rather than requiring explicit instrumentation inside each driver wrapper.

## **Collection methodology**

Metrics are collected via a custom `tracing::Layer` (`PerfTimingLayer`) that intercepts `#[instrument]` spans tagged with `target = "sf_core::perf"` on four hot-path methods in the chunk prefetch pipeline. On span close, the layer atomically accumulates elapsed nanoseconds into `AtomicU64` counters. Test apps reset before each fetch and read-and-reset after via a `#[repr(C)]` struct returned by value through FFI — no heap allocation, no serialization.

The `perf_timing` Cargo feature gates the layer and FFI logic. Span sites are unconditional (same pattern as the 48+ existing `#[instrument]` spans); without the layer registered they short-circuit in ~1-2ns.

## **Alternatives considered**

- **File/OTEL span export + post-processing**: Introduces I/O on the hot path (perturbing measurements), requires external infrastructure, and makes per-iteration reset/scoping complex. Rejected because measurement overhead must be near-zero.
- **In-memory span buffer (`Vec<SpanRecord>`)**: Requires `Mutex` or lock-free queue (contention under concurrent prefetch tasks), heap allocation per span, post-processing to aggregate, and complex FFI marshalling for a variable-size buffer. Atomic counters give the same aggregated result with zero allocation and zero contention.
- **OTEL Metrics API**: Adds a heavyweight dependency for three counters. The SDK's metric pipeline (readers, exporters, aggregation) is designed for observable production systems, not tight-loop benchmarking with per-iteration reset semantics.


sfc-gh-akolodziejczyk:
* **#787** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/787?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/787?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python/src/snowflake/connector/_internal/api_client/c_api.py
```diff
@@ -140,3 +165,21 @@ def register_default_logger_callback() -> None:
     Call this function explicitly to set up logging.
     """
     sf_core_init_logger(c_logger_callback)
+
+
+def sf_core_perf_enabled() -> bool:
+    return bool(core.sf_core_perf_enabled())
```
sfc-gh-turbaszek, line 173: Should we cache the output? It's 0/1 value that won't change during runtime right?

sfc-gh-akolodziejczyk: we can cache it, done

sfc-gh-jszczerbinski (approved):

sfc-gh-turbaszek (approved):

PR #793, opened 2026-04-07
SNOW-2314161: In-band telemetry core pipeline

sfc-gh-turbaszek:
## Summary
- Add OTel-based telemetry types (`TelemetryEventType`, `ErrorOrigin`) and serialization from OTel spans/metrics to Snowflake's in-band JSON format
- REST endpoint for `POST /telemetry/send` with auth token and error handling
- Custom `SpanExporter` and `PushMetricExporter` that serialize and ship telemetry via the Snowflake endpoint
- Add `metrics` feature to `opentelemetry` and `opentelemetry_sdk` deps; `experimental_metrics_custom_reader` as dev-dep

## Test plan
- [x] 17 unit tests (types, serialization, exporter behavior)
- [x] 20 integration tests (REST endpoint, exporter pipeline with wiremock)
- [ ] CI green

**Stack: 1/3** → #2 `telemetry/otlp-debug` → #3 `telemetry/wrapper-identity`

### DIFF
--- a/sf_core/src/rest/snowflake/telemetry.rs
+++ b/sf_core/src/rest/snowflake/telemetry.rs
@@ -15,6 +15,8 @@ const TELEMETRY_TIMEOUT: Duration = Duration::from_secs(30);
 #[derive(Debug, serde::Deserialize)]
 struct TelemetryResponse {
     success: bool,
+    message: Option<String>,
+    code: Option<String>,
 }
 
 /// POST a batch of telemetry events to the Snowflake in-band endpoint.
@@ -56,7 +58,11 @@ pub async fn send_telemetry(
         .context(InvalidSnowflakeResponseSnafu)?;
 
     if !parsed.success {
-        tracing::warn!("Telemetry endpoint returned success=false");
+        tracing::warn!(
+            code = parsed.code.as_deref().unwrap_or("none"),
+            message = parsed.message.as_deref().unwrap_or("none"),
+            "Telemetry endpoint returned success=false",
+        );
     }
 
     Ok(())
--- a/sf_core/src/rest/snowflake/telemetry.rs
+++ b/sf_core/src/rest/snowflake/telemetry.rs
@@ -1,0 +1,169 @@
+use std::time::Duration;
+
+use snafu::ResultExt;
+use url::Url;
+
+use crate::config::rest_parameters::ClientInfo;
+use crate::rest::snowflake::{
+    CommunicationSnafu, InvalidSnowflakeResponseSnafu, RequestConstructionSnafu, RestError,
+    UrlJoinSnafu, apply_json_content_type, apply_query_headers, read_response_json,
+};
+
+const TELEMETRY_SEND_PATH: &str = "/telemetry/send";
+const TELEMETRY_TIMEOUT: Duration = Duration::from_secs(30);
+
+#[derive(Debug, serde::Deserialize)]
+struct TelemetryResponse {
+    success: bool,
+}
+
+/// POST a batch of telemetry events to the Snowflake in-band endpoint.
+///
+/// The `payload` must conform to Snowflake's expected format:
+/// `{"logs": [{"message": {...}, "timestamp": "..."}]}`
+///
+/// Telemetry is best-effort: callers should handle errors gracefully.
+#[tracing::instrument(skip(client, client_info, session_token, payload))]
+pub async fn send_telemetry(
+    client: &reqwest::Client,
+    server_url: &Url,
+    client_info: &ClientInfo,
+    session_token: &str,
+    payload: &serde_json::Value,
+) -> Result<(), RestError> {
+    let url = server_url.join(TELEMETRY_SEND_PATH).context(UrlJoinSnafu {
+        path: TELEMETRY_SEND_PATH,
+    })?;
+
+    let request = apply_json_content_type(apply_query_headers(
+        client.post(url),
+        client_info,
+        session_token,
+    ))
+    .json(payload)
+    .timeout(TELEMETRY_TIMEOUT)
+    .build()
+    .context(RequestConstructionSnafu {
+        request: "telemetry",
+    })?;
+
+    let response = client.execute(request).await.context(CommunicationSnafu {
+        context: "Failed to execute telemetry request",
+    })?;
+
+    let parsed: TelemetryResponse = read_response_json(response)
+        .await
+        .context(InvalidSnowflakeResponseSnafu)?;
+
+    if !parsed.success {
+        tracing::warn!("Telemetry endpoint returned success=false");
+    }
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use serde_json::json;
+    use wiremock::matchers::{body_json, header_regex, method, path};
+    use wiremock::{Mock, MockServer, ResponseTemplate};
+
+    fn test_client_info() -> ClientInfo {
+        use crate::crl::config::CrlConfig;
+        use crate::tls::config::TlsConfig;
+        ClientInfo {
+            application: "TestApp".to_string(),
+            version: "1.0.0".to_string(),
+            os: "Linux".to_string(),
+            os_version: "5.15".to_string(),
+            ocsp_mode: None,
+            crl_config: CrlConfig::default(),
+            tls_config: TlsConfig::default(),
+        }
+    }
+
+    #[tokio::test]
+    async fn send_telemetry_success() {
+        let server = MockServer::start().await;
+
+        let payload = json!({
+            "logs": [{
+                "message": {"type": "test"},
+                "timestamp": "1234567890123"
+            }]
+        });
+
+        Mock::given(method("POST"))
+            .and(path("/telemetry/send"))
+            .and(header_regex("Authorization", r#"^Snowflake Token=".+"$"#))
+            .and(body_json(&payload))
+            .respond_with(ResponseTemplate::new(200).set_body_json(json!({
+                "success": true
+            })))
+            .expect(1)
+            .mount(&server)
+            .await;
+
+        let client = reqwest::Client::new();
+        let server_url = Url::parse(&server.uri()).unwrap();
+        let result = send_telemetry(
+            &client,
+            &server_url,
+            &test_client_info(),
+            "test_session_token",
+            &payload,
+        )
+        .await;
+
+        assert!(result.is_ok());
+    }
+
+    #[tokio::test]
+    async fn send_telemetry_server_error() {
+        let server = MockServer::start().await;
+
+        Mock::given(method("POST"))
+            .and(path("/telemetry/send"))
+            .respond_with(ResponseTemplate::new(500).set_body_string("Internal Server Error"))
+            .mount(&server)
+            .await;
+
+        let client = reqwest::Client::new();
+        let server_url = Url::parse(&server.uri()).unwrap();
+        let result = send_telemetry(
+            &client,
+            &server_url,
+            &test_client_info(),
+            "test_session_token",
+            &json!({"logs": []}),
+        )
+        .await;
+
+        assert!(result.is_err());
+    }
+
+    #[tokio::test]
+    async fn send_telemetry_session_expired() {
+        let server = MockServer::start().await;
+
+        Mock::given(method("POST"))
+            .and(path("/telemetry/send"))
+            .respond_with(ResponseTemplate::new(401))
+            .mount(&server)
+            .await;
+
+        let client = reqwest::Client::new();
+        let server_url = Url::parse(&server.uri()).unwrap();
+        let result = send_telemetry(
+            &client,
+            &server_url,
+            &test_client_info(),
+            "test_session_token",
+            &json!({"logs": []}),
+        )
+        .await;
+
+        assert!(result.is_err());
+    }
+}
--- a/sf_core/src/telemetry/mod.rs
+++ b/sf_core/src/telemetry/mod.rs
@@ -1,0 +1,6 @@
+pub mod types;
+
+#[doc(hidden)]
+pub mod serialization;
+#[doc(hidden)]
+pub mod snowflake_exporter;
--- a/sf_core/src/telemetry/mod.rs
+++ b/sf_core/src/telemetry/mod.rs
@@ -1,5 +1,6 @@
 pub mod types;
 
+// These modules are public for integration tests but are not part of the stable API.
 #[doc(hidden)]
 pub mod serialization;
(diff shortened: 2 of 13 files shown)


sfc-gh-turbaszek:
* **#796** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/796?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#795** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/795?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#794** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/794?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#793** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/793?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/793?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sf_core/src/telemetry/snowflake_exporter.rs
```diff
@@ -0,0 +1,445 @@
...
+
+            let payload = serialization::spans_to_snowflake_payload(&batch);
+            send_with_token(&session, &payload).await
+        }
+    }
+
+    fn shutdown_with_timeout(&mut self, _timeout: Duration) -> OTelSdkResult {
+        Ok(())
+    }
+}
```
sfc-gh-turbaszek, line 97: **TOCTOU race in metric exporter: token checked twice, can become None between checks**

The `PushMetricExporter::export` implementation reads the token lock to check availability (lines 88-93), drops the guard, serializes the payload, then calls `send_with_token` which acquires the lock *again*. Between the two lock acquisitions, the token can be revoked (set to `None`), causing `send_with_token` to silently drop the telemetry after the serialization work has already been done.

This isn't a correctness bug (the telemetry is best-effort), but the early check is misleading: it creates an impression that we avoid wasted serialization, yet it doesn't actually guarantee that. Consider either:
1. Removing the early token check and just relying on `send_with_token`'s check (simpler, same effective behavior), or
2. Holding the read guard across both serialization and send (but this would hold the lock longer, which may not be desirable).

Option 1 is probably the right call -- simplify and let `send_with_token` handle it.

sf_core/src/telemetry/types.rs
```diff
@@ -0,0 +1,52 @@
...
+/// Data provided by wrappers for session_init.
+#[derive(Debug, Clone)]
+pub struct SessionInitData {
+    pub driver_name: String,
+    pub driver_version: String,
+    pub language_runtime: String,
+    pub language_version: String,
+    pub language_compiler: Option<String>,
+    //pub release_date: Option<String>,
+    //pub is_lts: Option<bool>,
```
sfc-gh-turbaszek: **Commented-out fields should not be checked in**

These commented-out fields (`release_date`, `is_lts`) look like WIP placeholders. They should either be included with a TODO comment explaining when they'll be enabled, or removed entirely and tracked in a follow-up issue. Commented-out code tends to rot and creates confusion about whether these are planned, deferred, or abandoned.

sf_core/src/rest/snowflake/telemetry.rs
```diff
@@ -0,0 +1,169 @@
...
+    CommunicationSnafu, InvalidSnowflakeResponseSnafu, RequestConstructionSnafu, RestError,
+    UrlJoinSnafu, apply_json_content_type, apply_query_headers, read_response_json,
+};
+
+const TELEMETRY_SEND_PATH: &str = "/telemetry/send";
+const TELEMETRY_TIMEOUT: Duration = Duration::from_secs(30);
+
+#[derive(Debug, serde::Deserialize)]
+struct TelemetryResponse {
+    success: bool,
```
sfc-gh-turbaszek, line 17: **`TelemetryResponse` is missing `message` and `code` fields present in all other Snowflake response types**

Looking at `HeartbeatResponse`, `RefreshSessionResponse`, and `AuthenticatorRequestResponse`, they all include `message: Option<String>` and `code: Option<String>`. The `TelemetryResponse` only has `success: bool`.

When `success == false`, the warning log on line 59 provides no diagnostic information. Including `message` and `code` would allow a more useful log like:
```rust
tracing::warn!(
    "Telemetry endpoint returned success=false: code={}, message={}",
    parsed.code.as_deref().unwrap_or("none"),
    parsed.message.as_deref().unwrap_or("none"),
);
```

This follows the pattern used by `heartbeat.rs` and makes debugging production issues much easier.

sf_core/src/telemetry/serialization.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    json!({ "logs": logs })
+}
+
+macro_rules! collect_sum_data_points {
+    ($logs:expr, $metric_name:expr, $sum:expr, $ty:ty) => {{
+        let timestamp = system_time_to_epoch_millis($sum.time());
+        for dp in $sum.data_points() {
+            let mut message = serde_json::Map::new();
+            message.insert("type".to_string(), Value::String($metric_name.to_string()));
```
sfc-gh-turbaszek: **Macro defined after first use -- consider moving above or extracting to a helper function**

The `collect_sum_data_points!` macro is defined at line 56 but used starting at line 37. Rust allows this because of the `use collect_sum_data_points;` at line 73, but it makes the code harder to follow since readers encounter the invocations before seeing the definition.

More importantly, consider whether a macro is the right tool here. Since the three arms differ only in the Rust type (`u64`, `i64`, `f64`) and the `data_points()` / `value()` methods return types that all support `json!()` serialization, you could potentially use a generic helper function with a trait bound instead. That would give better error messages, IDE support, and type checking compared to a macro.

sf_core/src/telemetry/serialization.rs
```diff
@@ -0,0 +1,183 @@
...
+        }
+    }};
+}
+use collect_sum_data_points;
+
+fn otel_value_to_json(value: &opentelemetry::Value) -> Value {
+    use opentelemetry::Value as OtelValue;
+    match value {
+        OtelValue::Bool(b) => Value::Bool(*b),
+        OtelValue::I64(i) => json!(*i),
```
sfc-gh-turbaszek, line 102: **`unwrap_or_default()` silently maps pre-epoch times to 0**

If `SystemTime` is before `UNIX_EPOCH` (e.g., due to clock skew or a test bug), `duration_since(UNIX_EPOCH)` returns `Err` and `unwrap_or_default()` yields `Duration::ZERO`, producing a timestamp of `0`. This could make telemetry data confusing to debug on the server side -- a timestamp of `"0"` looks like missing data rather than a clock issue.

Consider logging a warning when this happens so it's at least observable:
```rust
fn system_time_to_epoch_millis(time: SystemTime) -> u128 {
    match time.duration_since(UNIX_EPOCH) {
        Ok(d) => d.as_millis(),
        Err(_) => {
            tracing::warn!("SystemTime before UNIX_EPOCH, using 0");
            0
        }
    }
}
```

sf_core/src/telemetry/types.rs
```diff
@@ -0,0 +1,52 @@
+/// Where the error originated.
```
sfc-gh-turbaszek, line 1: **`SessionInitData` and `WrapperErrorData` types are defined but never used**

These types are publicly exported but have no consumers in this PR. The `ErrorOrigin` enum is similarly unused beyond its own test. If these are intended for the next PR in the stack (#2 `telemetry/otlp-debug` or #3 `telemetry/wrapper-identity`), that's understandable, but it would be helpful to add a brief doc comment like `/// Used by wrapper-identity layer (see PR #N)` so reviewers know they're not dead code.

Without any usage, it's hard to review whether the API surface is correct -- for example, should `SessionInitData` fields like `driver_name` be `&str` or `Cow<'_, str>` to avoid unnecessary allocations when the data comes from static strings?

sf_core/src/telemetry/serialization.rs
```diff
@@ -0,0 +1,183 @@
...
+/// Format: `{"logs": [{"message": {...}, "timestamp": "..."}]}`
+pub fn spans_to_snowflake_payload(spans: &[SpanData]) -> Value {
+    let logs: Vec<Value> = spans.iter().map(span_to_log_entry).collect();
+    json!({ "logs": logs })
+}
+
+fn span_to_log_entry(span: &SpanData) -> Value {
+    let mut message = serde_json::Map::new();
+
+    message.insert("type".to_string(), Value::String(span.name.to_string()));
```
sfc-gh-turbaszek, line 18: **Span attribute keys can silently overwrite each other**

If a span has an attribute with key `"type"`, it will overwrite the span name that was inserted as `"type"` on line 18. Similarly, if an exception event has attributes with the same keys as the span's own attributes, the event attributes will silently overwrite them (since both use `message.insert()` on the same map).

This could lead to data loss in telemetry. Consider either:
- Prefixing the span name key (e.g., `"event.type"` or `"telemetry.type"`) to avoid collisions
- Documenting this as a known constraint ("span attributes must not use the key 'type'")
- Using a different data structure that preserves all values

sf_core/src/telemetry/mod.rs
```diff
@@ -0,0 +1,6 @@
+pub mod types;
+
+#[doc(hidden)]
+pub mod serialization;
```
sfc-gh-turbaszek, line 5: **`#[doc(hidden)]` on public modules: consider `pub(crate)` instead**

The `serialization` and `snowflake_exporter` modules are marked `#[doc(hidden)]` but remain fully `pub`. This means downstream code *can* still depend on them -- they just won't appear in docs. If the intent is to make these internal implementation details, `pub(crate)` would be a stronger guarantee that prevents external usage entirely. If they need to be `pub` for the integration tests, that's worth documenting with a comment.

sfc-gh-turbaszek (commented):
## Review Summary

This is a well-structured PR that adds the core in-band telemetry pipeline. The separation into REST endpoint, serialization, and exporter layers is clean, and the test coverage (37 tests across unit and integration) is solid. The best-effort error handling philosophy is correct for telemetry. A few substantive items to address:

### Key issues (posted as inline comments)

1. **Attribute key collision in serialization** -- the `"type"` key used for the span name can be overwritten by span attributes with the same key, and exception event attributes can overwrite span attributes. This is a silent data loss risk.

2. **TelemetryResponse missing `message`/`code` fields** -- every other Snowflake response type includes these, and they're needed for meaningful `success=false` diagnostics.

3. **TOCTOU in metric exporter** -- the early token-availability check provides a false sense of optimization since the token can change between the two lock acquisitions. Simplify by removing it.

4. **`#[doc(hidden)]` vs `pub(crate)`** -- if these modules are internal, use visibility restrictions rather than just hiding docs.

### Minor items

5. Commented-out fields in `SessionInitData` should be removed or documented.
6. `SessionInitData`, `WrapperErrorData`, `ErrorOrigin` are unused in this PR -- link to the consuming PR in the stack.
7. `system_time_to_epoch_millis` silently maps pre-epoch times to 0.
8. The `collect_sum_data_points!` macro could potentially be a generic function for better IDE/error support.

### What looks good

- The `send_with_token` pattern that holds the read guard across the reveal+send is a nice way to preserve `SensitiveString` zeroization guarantees.
- Delta temporality choice for metrics is correct for this use case (avoiding unbounded accumulation).
- The REST endpoint follows the existing `heartbeat.rs` pattern closely, making it easy to review.
- Integration tests cover the important edge cases: token revocation, token refresh, connection errors, server errors.

sf_core/src/telemetry/snowflake_exporter.rs
sfc-gh-turbaszek: Done. Removed the early token check entirely — `send_with_token` handles it.

sf_core/src/telemetry/types.rs
sfc-gh-turbaszek: Done. Removed the commented-out fields.

sf_core/src/rest/snowflake/telemetry.rs
sfc-gh-turbaszek: Done. Added `message` and `code` fields to `TelemetryResponse` and included them in the warning log, matching the `HeartbeatResponse` pattern.

sf_core/src/telemetry/serialization.rs
sfc-gh-turbaszek: Done. Moved the macro definition above `metrics_to_snowflake_payload` and removed the `use` statement. Kept it as a macro for now since the three arms differ in concrete types (`u64`, `i64`, `f64`) — a generic function would need a trait bound that the OTel SDK doesn't expose cleanly.

sf_core/src/telemetry/serialization.rs
sfc-gh-turbaszek: Done. Now logs a warning when `SystemTime` is before `UNIX_EPOCH` instead of silently returning 0.

sf_core/src/telemetry/types.rs
sfc-gh-turbaszek: Done. Added doc comments linking these types to the consuming PR #795. Also changed `WrapperErrorData.error_source` from `String` to `ErrorOrigin` for type safety.

sf_core/src/telemetry/serialization.rs
sfc-gh-turbaszek: Done. Two changes:
1. Span attributes with key `"type"` are now skipped (with a warning log) to protect the span name field.
2. Exception event attributes are prefixed with `"exception."` (unless they already have the prefix) to avoid overwriting span-level attributes.

sf_core/src/telemetry/mod.rs
sfc-gh-turbaszek: Can't use `pub(crate)` here — integration tests live outside the crate (`sf_core/tests/`) and need access to these modules. Added a comment explaining why they're `pub` + `#[doc(hidden)]` instead.

sf_core/src/telemetry/snowflake_exporter.rs
```diff
@@ -0,0 +1,445 @@
...
+
+    if let Err(e) = rest::send_telemetry(
+        &session.client,
+        &session.server_url,
+        &session.client_info,
+        tokens.session_token.reveal(),
+        payload,
+    )
+    .await
+    {
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: Done. `send_with_token` now clones the `SensitiveString` token under a short-lived read guard and drops the guard before the network call, avoiding lock contention with session refresh.

PR #798, opened 2026-04-07
Fix performance regression in python

sfc-gh-turbaszek:
The _requires_open and _requires_fetch_mode decorators on _fetchone() were executing on every row (1M times for a 1M row fetch), but these checks only need to run once per fetch operation. Move them to the caller level: fetchmany(), fetchall(), and the concrete fetchone() overrides.

Also inline the is_closed() method call in _requires_open to avoid an extra method dispatch per decorated call.


sfc-gh-turbaszek:
https://es-ci-legacy-mainvalidation-001.jenkinsdev1.us-west-2.aws-dev.app.snowflake.com/job/universal-driver-performance-tests/406/

PR #801, opened 2026-04-08
feat: support all C data types for boolean parameter binding

sfc-gh-makowalski:
Previously, ReadODBC for SnowflakeBoolean only handled SQL_C_BIT
(raw u8). Per the ODBC spec, SQL_BIT should accept all numeric C
types, SQL_C_CHAR, SQL_C_WCHAR, SQL_C_NUMERIC, and SQL_C_BINARY.

Added matching on binding.value_type for all integer types (nonzero
= true), float/double (nonzero = true, non-finite values rejected),
char/wchar (parse numeric string or "true"/"false"), numeric struct
(check val != 0), and binary (rejecting buffers with lenght other than 1).

CDataType::Default is handled explicitly (maps to u8, same as Bit).

Invalid boolean values (non-finite floats, unparseable strings) return
SQLSTATE 22018 (InvalidCharacterValueForCast) via a new
InvalidBooleanValue error variant in JsonBindingError.

Includes unit tests for each conversion path through the full
make_converter -> ParamConverter -> ReadODBC + WriteJson pipeline,
including edge cases: -0.0, NaN, ±inf, multi-byte binary buffers.

SNOW-3031772

Made-with: Cursor

### DIFF
--- a/odbc/src/conversion/boolean.rs
+++ b/odbc/src/conversion/boolean.rs
@@ -190,8 +190,26 @@ impl ReadODBC for SnowflakeBoolean {
             CDataType::UShort => Ok(read_unaligned::<u16>(binding) != 0),
             CDataType::SBigInt => Ok(read_unaligned::<i64>(binding) != 0),
             CDataType::UBigInt => Ok(read_unaligned::<u64>(binding) != 0),
-            CDataType::Float => Ok(read_unaligned::<f32>(binding) != 0.0),
-            CDataType::Double => Ok(read_unaligned::<f64>(binding) != 0.0),
+            CDataType::Float => {
+                let v = read_unaligned::<f32>(binding);
+                if !v.is_finite() {
+                    return UnsupportedCDataTypeSnafu {
+                        c_type: binding.value_type,
+                    }
+                    .fail();
+                }
+                Ok(v != 0.0)
+            }
+            CDataType::Double => {
+                let v = read_unaligned::<f64>(binding);
+                if !v.is_finite() {
+                    return UnsupportedCDataTypeSnafu {
+                        c_type: binding.value_type,
+                    }
+                    .fail();
+                }
+                Ok(v != 0.0)
+            }
             CDataType::Char => {
                 let s = read_char_str(binding)?;
                 parse_str_to_bool(&s, binding.value_type)
--- a/odbc/src/conversion/boolean.rs
+++ b/odbc/src/conversion/boolean.rs
@@ -4,14 +4,14 @@ use serde_json::Value;
 
 use crate::api::CDataType;
 use crate::api::ParameterBinding;
+use crate::conversion::error::JsonBindingError;
 use crate::conversion::error::UnsupportedCDataTypeSnafu;
-use crate::conversion::error::{JsonBindingError, NumericMagnitudeOverflowSnafu};
 use crate::conversion::error::{ReadArrowError, UnsupportedOdbcTypeSnafu, WriteOdbcError};
 use crate::conversion::numeric_helpers::{
     reject_multi_field_interval, write_interval_second, write_single_field_interval,
 };
 use crate::conversion::param_binding::{
-    read_char_str, read_numeric_struct, read_unaligned, read_wchar_str,
+    buffer_data_len, read_char_str, read_unaligned, read_wchar_str,
 };
 use crate::conversion::traits::Binding;
 use crate::conversion::traits::{ReadODBC, SnowflakeLogicalType, WriteJson};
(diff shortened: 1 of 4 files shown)


sfc-gh-makowalski:
* **#802** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/802?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#801** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/801?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/801?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

odbc/src/conversion/boolean.rs
```diff
@@ -152,12 +155,86 @@ impl WriteODBCType for SnowflakeBoolean {
...
+/// converted to a numeric value, then 0 → false, nonzero → true.
+/// Also accepts "true"/"false" literals for Snowflake compatibility.
+fn parse_str_to_bool(s: &str, c_type: CDataType) -> Result<bool, JsonBindingError> {
+    let trimmed = s.trim();
+    if let Ok(i) = trimmed.parse::<i64>() {
+        return Ok(i != 0);
+    }
+    if let Ok(f) = trimmed.parse::<f64>() {
+        return match f.is_finite() {
+            true => Ok(f != 0.0),
```
sfc-gh-pbulawa, line 170: What about `-0.0`?

odbc/src/conversion/boolean.rs
```diff
@@ -152,12 +155,86 @@ impl WriteODBCType for SnowflakeBoolean {
...
+            CDataType::UBigInt => Ok(read_unaligned::<u64>(binding) != 0),
+            CDataType::Float => {
+                let v = read_unaligned::<f32>(binding);
+                if !v.is_finite() {
+                    return UnsupportedCDataTypeSnafu {
+                        c_type: binding.value_type,
+                    }
+                    .fail();
+                }
+                Ok(v != 0.0)
```
sfc-gh-pbulawa, line 211: What about `-0.0`?

odbc/src/conversion/boolean.rs
```diff
@@ -152,12 +155,86 @@ impl WriteODBCType for SnowflakeBoolean {
...
+            }
+            CDataType::Double => {
+                let v = read_unaligned::<f64>(binding);
+                if !v.is_finite() {
+                    return UnsupportedCDataTypeSnafu {
+                        c_type: binding.value_type,
+                    }
+                    .fail();
+                }
+                Ok(v != 0.0)
```
sfc-gh-pbulawa, line 221: Again, `-0.0` is missing

odbc/src/conversion/boolean.rs
```diff
@@ -152,12 +155,86 @@ impl WriteODBCType for SnowflakeBoolean {
...
+/// Also accepts "true"/"false" literals for Snowflake compatibility.
+fn parse_str_to_bool(s: &str, c_type: CDataType) -> Result<bool, JsonBindingError> {
+    let trimmed = s.trim();
+    if let Ok(i) = trimmed.parse::<i64>() {
+        return Ok(i != 0);
+    }
+    if let Ok(f) = trimmed.parse::<f64>() {
+        return match f.is_finite() {
+            true => Ok(f != 0.0),
+            false => UnsupportedCDataTypeSnafu { c_type }.fail(),
```
sfc-gh-pbulawa: The datatype is supported, but the value is wrong. Can you introduce a new error, e.g. `InvalidValue/InvalidBooleanValue`?

odbc/src/conversion/boolean.rs
```diff
@@ -152,12 +155,86 @@ impl WriteODBCType for SnowflakeBoolean {
...
+            CDataType::Numeric => {
+                let n = read_unaligned::<sql::Numeric>(binding);
+                Ok(u128::from_le_bytes(n.val) != 0)
+            }
+            CDataType::Binary => {
+                let len = buffer_data_len(binding);
+                if len == 0 {
+                    return Ok(false);
+                }
+                let first = unsafe { *(binding.parameter_value_ptr as *const u8) };
```
sfc-gh-pbulawa: The whole buffer should be checked, as any non-zero value should produce `true`

sfc-gh-pbulawa:
Add E2E tests for the changes.

sfc-gh-pbulawa (changes requested):

odbc/src/conversion/boolean.rs
sfc-gh-makowalski: Done — now rejects len != 1 with SQLSTATE 22003 (NumericValueOutOfRange).

The ODBC spec for SQL_C_BINARY → SQL_BIT says the byte length must equal the SQL data length (1 byte), otherwise SQLSTATE 22003.

sfc-gh-pbulawa: If that's the case, let's add a length check and make it on-par with the ODBC API and the old driver. Is the old driver spec-compliant with the explicit check?

sfc-gh-makowalski:
> Add E2E tests for the changes.

​That's added in the next PR (802), here we have unit tests only.

odbc/src/conversion/boolean.rs
sfc-gh-makowalski: Confirmed on CI: the old ODBC driver also rejects multi-byte binary buffers for SQL_BIT. Our implementation now matches both the spec and the old driver.

sfc-gh-pbulawa (approved):

PR #806, opened 2026-04-08
SNOW-2314161: In-band telemetry

sfc-gh-turbaszek:
Add OTel-based telemetry infrastructure to sf_core:
- Telemetry types (TelemetryEventType, ErrorOrigin)
- Serialization from OTel spans/metrics to Snowflake JSON format
- REST endpoint for POST /telemetry/send
- Custom SpanExporter and PushMetricExporter that serialize and ship
  telemetry data through the Snowflake in-band telemetry endpoint

Example event 
```json
{
  "host.arch": "x86_64",
  "os.type": "linux",
  "os.version": "24.04",
  "process.runtime.compiler": "Clang 21.1.4 ",
  "process.runtime.name": "CPython",
  "process.runtime.version": "3.14.2",
  "service.name": "snowflake-connector-python",
  "service.version": "2026.0.0",
  "snowflake.session.id": 3563548706699139,
  "type": "session_init"
}
```


sf_core/src/telemetry/serialization.rs
```diff
@@ -0,0 +1,198 @@
...
+            for kv in dp.attributes() {
+                message.insert(kv.key.as_str().to_string(), otel_value_to_json(&kv.value));
+            }
+            $logs.push(json!({
+                "message": message,
+                "timestamp": timestamp.to_string()
+            }));
+        }
+    }};
+}
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: done

sf_core/src/apis/database_driver_v1/connection.rs
```diff
@@ -123,21 +123,39 @@ impl DatabaseDriverV1 {
...
-                        host,
-                        port,
-                        login_parameters.server_url.clone(),
-                        login_parameters.client_info.clone(),
-                        merged_params,
-                        login_final_names,
-                        resolved_snapshot,
-                    )
-                    .await;
+                    .get("CLIENT_TELEMETRY_ENABLED")
```
sfc-gh-jszczerbinski: Use param registry

protobuf/database_driver_v1.proto
```diff
@@ -342,6 +342,12 @@ message ConnectionSetOptionBoolResponse {
 message ConnectionInitRequest {
   ConnectionHandle conn_handle = 1;
   DatabaseHandle db_handle = 2;
+  // Wrapper identity fields — used to emit session_init telemetry during connection init.
```
sfc-gh-asolarski: is this a static set of fields or can it evolve?
maybe we could encapsulate that into something meaningful, like
```
TelemetryInitParams {
  optional string driver_name = 1;
  optional string driver_version = 2;
  optional string language_runtime = 3;
  optional string language_version = 4;
  optional string language_compiler = 5;
}
```

python/src/snowflake/connector/connection.py
```diff
@@ -179,7 +183,17 @@ def __init__(
                 ConnectionSetSessionParametersRequest(conn_handle=self.conn_handle, parameters=session_params)
             )
 
-        self.db_api.connection_init(ConnectionInitRequest(conn_handle=self.conn_handle, db_handle=self.db_handle))
+        self.db_api.connection_init(
+            ConnectionInitRequest(
+                conn_handle=self.conn_handle,
+                db_handle=self.db_handle,
+                driver_name="snowflake-connector-python",
```
sfc-gh-asolarski: maybe let's extract that to a constant?

python/tests/integ/telemetry/test_connection_telemetry.py
```diff
@@ -0,0 +1,85 @@
...
+    the request path, headers, and top-level payload structure. It runs against both
+    the universal and old drivers for comparison.
+    """
+    with WiremockClient().start() as wiremock:
+        wiremock.add_mapping("auth/login_success_jwt.json")
+        wiremock.add_mapping("telemetry/telemetry_send_success.json")
+
+        # The old driver needs private_key as DER bytes; the universal driver uses private_key_file
+        extra_params = {}
+        if not IS_UNIVERSAL_DRIVER:
```
sfc-gh-asolarski: we could stick to the convention and use `is_new_driver()` from the `compatibility.py`

python/tests/unit/test_internal_telemetry.py
```diff
@@ -0,0 +1,43 @@
...
+
+from snowflake.connector._internal.protobuf_gen.database_driver_v1_pb2 import (
+    ConnectionHandle,
+    DatabaseHandle,
+)
+from snowflake.connector.version import __version__
+from tests.compatibility import IS_UNIVERSAL_DRIVER
+
+
+pytestmark = pytest.mark.skipif(not IS_UNIVERSAL_DRIVER, reason="Requires universal driver")
```
sfc-gh-asolarski: why is it needed in the unit tests? (I don't even think it's really used)

python/tests/wiremock_client.py
```diff
@@ -119,6 +119,24 @@ def add_mapping(self, mapping_path: str, placeholders: Optional[dict[str, str]]
...
+            url_path_pattern: Regex pattern to match against request URL paths
+                             (e.g., "/telemetry/send")
+
+        Returns:
+            List of request objects captured by Wiremock.
+        """
+        response = requests.post(
+            f"{self.http_url()}/__admin/requests/find",
+            json={"urlPathPattern": url_path_pattern},
+        )
```
sfc-gh-asolarski, line 136: isn't copilot comment valid here? a timeout on this request might be useful

python/pyproject.toml
```diff
@@ -67,6 +67,7 @@ test = [
 dev = [
     "ruff",
     "mypy",
+    "mypy-protobuf",
```
sfc-gh-asolarski, line 70: what's the reasoning about adding this dep at this point?

sf_core/src/apis/database_driver_v1/connection.rs
```diff
@@ -123,21 +123,39 @@ impl DatabaseDriverV1 {
...
+                    .map(|v| v.eq_ignore_ascii_case("true"))
+                    .unwrap_or(true);
```
sfc-gh-asolarski, line 276: hmm, are we correctly assuming here that no information about telemetry returned from the backend equals "enabled"?

sf_core/src/apis/database_driver_v1/connection.rs
```diff
@@ -123,21 +123,39 @@ impl DatabaseDriverV1 {
...
+                drop(conn);
+
+                // Best-effort session_init telemetry — spawned as a background task
+                // so connection open never blocks on telemetry latency.
+                if telemetry_enabled {
+                    let conn_ptr_clone = conn_ptr.clone();
+                    tokio::spawn(async move {
+                        Self::send_session_init_telemetry(&conn_ptr_clone).await;
+                    });
+                }
```
sfc-gh-asolarski, line 286: so here we
1. drop the lock
2. spawn new task
3. reaquire the lock
4. copy needed data from the connection
5. unlock
6. send the span

can't we copy/extract needed data before dropping the lock and avoid locking connection for the second time at all? 
then we can do this extraction in a separate function, and we don't even need to clone the pointer to the connection just make `send_session_init_telemetry` function accept the extracted parameters

sf_core/src/apis/database_driver_v1/connection.rs
```diff
@@ -147,6 +165,53 @@ impl DatabaseDriverV1 {
...
+        };
+        // Mutex is now dropped.
+
+        // Step 2: Read session_id from the tokens RwLock without holding the Mutex.
+        let session_id = tokens_arc
+            .read()
+            .await
+            .as_ref()
+            .map(|t| t.session_id)
+            .unwrap_or(0);
```
sfc-gh-asolarski, line 334: is it safe to set the default session id to 0?

sf_core/src/apis/database_driver_v1/connection.rs
```diff
@@ -295,6 +360,41 @@ impl DatabaseDriverV1 {
...
+
+/// Wrapper identity set once via `ConnectionInit` and attached to all subsequent telemetry events.
+#[derive(Debug, Clone, Default)]
+pub struct WrapperIdentity {
+    pub driver_name: String,
+    pub driver_version: String,
+    pub language_runtime: String,
+    pub language_version: String,
+    /// `None` means compiler info is not applicable for this language.
+    pub language_compiler: Option<String>,
```
sfc-gh-asolarski, line 529: here only the `langugage_compiler` is optional, but in protobuf message definition all fields were defined optional, shouldn't we be consistent?

sf_core/src/telemetry/environment.rs
```diff
@@ -0,0 +1,143 @@
...
+            driver_name: String::new(),
+            driver_version: String::new(),
+            language_runtime: String::new(),
+            language_version: String::new(),
+            language_compiler: None,
+        }
+    }
+}
+
+fn detect_os_version() -> String {
```
sfc-gh-asolarski, line 47: couldn't we use some create for that? for example https://crates.io/crates/os_info

protobuf/database_driver_v1.proto
sfc-gh-turbaszek: Good point, this should be `WrapperIdentity` as similar set will be required for `ClinetInfo`

sfc-gh-jszczerbinski (approved):

python/pyproject.toml
sfc-gh-turbaszek: It's missing from dev setup, it's unrelated change

sf_core/src/apis/database_driver_v1/connection.rs
sfc-gh-turbaszek: Done — added `CLIENT_TELEMETRY_ENABLED` to `param_names` in `param_registry.rs` and replaced the string literal.

protobuf/database_driver_v1.proto
sfc-gh-turbaszek: Done — created a `WrapperIdentity` message and replaced the flat fields with `optional WrapperIdentity wrapper_identity = 3` in `ConnectionInitRequest`.

python/src/snowflake/connector/connection.py
sfc-gh-turbaszek: Done — extracted to `CLIENT_NAME = "snowflake-connector-python"` at module level.

python/tests/integ/telemetry/test_connection_telemetry.py
sfc-gh-turbaszek: Done — switched to `is_new_driver()` from `compatibility.py`.

python/tests/unit/test_internal_telemetry.py
sfc-gh-turbaszek: The `skipif` is needed — this test exercises the protobuf-based `connection_init` which only exists in the universal driver. Without it the test would fail with an ImportError on the old connector. Switched to `is_new_driver()` for consistency.

python/tests/wiremock_client.py
sfc-gh-turbaszek: Done — added `timeout=5` to all `requests.post()` calls in wiremock_client.py.

python/pyproject.toml
sfc-gh-turbaszek: `mypy-protobuf` is already in `build-system.requires` for wheel builds. Adding it to `dev` deps ensures developers can run `mypy` locally with protobuf stubs without doing a full build — standard convenience for local type checking.

sf_core/src/apis/database_driver_v1/connection.rs
sfc-gh-turbaszek: Yes — Snowflake server defaults `CLIENT_TELEMETRY_ENABLED` to true and only sends `"false"` when the account/user has opted out. If the parameter is absent (e.g. older server), defaulting to enabled matches server behavior. Added a comment explaining this reasoning.

sf_core/src/apis/database_driver_v1/connection.rs
sfc-gh-turbaszek: Agreed and already addressed — `send_session_init_telemetry` now extracts all needed data (env_info, tokens Arc, query_parameters, http_client) under the connection mutex in a single block, drops the mutex, then reads the tokens RwLock and sends. No double-locking.

sf_core/src/apis/database_driver_v1/connection.rs
sfc-gh-turbaszek: Yes — after a successful login, tokens are always present so this path shouldn't normally trigger. The `unwrap_or(0)` is a defensive fallback; session_id 0 acts as a harmless "no session" sentinel in the telemetry payload. Added a comment explaining this.

sf_core/src/apis/database_driver_v1/connection.rs
sfc-gh-turbaszek: Keeping the Rust struct fields as plain `String` (with `unwrap_or_default()` at the proto boundary) since it simplifies all downstream usage — every consumer can treat them as always-present strings. The `language_compiler` stays `Option<String>` because `None` has distinct semantic meaning ("not applicable for this language").

sf_core/src/telemetry/environment.rs
sfc-gh-turbaszek: Done — replaced the manual platform-specific detection with `os_info = "3"`. Simplifies ~40 lines of `#[cfg]` blocks down to a few lines.

PR #809, opened 2026-04-09
SNOW-2433865 validate returned rows in perf tests, improve error propagation

sfc-gh-akolodziejczyk:
Row count validation (all drivers):
Auto-extract LIMIT N from SQL commands and pass as EXPECTED_ROW_COUNT to driver containers, so every SELECT test validates it fetched exactly the expected number of rows
Applies to both E2E and recorded HTTP tests with zero changes to test definitions
Reject 0-row baselines immediately with a clear error message pointing to likely causes (async timeout, server error)

Error propagation fixes:
run_container now raises RuntimeError with the last 20 lines of container output on non-zero exit, instead of silently continuing to verify_results which would report a generic "No result files found"
Rust core fails explicitly when response.result is None instead of silently returning 0 rows
Python driver now logs the actual exception for setup query and test execution failures before sys.exit(1)
WireMock CA cert export failure raises instead of silently returning, which caused misleading downstream errors

### DIFF
--- a/tests/performance/drivers/core/app/src/query_execution.rs
+++ b/tests/performance/drivers/core/app/src/query_execution.rs
@@ -89,7 +89,7 @@ pub fn run_test_iterations(
     iterations: usize,
 ) -> Result<Vec<IterationResult>> {
     let mut results = Vec::with_capacity(iterations);
-    let mut expected_row_count = get_expected_row_count();
+    let mut expected_row_count = get_expected_row_count()?;
 
     for i in 0..iterations {
         let (query_time, fetch_time, row_count, cpu_time_s, peak_rss_mb) =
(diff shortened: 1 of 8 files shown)


sfc-gh-akolodziejczyk:
* **#809** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/809?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/809?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sfc-gh-pcyrek (approved):
lgtm

PR #814, opened 2026-04-09
NO-SNOW: Implement callproc method on cursor and clean up stubs

sfc-gh-turbaszek:
Implement the PEP 249 callproc method following the reference driver's approach: build a CALL statement with paramstyle-aware placeholders and delegate to execute(). Also remove unused cancel/ping stubs from connection, switch nextset to NotImplementedError, and remove ResultMetadataV2 from the API comparison exclusion list.

Made-with: Cursor

### DIFF
--- a/python/src/snowflake/connector/connection.py
+++ b/python/src/snowflake/connector/connection.py
@@ -42,7 +42,7 @@ from ._internal.extras import numpy as np
 from ._internal.text_utils import split_statements
 from .constants import QueryStatus
 from .cursor import CursorInstance, CursorType, DictCursor, SnowflakeCursor
-from .errors import Error, InterfaceError, NotSupportedError, ProgrammingError
+from .errors import Error, InterfaceError, ProgrammingError
 from .telemetry import TelemetryClient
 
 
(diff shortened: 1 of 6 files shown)


python/src/snowflake/connector/connection.py
```diff
@@ -262,28 +262,6 @@ def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
...
-        """
-        Check if the connection to the server is still alive.
-
-        Returns:
-            bool: True if connection is alive, False otherwise
-
-        Raises:
-            NotSupportedError: If not implemented
-        """
-        raise NotSupportedError("ping is not implemented")
```
sfc-gh-turbaszek, line 285: Removing as those do not exist on old driver

python/src/snowflake/connector/cursor/_base.py
```diff
@@ -283,22 +269,41 @@ def sqlstate(self) -> str | None:
         """The SQLSTATE code of the last executed operation."""
         return self._sqlstate
 
+    @overload
```
sfc-gh-asolarski, line 275: do we really need those overloads?

python/src/snowflake/connector/cursor/_base.py
```diff
@@ -283,22 +269,41 @@ def sqlstate(self) -> str | None:
...
+            raise TypeError(f"callproc args must be a sequence (e.g. list or tuple), not {type(args).__name__}")
+        if not isinstance(args, Sequence):
+            raise TypeError(f"callproc args must be a sequence (e.g. list or tuple), not {type(args).__name__}")
+        paramstyle = self._connection.paramstyle
+        if paramstyle.is_client_side():
+            placeholders = ", ".join("%s" for _ in range(len(args)))
+        elif paramstyle == ParamStyle.NUMERIC:
+            placeholders = ", ".join(f":{i}" for i in range(1, len(args) + 1))
+        else:
+            placeholders = ", ".join("?" for _ in range(len(args)))
```
sfc-gh-asolarski: I would extract this logic to a function at least

maybe it shouldn't even belong to the cursor, it seems to be closer to binding_converters.py where we define `ParamStyle` 🤔

sfc-gh-asolarski (approved):
small remarks, overall lgtm

python/src/snowflake/connector/cursor/_base.py
sfc-gh-turbaszek: This is part of back compat, I think they do not hurt

PR #823, opened 2026-04-10
Python: implement Connection.paramstyle setter

sfc-gh-pczajka:
Implement paramstyle setter to avoid [error detected in SnowPy tests](https://github.com/snowflake-eng/snowpy/actions/runs/24240715059/job/70774327960):
`AttributeError: 'str' object has no attribute 'is_client_side'`

### DIFF
--- a/python/src/snowflake/connector/_internal/binding_converters.py
+++ b/python/src/snowflake/connector/_internal/binding_converters.py
@@ -65,15 +65,6 @@ class ParamStyle(Enum):
         available = [s.value for s in cls]
         raise ProgrammingError(f"Invalid paramstyle: {value!r}. Supported: {', '.join(sorted(available))}")
 
-    @classmethod
-    def from_str_or_class(cls, value: str | ParamStyle) -> ParamStyle:
-        """Return a member from a :class:`ParamStyle` or a PEP 249 paramstyle string."""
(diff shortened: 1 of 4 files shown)


python/src/snowflake/connector/_internal/binding_converters.py
```diff
@@ -65,6 +65,15 @@ def from_string(cls, value: str) -> ParamStyle:
         available = [s.value for s in cls]
         raise ProgrammingError(f"Invalid paramstyle: {value!r}. Supported: {', '.join(sorted(available))}")
 
+    @classmethod
+    def from_str_or_class(cls, value: str | ParamStyle) -> ParamStyle:
```
sfc-gh-turbaszek: Should we have `from_str` and `from_param_style` and handle if case in setter? I'm personally not big fan of overloading classmethods

sfc-gh-pczajka: This way we would have to implement that logic twice - in setter and in __init__. We can also create helper outside of this class (parse_paramstyle), WDYTH?

sfc-gh-turbaszek: In that case let's keep this approach

sfc-gh-turbaszek (approved):

PR #830, opened 2026-04-13
SNOW-3072787: python - introduce pd_writer for sqlalchemy integration

sfc-gh-asolarski:


sfc-gh-asolarski:
* **#830** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/830?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/830?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python/src/snowflake/connector/pandas_tools.py
```diff
@@ -74,3 +101,56 @@ def write_pandas(
         **kwargs,
     )
     return WritePandasOperation(cfg).execute()
+
+
+@requires_dependency(pandas, sqlalchemy)
+@_reject_kwargs("table", "conn", "keys", "data_iter")
```
sfc-gh-turbaszek, line 125: I don't understand. What exactly is the purpose of this code? Are we only passing `kwargs`?

sfc-gh-asolarski: I've expanded the docstring on `_reject_kwargs` to explain the rationale in detail. 
In short: both `pd_writer` and `make_pd_writer` forward `**kwargs` to `write_pandas()`, but some kwarg names overlap with arguments that pandas to_sql supplies positionally at call time. Without this guard, `functools.partial` would silently pre-bind a conflicting value, leading to a confusing `TypeError: got multiple values for argument`. The decorator catches this early and raises a clear `ProgrammingError` instead.

python/src/snowflake/connector/pandas_tools.py
```diff
@@ -74,3 +119,56 @@ def write_pandas(
...
+
+    Example::
+
+        df.to_sql("t", engine, index=False, method=make_pd_writer(parallel=1, quote_identifiers=False))
+    """
+    return partial(pd_writer, **kwargs)
+
+
+@requires_dependency(pandas, sqlalchemy)
+@_reject_kwargs("conn", "df", "table_name", "schema")
```
sfc-gh-turbaszek, line 142: Should we add a test that all arguments from pd_writer are rejected? Such a test can catch need for update of rejected kwargs when signature is extended. WDYT?

sfc-gh-asolarski: done

sfc-gh-turbaszek (approved):

PR #850, opened 2026-04-14
Fixup SnowPy tests

sfc-gh-pczajka:
SnowPy uses private member "_paramstyle", which was breaking the value parsing. This PR fixes it up


sfc-gh-fpawlowski (approved):
lgtm

PR #856, opened 2026-04-15
SNOW-3282031 Add a test suite for running all tests with JSON result format

sfc-gh-pfus:


sfc-gh-pfus:
* **#856** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/856?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/856?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/test-python.yml
sfc-gh-asolarski, line 1: please add it also for the "test-pandas" hatch_env, then we will have it covered across all cases (including directly fetching via pyarrow/pandas)

sfc-gh-asolarski (changes requested):

python/tests/e2e/types/test_float.py
```diff
@@ -129,6 +129,7 @@ def test_should_handle_special_float_values_from_literals_for_float_and_synonyms
         BOUNDARY_LITERAL_CASES,
         ids=["max", "min"],
     )
+    @pytest.mark.skip_for_json_result_set(reason="JSON format loses precision for Float.MAX_VALUE boundary values")
```
sfc-gh-asolarski: do we lose so much precision that this test doesn't make sense at all or could we introduce a bit softer assert in json case?

sfc-gh-pfus (commented):
Re: softer assert for float boundary test - unfortunately we can't use a relaxed assertion here. The issue isn't just precision loss, but that the value crosses from finite to infinite:

**What happens:**
- Expected: `1.7976931348623157e308` (Double.MAX_VALUE, finite)
- JSON from Snowflake: `"1.79769313486232e+308"` (15 significant digits)
- Parsed value: `inf` (infinity, not finite)

Python test:
```python
>>> val = 1.79769313486232e+308
>>> val
inf
>>> val == float('inf')
True
```

The test explicitly checks `Double.isFinite(actual)` which will always fail because the parsed JSON value is infinity. There's no tolerance range that would make `inf ≈ 1.797...e308` - they're categorically different (infinite vs finite).

This is why we skip the test for JSON format rather than trying to relax the assertion.

.github/workflows/test-python.yml
sfc-gh-pfus: Done.

python/tests/e2e/types/test_float.py
sfc-gh-pfus: Yes, it doesn't make sense, unfortunately.

sfc-gh-asolarski:
to have it documented

> Expected: 1.7976931348623157e308 (Double.MAX_VALUE, finite)
> JSON from Snowflake: "1.79769313486232e+308" (15 significant digits)
> Parsed value: inf (infinity, not finite)

Maybe we should rather have something like
```
if json_result_format():
    assert val == float('inf')
else:
    usual assert
```

jdbc/src/test/java/net/snowflake/client/SnowflakeIntegrationTestBase.java
```diff
@@ -75,6 +75,13 @@ protected Properties loadConnectionProperties() throws Exception {
 
   protected Connection openConnection() throws Exception {
     Properties props = loadConnectionProperties();
+
+    // Read QUERY_RESULT_FORMAT from environment
+    String resultFormat = System.getenv("QUERY_RESULT_FORMAT");
+    if (resultFormat != null && !resultFormat.isEmpty()) {
+      props.setProperty("PYTHON_CONNECTOR_QUERY_RESULT_FORMAT", resultFormat);
```
sfc-gh-asolarski, line 82: it sets `PYTHON_CONNECTOR_QUERY_RESULT_FORMAT` because that's the name of the property in core, right? 
but this property name is really misleading, at least we could leave a comment there explaining the reason

sfc-gh-asolarski: also applicable to the setup of the odbc wrapper

sfc-gh-asolarski (approved):
lgtm, just a small remark

jdbc/src/test/java/net/snowflake/client/SnowflakeIntegrationTestBase.java
sfc-gh-pfus: Yeah, this will need to be changed when all wrappers introduce themselves correctly.

PR #862, opened 2026-04-15
Python - add nativeokta tests

sfc-gh-pczajka:
During reference testing there was a [bug found in okta on-error behavior in reference driver](https://snowflakecomputing.atlassian.net/browse/SNOW-3388171).

### DIFF
--- a/python/tests/e2e/authentication/auth_helpers.py
+++ b/python/tests/e2e/authentication/auth_helpers.py
@@ -10,15 +10,17 @@ def verify_simple_query_execution(connection):
         assert result[0] == 1
 
 
-def verify_login_error(exception):
-    """Verify that an exception contains a valid login error with code and message."""
-    # Debug information about the exception can be seen when tests fail
+def verify_login_error(exception, keywords):
+    """Verify that an exception is a DatabaseError from an authentication failure.
+
+    Asserts that every keyword in *keywords* appears in the error message
+    (case-insensitive).
+    """
     assert exception is not None
     assert str(exception).strip() != "", "Login error message should not be empty"
 
     assert isinstance(exception.value, DatabaseError), f"Expected DatabaseError, got: {type(exception.value)}"
-    # Verify it's specifically an authentication/JWT error
+
     error_msg = str(exception.value).lower()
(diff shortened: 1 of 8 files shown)


python/tests/e2e/authentication/test_native_okta.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    params = get_test_parameters()
+    host = params.get("SNOWFLAKE_TEST_OKTA_HOST")
+    account = params.get("SNOWFLAKE_TEST_OKTA_ACCOUNT")
+    user = params.get("SNOWFLAKE_TEST_OKTA_USER")
+    password = params.get("SNOWFLAKE_TEST_OKTA_PASSWORD")
+    okta_url = params.get("SNOWFLAKE_TEST_OKTA_URL")
+
+    if not all([host, account, user, password, okta_url]):
+        pytest.skip(
```
sfc-gh-asolarski: should we skip or rather expect those to be present ?

sfc-gh-asolarski (approved):
lgtm in general

sfc-gh-asolarski:
but i;m afraid it turned out to be incompatible with the old driver

python/tests/e2e/authentication/auth_helpers.py
```diff
@@ -10,15 +10,17 @@ def verify_simple_query_execution(connection):
...
     assert exception is not None
     assert str(exception).strip() != "", "Login error message should not be empty"
 
     assert isinstance(exception.value, DatabaseError), f"Expected DatabaseError, got: {type(exception.value)}"
-    # Verify it's specifically an authentication/JWT error
+
     error_msg = str(exception.value).lower()
-    assert "jwt" in error_msg or "token" in error_msg or "invalid" in error_msg, (
-        f"Expected authentication-related error, got: {exception.value}"
-    )
```
sfc-gh-turbaszek, line 24: If we diff of sets then in single assert we can show all missing keys. Current implementation will fail with first missing key potentially hiding information about other missing keys. WDYT?

sfc-gh-pczajka: Nice, added

sfc-gh-turbaszek (approved):

PR #863, opened 2026-04-15
NO-SNOW: python - connection close releases handles

sfc-gh-asolarski:


sfc-gh-asolarski:
* **#863** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/863?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/863?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python/src/snowflake/connector/connection.py
```diff
@@ -192,8 +195,37 @@ def __init__(
     @pep249
     def close(self) -> None:
         """Close the connection now."""
+        if self._closed:
```
sfc-gh-fpawlowski: can we use .is_closed() as state should be determined by core and I would strive to not duplicating it in wrapper and in close future removing conn._closed bool at all

sfc-gh-fpawlowski (commented):
LGTM, good catch!

sfc-gh-fpawlowski (approved):
LGTM, good catch!

python/src/snowflake/connector/connection.py
sfc-gh-asolarski: good idea indeed, addressed

PR #868, opened 2026-04-16
Snow-2314136: logout config override on close

sfc-gh-fpawlowski:
## Summary

- Replace the non-atomic two-RPC pattern (`connection_set_option_int` then `connection_close`) with 6 optional override fields directly in `ConnectionCloseRequest`
- Core merges overrides into init-time config: close-request > init-time > defaults
- Python `close(retry=False)` now passes `max_attempts=1` atomically instead of mutating connection_seed

## Key design decisions

- **`max_attempts` is 1-based** (total attempts) everywhere — proto, Rust, Python. No 0-based retry count conversion
- **`error_strategy` is a string** in the proto (`"best_effort"` / `"strict"`), consistent with the existing `connection_set_option_string` pattern — not a proto enum
- **Shared `validate_positive_seconds`** validator used by both `from_settings()` (init-time) and `merge_with_request()` (close-time) to prevent validation drift

## Changed files

| File | Change |
|------|--------|
| `protobuf/database_driver_v1.proto` | 6 optional fields on `ConnectionCloseRequest` |
| `sf_core/src/config/logout.rs` | `merge_with_request()`, shared validator, 5 unit tests |
| `sf_core/src/apis/.../connection.rs` | Extended `connection_close()` signature + merge logic |
| `sf_core/src/protobuf/.../database_driver_v1.rs` | Bridge extracts overrides, parses error_strategy string |
| `python/.../connection.py` | Atomic `max_attempts=1` in close request |

## Test plan

- [x] 526 Rust unit tests pass (including 5 new `merge_with_request` tests)
- [x] `cargo clippy` + `cargo fmt` clean
- [x] Python precommit (ruff format + ruff lint + mypy) clean
- [x] Gherkin format validator passes
- [ ] Rust integration tests (WireMock-based)
- [ ] Python e2e tests against live server


protobuf/database_driver_v1.proto
sfc-gh-boler, line 311: What if an error happens and we can't tell whether the connection is closed or not?

For example, what is the expected value of this bool if close is called but logout is needs a refreshed token and is still pending?

sf_core/src/protobuf/apis/database_driver_v1/mod.rs
```diff
@@ -313,8 +250,30 @@ impl DatabaseDriver for DatabaseDriverImpl {
...
+            .map(|s| s.parse::<ErrorStrategy>())
+            .transpose()
+            .map_err(|e| DriverException {
+                message: format!("Invalid error_strategy: {e}"),
+                status_code: StatusCode::InvalidArgument as i32,
+                ..Default::default()
+            })?;
+
+        let overrides = CloseParamsOverrides {
+            server_session_keep_alive: input.server_session_keep_alive,
```
sfc-gh-boler, line 266: Instead of passing the bools, could we conjure an enum here representing the handful of real states that the core cares about? So the big comment in `should_send_logout` and big chunk of the body moves here?

That lets us deal with the "ugly" params early an and keep the rest of the core simpler.

sfc-gh-boler (commented):
(not a full review, just a drive by about the proto layer)

sfc-gh-boler (changes requested):

PR #869, opened 2026-04-16
Python: build release artifacts workflow

sfc-gh-pczajka:
Changes:
  1. New reusable workflow (`_build-python-wheels.yml`) — builds sf_core and Python
   wheels, shared between CI and release pipelines.
  2. Add release workflow (build-python-release.yml) — full matrix build (6      
  Python versions), tests wheels against all cloud providers (aws/gcp/azure) with
   and without pandas, builds and tests sdist, cleans up intermediate artifacts.
   currently only builds manylinux wheels
  3. Portable Linux wheels — static OpenSSL linking + auditwheel repair retags   
  wheels as manylinux for PyPI compatibility.                                    
  4. Smaller binaries — enabled thin LTO and single codegen unit in the Cargo
  release profile.


.github/workflows/_build-python-wheels.yml
```diff
@@ -0,0 +1,258 @@
...
+
+      - name: Cache proto generator target
+        uses: actions/cache@v5
+        with:
+          path: proto-target
+          key: ${{ runner.os }}-proto-target-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-proto-target-
+
+      - name: Build python wheel (Linux)
```
sfc-gh-turbaszek: For linux we should run `auditwheel` on the built wheel. 

Have you check if we can use industry standard of [cibuildwheel](https://cibuildwheel.pypa.io/en/stable/)?

sfc-gh-pczajka:
Moved to #918

PR #873, opened 2026-04-16
Add identifiable naming to test PATs for traceability

sfc-gh-makowalski:
## Summary

- PATs created during E2E tests now use the naming pattern `UD_{DRIVER}_{CI}_{BUILD}_{RANDOM}` (e.g. `UD_RUST_JNK_988_a3f2b1c0`) instead of the previous opaque `pat_{random}` format
- The naming encodes: driver origin (RUST/PYTHON/ODBC), CI system (BK=Buildkite, JNK=Jenkins, GHA=GitHub Actions, LOCAL=dev), build number, and a random suffix for uniqueness
- This allows orphaned PATs on the sfctest0 account to be immediately traced back to the specific build that created them

## Context

PATs are accumulating on the sfctest0 account because cleanup relies on in-process destructors (`Drop` in Rust, `finally` in Python, destructors in C++) which don't execute when builds are aborted/killed (e.g. the 16 consecutive timeout-aborted builds #965-#980). With identifiable names, we can determine which CI system and build left orphaned PATs.

This is the first of two PRs — the follow-up (#876) will add a CI-level cleanup step to remove stale PATs belonging to the current build.

## Changes

| File | Change |
|------|--------|
| `sf_core/tests/e2e/authentication/pat.rs` | PAT name: `UD_RUST_{ci}_{build}_{hex}` + `ci_build_tag()` helper |
| `python/tests/e2e/authentication/test_pat.py` | PAT name: `UD_PYTHON_{ci}_{build}_{hex}` + `_ci_build_tag()` helper |
| `odbc_tests/tests/e2e/authentication/pat.cpp` | PAT name: `UD_ODBC_{ci}_{build}_{hex}` + `ci_build_tag()` static method |

CI systems detected via environment variables:
- `BUILDKITE_BUILD_NUMBER` → `BK_{build}`
- `BUILD_NUMBER` → `JNK_{build}`
- `GITHUB_RUN_NUMBER` → `GHA_{build}`
- None set → `LOCAL_0`

Build numbers are sanitized to ASCII alphanumeric only.

## Test plan

- [ ] Verify PAT E2E tests still pass (naming change is transparent to Snowflake)
- [ ] Confirm PAT names in CI logs contain the expected `UD_*` prefix
- [ ] Check that cleanup (Drop/finally/destructor) still works with the new naming

PR #876, opened 2026-04-16
Add CI-level PAT cleanup for aborted/failed builds

sfc-gh-makowalski:
## Summary

When a CI build is aborted or killed by a timeout, in-process teardown (Rust `Drop`, Python `finally`, C++ destructors) never executes, leaving orphaned PATs on the `sfctest0` account. This PR adds a safety-net cleanup step that removes any PATs matching the current build's tag after tests complete (or fail). Builds on the identifiable PAT naming from #873.

### Changes

- **`scripts/cleanup_test_pats.py`** — new script that connects to Snowflake, lists PATs for the test user, and removes any matching the `UD_{DRIVER}_{BUILD_TAG}_{RANDOM}` pattern for the current build. Supports both password and key-pair auth. Auto-detects the build tag from `BUILDKITE_BUILD_NUMBER` / `BUILD_NUMBER` / `GITHUB_RUN_NUMBER` env vars.

- **`ci/Jenkinsfile`** — wraps VPN E2E test steps in `try/finally` so cleanup runs even when tests fail or the stage is aborted.

- **`ci/Jenkinsfile.coverage`** — adds cleanup after the test execution step in each coverage stage (Rust Core, Python, ODBC). Uses `returnStatus: true` so cleanup failure doesn't mask test results.

- **`.buildkite/pipelines/ci-tests/pipeline.yml`** — adds a dedicated `:broom: Cleanup test PATs` step that depends on all test groups with `allow_failure: true` and uses `soft_fail: true`, ensuring it always runs regardless of test outcomes.

- **`.github/workflows/test-rust-core.yml`**, **`test-python.yml`**, **`test-odbc.yml`** — adds a `Cleanup test PATs` step to each workflow that creates PATs, using `if: always()` + `continue-on-error: true` so the step runs after both pass and fail paths without ever blocking the overall build.

### How it works

The test code from #873 creates PATs with identifiable names like `UD_RUST_BK_1234_a3f2b1c0` (for Buildkite), `UD_PYTHON_GHA_555_deadbeef` (for GitHub Actions), `UD_ODBC_JNK_988_cafef00d` (for Jenkins), etc. The cleanup script matches this pattern — it finds all PATs whose name starts with `UD_` and contains the current build's tag, then removes them. This is purely a safety net; the normal test teardown still handles cleanup in the happy path.

## Test plan

- [ ] Verify Jenkins VPN E2E stage runs cleanup in `finally` block (visible in console logs)
- [ ] Verify Buildkite cleanup step runs after all test groups complete
- [ ] Verify each GHA workflow runs the cleanup step on both success and failure paths
- [ ] Verify cleanup script correctly identifies and removes matching PATs
- [ ] Verify cleanup failure does not fail the overall build

### DIFF
--- a/.buildkite/pipelines/ci-tests/pipeline.yml
+++ b/.buildkite/pipelines/ci-tests/pipeline.yml
@@ -333,8 +333,9 @@ steps:
         -e BUILDKITE_BUILD_NUMBER \
         "$$DOCKER_IMAGE" \
         bash -c '
+          set -euo pipefail
           ./scripts/decode_secrets.sh
-          pip3 install snowflake-connector-python -q
+          python3 -m pip install snowflake-connector-python -q
           python3 scripts/cleanup_test_pats.py \
             --parameter-path /workdir/parameters.json
         '
--- a/.buildkite/pipelines/ci-tests/pipeline.yml
+++ b/.buildkite/pipelines/ci-tests/pipeline.yml
@@ -300,3 +300,41 @@ steps:
             ' || DOCKER_EXIT=$$?
 
           ci/upload_test_results.sh jdbc "JDBC" "$$DOCKER_EXIT" "junit-results/jdbc-junit.xml"
+
+  # ==========================================================================
+  # PAT Cleanup (safety net for aborted/timed-out builds)
+  # ==========================================================================
+  - label: ":broom: Cleanup test PATs"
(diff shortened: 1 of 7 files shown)


sfc-gh-pcyrek (approved):
LGTM!

PR #890, opened 2026-04-18
perf: remove per-row heap allocations from DECIMAL/REAL/TIMESTAMP → CHAR paths

sfc-gh-makowalski:
## Summary

Three hot paths in the ODBC conversion layer — `SnowflakeNumber`, `SnowflakeReal`, and the shared `write_timestamp_to_odbc` helper — allocated one or two `String`s per row when converting values bound as `SQL_C_CHAR` / `SQL_C_WCHAR`. On bulk-fetch workloads (1M+ rows) this was the dominant cost of the conversion layer on Linux, where `glibc`'s allocator amplifies per-row alloc churn.

This PR replaces each with a **stack-buffered formatter that returns `Result<&str, WriteOdbcError>`** and produces byte-identical output for every in-range input. It also tightens two pre-existing correctness gaps in the SQL_C_NUMERIC re-scaling path.

Complements #881 (DATE/TIME + `mask_non_ascii_characters` → `Cow`). That PR eliminated the first heap allocation on the DATE/TIME paths; this one eliminates the equivalent allocations on the three remaining conversion hot paths.

### A. `SnowflakeNumber::format_decimal` → `format_decimal_into` (stack buffer)

The previous implementation had three separate allocation sources per row:

- `value.to_string()` → heap `String`
- `String::insert(0, '0')` called inside a `while` loop when `scale > digits.len()` → **O(scale²)** byte shifts
- `String::insert(decimal_pos, '.')` and `String::insert(0, '-')` → additional O(n) shifts per call

`format_decimal_into(value, scale, buf: &mut [u8; 48]) -> Result<&str, WriteOdbcError>` writes the absolute-value digits into a small scratch buffer via `write!`, then lays out `[sign][whole].[frac]` directly into a caller-provided 48-byte stack buffer.

### B. `SnowflakeReal` CHAR / WChar: `f64::to_string()` → `format_f64_display_into`

`f64::to_string()` allocates. `ryu` is ~6× faster but produces **different output** for integer-valued floats (`"1.0"` vs `"1"`) and for large/small magnitudes (`"1e20"` vs `"100000000000000000000"`, `"1e-10"` vs `"0.0000000001"`), which would be a user-visible behavioral change — so we deliberately don't use it. Instead, `format_f64_display_into(value, buf: &mut [u8; 384]) -> Result<&str, WriteOdbcError>` writes via the std `Display` impl into a 384-byte stack buffer (sized for the widest possible `f64` Display output — `1e308` prints as 309 digits because Rust's `Display` for `f64` uses full decimal notation, not scientific).

### C. `format_timestamp_string` → `format_timestamp_string_into` (stack buffer)

The fractional-seconds branch previously allocated twice per row (one `format!` for the 9-digit fraction, a second `format!` for the full string). The new `format_timestamp_string_into(dt, buf: &mut [u8; 48]) -> Result<&str, WriteOdbcError>` writes directly into a 48-byte caller buffer and trims trailing zeros in place. Preserves year padding, nanosecond trimming, and every other observable behavior of the old formatter.

### G. `10i128.pow(scale)` / `10u128.pow(scale)` → `pow10_i128` / `pow10_u128` (precomputed LUT)

Const `[i128; 39]` and `[u128; 39]` tables (`POW10_I128` / `POW10_U128`) replace the non-const `i128::pow` call that ran in the hot path, plus the two `10u128.pow` calls in the SQL_C_NUMERIC re-scaling arm. The helpers bounds-check via `.get()` and return `NumericValueOutOfRange` (SQLSTATE 22003) for `scale > MAX_DECIMAL_SCALE` (`= 38`). Modest absolute win on its own (~1 ms / 1M rows) but the hot loop is now branch-free and one less function call per row.

### Correctness fixes in the SQL_C_NUMERIC re-scaling arm

Two pre-existing bugs surfaced during review and are fixed in this PR:

1. **u128 overflow on upscale.** When `scale_diff > 0`, the arm multiplied `abs_value * pow10_u128(scale_diff)`. Both factors are user-influenced (`abs_value` is the server-supplied mantissa, `scale_diff = target_scale - self.scale` where `target_scale` comes from the application's ARD). Large combinations — e.g. `10²¹ × 10²⁰ = 10⁴¹` vs. `u128::MAX ≈ 3.4×10³⁸` — wrapped silently in release and wrote a corrupted `SQL_NUMERIC_STRUCT`. Now uses `checked_mul(...).ok_or_else(|| NumericValueOutOfRangeSnafu{...}.build())?` so the caller gets SQLSTATE 22003 on overflow (see test `numeric_upscale_overflow_returns_22003`).

2. **Non-typed errors for out-of-range `scale`.** Previously `10i128.pow(scale)` panicked in debug / overflowed in release for `scale > 38`. Now `pow10_i128` / `pow10_u128` return `NumericValueOutOfRange` cleanly.

## Behavior changes

| Input | Before this PR | After this PR |
|---|---|---|
| `self.scale > 38` in `write_odbc_type` | panic in debug, silent integer overflow in release | typed `NumericValueOutOfRange` (SQLSTATE 22003) |
| SQL_C_NUMERIC re-scale `\|scale_diff\| > 38` | panic in debug, silent overflow in release | typed `NumericValueOutOfRange` (SQLSTATE 22003) |
| SQL_C_NUMERIC upscale where `abs_value * 10^scale_diff` overflows u128 | silent wrap → corrupted `SQL_NUMERIC_STRUCT` | typed `NumericValueOutOfRange` (SQLSTATE 22003) |
| In-range input (every case the existing suite exercises) | — | byte-identical output |

All four changes are strict improvements: panic / silent corruption → typed SQL error. No public API change, no `Warnings` change, no SQLSTATE change on the success path.

## Microbench results

Measured on macOS release with 1M rows per iteration, averaged over 5 iterations. On Linux with `glibc` malloc the per-row allocation cost is typically 2–3× worse, so in-CI savings should exceed these numbers.

| Path | Old | New | Speedup | Saved / 1M rows |
|---|---|---|---|---|
| DECIMAL → CHAR, typical (scale = 6) | 46.2 ms | 11.1 ms | **4.2×** | 35 ms |
| DECIMAL → CHAR, pathological (v=1, scale=30) | 208.5 ms | 9.7 ms | **21.6×** | 199 ms |
| TIMESTAMP → CHAR, no fractional seconds | 174 ms | 98 ms | 1.78× | 76 ms |
| TIMESTAMP → CHAR, with fractional seconds | 257 ms | 115 ms | **2.24×** | 142 ms |
| REAL → CHAR (`f64::to_string`) | 82.9 ms | 44.5 ms | 1.86× | 38 ms |
| `10^scale` per row | 2.3 ms | 1.0 ms | 2.2× | 1.3 ms |

**Combined estimated savings** on a NUMBER-heavy 15-column fetch: ≈ 200–400 ms per 1M rows on Linux.

## Error handling design

All three helpers return `Result<&str, WriteOdbcError>`. They have only one kind of failure mode — buffer too small — which is surfaced as `NumericValueOutOfRange` for the caller to propagate. This matches the rest of the `write_odbc_type` contract (which already returns `Result<Warnings, WriteOdbcError>`), so each call site is just `let s = helper(...)?;`. No panics, no `debug_assert!` fallbacks, no `from_utf8_unchecked` on a potentially-truncated slice.

## Correctness

**983 odbc lib tests pass** (+20 new tests vs. main; 0 modified semantics for existing tests):

- **8 tests** for `format_decimal_into` covering every branch: `scale == 0`, `digits.len() > scale`, `digits.len() == scale`, `digits.len() < scale`, `i128::MIN` / `i128::MAX` at scale 0 and 10, large fractional values, `value == 0` at multiple scales, the pathological `(1, MAX_DECIMAL_SCALE)` case, and the new `scale > MAX_DECIMAL_SCALE` error return.
- **4 tests** for `format_timestamp_string_into` covering no-fractional-seconds, various trailing-zero counts, pre-epoch dates, year padding.
- **4 tests** for `format_f64_display_into` covering integer-valued floats, fractional values, very large/small magnitudes, `NaN` / `±inf`.
- **5 tests** for `pow10_i128` / `pow10_u128` covering in-range agreement with the table, out-of-range returning SQLSTATE 22003, and far-out-of-range (`u32::MAX`) not panicking.
- **1 test** `numeric_upscale_overflow_returns_22003` pinning the `checked_mul` behavior for SQL_C_NUMERIC.

Each test asserts against explicit expected output strings rather than a reference implementation, so the test cases document the exact wire shape the driver produces.

## Commits

| Hash | What |
|---|---|
| `bad27546f` | Initial perf rewrites (A, B, C, G) |
| `c3c601360` | Addresses Copilot 1st-round feedback — `pow10_*` helpers with SQLSTATE 22003 instead of panic |
| `6cf59f15c` | Addresses Copilot 2nd-round feedback — doc accuracy, `debug_assert!` on buffer writes |
| `eaf0257d9` | Addresses @sfc-gh-pbulawa 1st-round review — comment trimming, scale precondition, pow10 extract, tests rewritten to use explicit expected values instead of a reference impl |
| `87de73e11` | Addresses Copilot 3rd-round — SQL_C_NUMERIC upscale `checked_mul` (correctness) |
| `bdd4a5d27` | Addresses @sfc-gh-pbulawa 2nd-round — helpers now return `Result<&str, WriteOdbcError>` instead of panicking on any internal invariant violation |

## Test plan

- [x] `cargo test --lib` → 983 passed
- [x] `pre-commit run` → rustfmt + clippy + cargo check + tests-format-validator all clean
- [ ] Full `odbc_tests` suite in CI
- [ ] Perf dashboard run to confirm real-world numbers

### DIFF
--- a/odbc/src/conversion/number.rs
+++ b/odbc/src/conversion/number.rs
@@ -157,8 +157,9 @@ const POW10_U128: [u128; (MAX_DECIMAL_SCALE + 1) as usize] = {
 /// The previous `10i128.pow(scale)` implementation would panic in debug
 /// builds and silently overflow in release for `scale > 38`; this helper
 /// normalizes both to a typed conversion error. In practice Snowflake
-/// DECIMAL scale is always in `0..=37`, so the error branch is not
-/// expected on well-formed server output.
+/// DECIMAL scale is always in `0..=38` (the maximum precision is 38, and
+/// `0 ≤ scale ≤ precision`), so the error branch is not expected on
+/// well-formed server output.
 fn pow10_i128(scale: u32) -> Result<i128, WriteOdbcError> {
     POW10_I128.get(scale as usize).copied().ok_or_else(|| {
         NumericValueOutOfRangeSnafu {
@@ -203,8 +204,11 @@ impl SnowflakeNumber {
         let abs_len = {
             let mut cur = std::io::Cursor::new(&mut abs_tmp[..]);
             use std::io::Write as _;
-            // Infallible: abs_tmp is large enough for any u128.
-            let _ = write!(cur, "{}", value.unsigned_abs());
+            // 40 bytes always fits the decimal expansion of any u128
+            // (max is 39 digits). debug_assert guards against future code
+            // changes that might shrink the scratch buffer or feed a wider type.
+            let res = write!(cur, "{}", value.unsigned_abs());
+            debug_assert!(res.is_ok(), "abs_tmp[40] too small for u128 Display");
             cur.position() as usize
         };
         let digits = &abs_tmp[..abs_len];
--- a/odbc/src/conversion/number.rs
+++ b/odbc/src/conversion/number.rs
@@ -151,6 +151,39 @@ const POW10_U128: [u128; (MAX_DECIMAL_SCALE + 1) as usize] = {
     a
 };
 
+/// Fetch `10^scale` as `i128` from the precomputed table, returning an
+/// `SQLSTATE 22003` error instead of panicking for `scale > 38`.
+///
+/// The previous `10i128.pow(scale)` implementation would panic in debug
+/// builds and silently overflow in release for `scale > 38`; this helper
+/// normalizes both to a typed conversion error. In practice Snowflake
+/// DECIMAL scale is always in `0..=37`, so the error branch is not
+/// expected on well-formed server output.
+fn pow10_i128(scale: u32) -> Result<i128, WriteOdbcError> {
+    POW10_I128.get(scale as usize).copied().ok_or_else(|| {
+        NumericValueOutOfRangeSnafu {
+            reason: format!(
+                "DECIMAL scale {scale} exceeds supported range 0..={MAX_DECIMAL_SCALE}"
+            ),
+        }
+        .build()
+    })
+}
+
+/// `u128` counterpart to [`pow10_i128`], used for re-scaling in the
+/// `SQL_C_NUMERIC` arm where the scale difference comes from the
+/// application-supplied ARD and is not otherwise validated.
+fn pow10_u128(scale: u32) -> Result<u128, WriteOdbcError> {
+    POW10_U128.get(scale as usize).copied().ok_or_else(|| {
+        NumericValueOutOfRangeSnafu {
+            reason: format!(
+                "scale {scale} exceeds supported range 0..={MAX_DECIMAL_SCALE} for SQL_C_NUMERIC"
+            ),
+        }
+        .build()
+    })
+}
+
 impl SnowflakeNumber {
     /// Format a scaled `i128` as a decimal string into `buf` without any heap
     /// allocation, returning the filled slice as `&str`.
@@ -240,7 +273,7 @@ impl WriteODBCType for SnowflakeNumber {
             other => other,
         };
 
-        let scale_factor = POW10_I128[self.scale as usize];
+        let scale_factor = pow10_i128(self.scale)?;
         let int_value = snowflake_value / scale_factor;
         let has_fractional = self.scale > 0 && snowflake_value % scale_factor != 0;
 
@@ -368,15 +401,15 @@ impl WriteODBCType for SnowflakeNumber {
 
                 let scale_diff = target_scale as i32 - self.scale as i32;
                 let truncated = if scale_diff < 0 {
-                    let divisor = POW10_U128[(-scale_diff) as usize];
+                    let divisor = pow10_u128((-scale_diff) as u32)?;
                     abs_value % divisor != 0
                 } else {
                     false
                 };
                 let unscaled: u128 = if scale_diff >= 0 {
-                    abs_value * POW10_U128[scale_diff as usize]
+                    abs_value * pow10_u128(scale_diff as u32)?
                 } else {
-                    abs_value / POW10_U128[(-scale_diff) as usize]
+                    abs_value / pow10_u128((-scale_diff) as u32)?
                 };
 
                 let numeric = sql::Numeric {
@@ -680,5 +713,59 @@ mod format_decimal_into_tests {
             assert_matches(1234567890, scale);
             assert_matches(-1234567890, scale);
         }
+    }
+}
+
+#[cfg(test)]
+mod pow10_tests {
+    use super::{MAX_DECIMAL_SCALE, POW10_I128, POW10_U128, pow10_i128, pow10_u128};
+    use crate::conversion::error::WriteOdbcError;
+
+    #[test]
+    fn pow10_i128_matches_table_in_range() {
+        for scale in 0..=MAX_DECIMAL_SCALE {
+            assert_eq!(
+                pow10_i128(scale).unwrap(),
+                POW10_I128[scale as usize],
+                "pow10_i128({scale}) disagrees with table"
+            );
+        }
+    }
+
+    #[test]
+    fn pow10_u128_matches_table_in_range() {
+        for scale in 0..=MAX_DECIMAL_SCALE {
+            assert_eq!(
+                pow10_u128(scale).unwrap(),
+                POW10_U128[scale as usize],
+                "pow10_u128({scale}) disagrees with table"
+            );
+        }
+    }
+
+    #[test]
+    fn pow10_i128_out_of_range_returns_22003() {
+        let err = pow10_i128(MAX_DECIMAL_SCALE + 1).unwrap_err();
+        assert!(
+            matches!(err, WriteOdbcError::NumericValueOutOfRange { .. }),
+            "expected NumericValueOutOfRange, got {err:?}"
+        );
+    }
+
+    #[test]
+    fn pow10_u128_out_of_range_returns_22003() {
+        let err = pow10_u128(MAX_DECIMAL_SCALE + 1).unwrap_err();
+        assert!(
+            matches!(err, WriteOdbcError::NumericValueOutOfRange { .. }),
+            "expected NumericValueOutOfRange, got {err:?}"
+        );
+    }
+
+    #[test]
+    fn pow10_far_out_of_range_does_not_panic() {
+        // The previous `10i128.pow(n)` would panic in debug for n > 38.
+        // The new helpers return a typed error instead.
+        assert!(pow10_i128(100).is_err());
+        assert!(pow10_u128(u32::MAX).is_err());
     }
 }
--- a/odbc/src/conversion/number.rs
+++ b/odbc/src/conversion/number.rs
@@ -647,7 +647,7 @@ impl WriteJson for SnowflakeNumber {
 
 #[cfg(test)]
 mod format_decimal_into_tests {
-    use super::{SnowflakeNumber, MAX_DECIMAL_SCALE};
+    use super::{MAX_DECIMAL_SCALE, SnowflakeNumber};
     use crate::conversion::error::WriteOdbcError;
 
     fn fmt(value: i128, scale: u32) -> String {
@@ -762,7 +762,7 @@ mod format_decimal_into_tests {
 
 #[cfg(test)]
 mod pow10_tests {
-    use super::{pow10_i128, pow10_u128, MAX_DECIMAL_SCALE, POW10_I128, POW10_U128};
+    use super::{MAX_DECIMAL_SCALE, POW10_I128, POW10_U128, pow10_i128, pow10_u128};
     use crate::conversion::error::WriteOdbcError;
 
     #[test]
--- a/odbc/src/conversion/number.rs
+++ b/odbc/src/conversion/number.rs
@@ -186,20 +186,31 @@ impl SnowflakeNumber {
     /// Format a scaled `i128` as a decimal string into `buf` without any heap
     /// allocation, returning the filled slice as `&str`.
     ///
-    /// `buf` must be large enough for the widest possible output: optional
-    /// `-`, up to 39 digits of whole part, a `.`, and up to `MAX_DECIMAL_SCALE`
-    /// fractional digits. 48 bytes is sufficient. `scale` is asserted to be in
-    /// `0..=MAX_DECIMAL_SCALE`.
+    /// Preconditions validated at runtime:
+    /// - `scale <= MAX_DECIMAL_SCALE` — otherwise returns
+    ///   `NumericValueOutOfRange` (a scale of 39+ would overflow the 48-byte
+    ///   output buffer).
+    /// - The scratch buffer for absolute-value digits is 40 bytes (fits the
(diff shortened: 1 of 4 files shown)


odbc/src/conversion/number.rs
```diff
@@ -123,26 +123,133 @@ where
...
+    /// Format a scaled `i128` as a decimal string into `buf` without any heap
+    /// allocation, returning the filled slice as `&str`.
+    ///
+    /// `buf` must be large enough for the widest possible output:
+    /// optional `-`, up to 39 digits of whole part, a `.`, and up to 38
+    /// fractional digits. 48 bytes is sufficient.
+    ///
+    /// The output is byte-identical to the previous
+    /// `format!`/`String::insert`-based implementation, including the shape
+    /// `"0.000...digits"` when `scale > digits`.
```
sfc-gh-pbulawa: Remove the comment

odbc/src/conversion/number.rs
```diff
@@ -123,26 +123,133 @@ where
...
+
+/// Fetch `10^scale` as `i128` from the precomputed table, returning an
+/// `SQLSTATE 22003` error instead of panicking for `scale > 38`.
+///
+/// The previous `10i128.pow(scale)` implementation would panic in debug
+/// builds and silently overflow in release for `scale > 38`; this helper
+/// normalizes both to a typed conversion error. In practice Snowflake
+/// DECIMAL scale is always in `0..=38` (the maximum precision is 38, and
+/// `0 ≤ scale ≤ precision`), so the error branch is not expected on
+/// well-formed server output.
```
sfc-gh-pbulawa: Remove the comment

odbc/src/conversion/number.rs
```diff
@@ -123,26 +123,133 @@ where
...
+            buf[len] = b'.';
+            len += 1;
+            buf[len..len + scale].copy_from_slice(&digits[int_part..]);
+            len += scale;
         } else {
-            value.to_string()
+            // "0." + (scale - digits.len()) zero-pad + digits.
+            // Matches the original implementation which wrote a leading "0"
+            // whenever `scale >= digits.len()` (the `while s.len() <= scale`
+            // loop grew `s` until the decimal insert position was zero).
```
sfc-gh-pbulawa: Rephrase/remove the comment as it refers to the original implementation.

odbc/src/conversion/number.rs
```diff
@@ -514,3 +623,153 @@ impl WriteJson for SnowflakeNumber {
...
     }
 }
+
+#[cfg(test)]
+mod format_decimal_into_tests {
+    use super::{MAX_DECIMAL_SCALE, SnowflakeNumber};
+
+    /// Reference implementation of the previous `format_decimal` — kept
+    /// inside the test module so that we can assert byte-identical output
+    /// for the new stack-buffer implementation.
```
sfc-gh-pbulawa: Remove the information about the reference implementation

odbc/src/conversion/number.rs
```diff
@@ -514,3 +623,153 @@ impl WriteJson for SnowflakeNumber {
...
+    fn format_decimal_reference(value: i128, scale: u32) -> String {
```
sfc-gh-pbulawa: I don't think we should keep the reference/old implementation here. Focus on the actual test cases implementations.

odbc/src/conversion/number.rs
```diff
@@ -123,26 +123,133 @@ where
...
+    /// allocation, returning the filled slice as `&str`.
+    ///
+    /// `buf` must be large enough for the widest possible output:
+    /// optional `-`, up to 39 digits of whole part, a `.`, and up to 38
+    /// fractional digits. 48 bytes is sufficient.
+    ///
+    /// The output is byte-identical to the previous
+    /// `format!`/`String::insert`-based implementation, including the shape
+    /// `"0.000...digits"` when `scale > digits`.
+    fn format_decimal_into(value: i128, scale: u32, buf: &mut [u8; 48]) -> &str {
```
sfc-gh-pbulawa: Verify the scale here, as it's done implicitly right now.

odbc/src/conversion/timestamp.rs
```diff
@@ -223,32 +225,50 @@ fn read_struct_timestamp_tz(
...
-            dt.year(),
-            dt.month(),
-            dt.day(),
-            dt.hour(),
-            dt.minute(),
-            dt.second(),
-            trimmed
-        )
+        );
+        debug_assert!(date_res.is_ok(), "timestamp buf too small for date+time");
```
sfc-gh-pbulawa: Don't use `debug_assert`; Replace them with actual assertions that throw the error.

odbc/src/conversion/timestamp.rs
```diff
@@ -223,32 +225,50 @@ fn read_struct_timestamp_tz(
...
+        if nanos != 0 {
+            let frac_res = write!(cur, ".{nanos:09}");
+            debug_assert!(frac_res.is_ok(), "timestamp buf too small for fraction");
```
sfc-gh-pbulawa: Don't use `debug_assert`; Replace them with actual assertions that throw the error.

odbc/src/conversion/real.rs
```diff
@@ -48,6 +50,33 @@ impl ReadArrowType<Float64Array> for SnowflakeReal {
...
+    let len = {
+        let mut cur = Cursor::new(&mut buf[..]);
+        // 384 bytes is comfortably larger than the widest output Rust's
+        // `<f64 as Display>::fmt` produces today (~325 bytes for `1e-300`,
+        // `1e308`, etc., since Display uses non-scientific notation for
+        // all finite values). The debug_assert guards against a future
+        // libcore change widening the format, which would otherwise be
+        // silently truncated.
+        let res = write!(cur, "{value}");
+        debug_assert!(res.is_ok(), "f64 buf[384] too small for Display output");
```
sfc-gh-pbulawa: Don't use `debug_assert`; Replace them with actual assertions that throw the error.

odbc/src/conversion/number.rs
```diff
@@ -123,26 +123,133 @@ where
...
+        // directly into the provided buffer without heap allocation).
+        let mut abs_tmp = [0u8; 40]; // 39 digits for u128::MAX + headroom
+        let abs_len = {
+            let mut cur = std::io::Cursor::new(&mut abs_tmp[..]);
+            use std::io::Write as _;
+            // 40 bytes always fits the decimal expansion of any u128
+            // (max is 39 digits). debug_assert guards against future code
+            // changes that might shrink the scratch buffer or feed a wider type.
+            let res = write!(cur, "{}", value.unsigned_abs());
+            debug_assert!(res.is_ok(), "abs_tmp[40] too small for u128 Display");
```
sfc-gh-pbulawa: Don't use `debug_assert`; Replace them with actual assertions that throw the error.

odbc/src/conversion/number.rs
```diff
@@ -296,15 +405,15 @@ impl WriteODBCType for SnowflakeNumber {
 
                 let scale_diff = target_scale as i32 - self.scale as i32;
                 let truncated = if scale_diff < 0 {
-                    let divisor = 10u128.pow((-scale_diff) as u32);
+                    let divisor = pow10_u128((-scale_diff) as u32)?;
```
sfc-gh-pbulawa: It seems that it's worth to extract `pow10_u128((-scale_diff) as u32)` into a separate variable, so it won't be computed twice

sfc-gh-pbulawa (changes requested):
PTAL at the comments

odbc/src/conversion/number.rs
```diff
@@ -123,26 +123,133 @@ where
...
+    /// `-`, up to 39 digits of whole part, a `.`, and up to `MAX_DECIMAL_SCALE`
+    /// fractional digits. 48 bytes is sufficient. `scale` is asserted to be in
+    /// `0..=MAX_DECIMAL_SCALE`.
+    ///
+    /// Output shape:
+    /// - `scale == 0`              → `[-]<digits>`
+    /// - `digits.len() > scale`    → `[-]<int>.<frac>`
+    /// - `digits.len() <= scale`   → `[-]0.<zero-pad><digits>`
+    fn format_decimal_into(value: i128, scale: u32, buf: &mut [u8; 48]) -> &str {
+        assert!(
```
sfc-gh-pbulawa: Don't use assert, return error instead.

odbc/src/conversion/real.rs
```diff
@@ -48,6 +50,31 @@ impl ReadArrowType<Float64Array> for SnowflakeReal {
...
+/// bytes for headroom.
+fn format_f64_display_into(value: f64, buf: &mut [u8; 384]) -> &str {
+    let len = {
+        let mut cur = Cursor::new(&mut buf[..]);
+        // 384 bytes is comfortably larger than the widest output Rust's
+        // `<f64 as Display>::fmt` produces (~325 bytes for `1e-300`, `1e308`
+        // — Display uses non-scientific notation for all finite values).
+        // The .expect() guards against a future libcore change widening the
+        // format, which would otherwise be silently truncated.
+        write!(cur, "{value}").expect("f64 buf[384] too small for Display output");
```
sfc-gh-pbulawa: Return error here, as the expect will panic.

odbc/src/conversion/timestamp.rs
```diff
@@ -235,20 +251,23 @@ fn format_timestamp_string(dt: &NaiveDateTime) -> String {
...
-            dt.month(),
-            dt.day(),
-            dt.hour(),
-            dt.minute(),
-            dt.second(),
-            trimmed
-        )
+        .expect("timestamp buf too small for date+time");
+        if nanos != 0 {
+            write!(cur, ".{nanos:09}").expect("timestamp buf too small for fraction");
```
sfc-gh-pbulawa: Return error instead of expect, as this will panic.

sfc-gh-pbulawa (changes requested):

odbc/src/conversion/number.rs
```diff
@@ -514,3 +644,171 @@ impl WriteJson for SnowflakeNumber {
...
 }
+
+#[cfg(test)]
+mod format_decimal_into_tests {
+    use super::{MAX_DECIMAL_SCALE, SnowflakeNumber};
+
+    fn fmt(value: i128, scale: u32) -> String {
+        let mut buf = [0u8; 48];
+        SnowflakeNumber::format_decimal_into(value, scale, &mut buf)
+            .expect("unexpected format error for in-range scale")
```
sfc-gh-pbulawa: This will panic. Replace with an error.

sfc-gh-pbulawa (changes requested):

sfc-gh-pbulawa (approved):

PR #889, opened 2026-04-17
SNOW-3396010: Add env-based platform detectors

sfc-gh-rsavenok:
## Summary
- Added 5 env-only platform detectors (`is_aws_lambda`, `is_azure_function`, `is_gce_cloud_run_service`, `is_gce_cloud_run_job`, `is_github_action`) driven by a stable `ENV_DETECTORS` table in `sf_core/src/telemetry/platform_detection.rs`.
- `SNOWFLAKE_DISABLE_PLATFORM_DETECTION` now accepts `"true"` (case-insensitive) or literal `"1"` as truthy disable values.
- Introduced `platform_detection_env_vars(overrides)` (gated behind the existing `test-utils` feature) so unit and integration tests can clear every detection env key by default and override only what each test cares about. Prevents CI-runner env leakage (e.g., `GITHUB_ACTIONS=true` on the host).
- Integration test gained an end-to-end `should_detect_aws_lambda` scenario that asserts `CLIENT_ENVIRONMENT.PLATFORM == ["is_aws_lambda"]` in the login-request body; existing two scenarios migrated onto the new helper.
- Feature file updated: new `@core_int` scenario "should detect AWS Lambda" and tightened empty-detection assertion to `PLATFORM == []` (matches the actual wire format).

## Context
Builds on the platform-detection foundation (`rsavenok/platform-detection-foundation`). This PR lands the env-only bucket from the reference draft (PR #885). HTTP (IMDS) and AWS STS detectors are intentionally out of scope and will follow in a separate PR to keep this one reviewable.

## Test plan
- `cargo test -p sf_core --lib telemetry` — 13 unit tests pass (every detector, case-insensitive disable flag + `"1"`, partial-Azure rejection, empty-value rejection, multi-detection ordering, disable-wins precedence).
- `cargo test -p sf_core --test integration_tests telemetry::platform_detection` — 3 integration tests pass, including the new `should_detect_aws_lambda` scenario.
- `cargo test -p sf_core --lib apis::database_driver_v1` — 125 pass (driver-level `platforms()` cache + connection tests unaffected).
- `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean (pre-commit hooks also ran and passed).
- `tests/tests_format_validator` validates the updated `platform_detection.feature` against the Rust integration test file.

PR #888, opened 2026-04-17
SNOW-3396010: Platform detection foundation

sfc-gh-rsavenok:
## Summary
- Add `PLATFORM` field to `CLIENT_ENVIRONMENT` in the Snowflake login request (`AuthRequestClientEnvironment` in `sf_core/src/rest/snowflake/auth.rs`).
- Always serialize the `PLATFORM` array — including when empty (`[]`) — so the server can reliably distinguish "detection ran, no platforms" from "detection not supported". When `SNOWFLAKE_DISABLE_PLATFORM_DETECTION=true`, the array is populated with `["disabled"]`.
- Wire platform detection results from global state into the login request payload.
- Add integration tests under `sf_core/tests/integration/telemetry/platform_detection.rs` and the corresponding Gherkin scenarios in `tests/definitions/core/telemetry/platform_detection.feature` covering both the disabled-via-env-var and empty-detection cases.

## Context
This PR is the **foundation for platform detection** telemetry in the core driver. It establishes:

1. The wire-format contract with Snowflake for reporting client runtime platforms in the login request.
2. The plumbing from `GlobalState::platforms()` / `detect_platforms()` into the login flow.
3. The env-var escape hatch (`SNOWFLAKE_DISABLE_PLATFORM_DETECTION`) that disables detection while still reporting that state to the server.
4. An integration test harness (mock login server + assertions on the login-request body) that follow-up PRs can reuse to validate detection of specific platforms (containers, cloud runtimes, CI environments, etc.).

Actual platform detectors will be added on top of this foundation in subsequent PRs.

## Test plan
- Ran the new integration tests:
  ```
  cargo test -p sf_core --test integration_tests telemetry::platform_detection
  ```
  Both scenarios pass:
  - `should_send_platform_disabled_when_detection_is_disabled_via_env_var` — verifies `PLATFORM == ["disabled"]` when the env var is set.
  - `should_send_empty_platform_array_when_detection_produces_no_platforms` — verifies `PLATFORM == []` when detection runs but returns nothing.
- Assertions inspect the captured wiremock login-request body to confirm `CLIENT_ENVIRONMENT.PLATFORM` is always present and correctly shaped.

PR #887, opened 2026-04-17
Introduce test-utils and reusable ClientInfo for tests

sfc-gh-rsavenok:
## Summary

Consolidates 7 duplicated `ClientInfo { ... }` test fixtures into a single shared `test_client_info()` helper colocated with the struct definition in `sf_core/src/config/rest_parameters.rs`.

## Changes

- Added `test_fixtures::test_client_info()` gated on `#[cfg(any(test, feature = "test-utils"))]`.
- Added `test-utils` cargo feature (off by default) + self-referential dev-dep so `tests/**` can access it.
- Replaced 7 call sites across `src/rest/snowflake/{mod,heartbeat}.rs`, `src/crl/integration_test.rs`, `tests/integration/{session,http}/session_refresh.rs`, `tests/integration/http/sync_retry.rs`, and `tests/e2e/session/session_refresh.rs`. Sites with value-specific assertions use struct-update syntax (`..test_client_info()`) to override only the fields they care about.
- Out of scope: `src/bin/collect_chunk_data.rs` (production code) and the `python/sf_core/**` parallel tree.

## Test plan

- [x] `cargo check -p sf_core` — fixture doesn't leak into non-test builds.
- [x] `cargo test -p sf_core --lib` — all 475 lib tests pass.
- [x] `cargo test -p sf_core --test integration_tests -- session::session_refresh http::session_refresh http::sync_retry` — all 7 affected integration tests pass.

PR #891, opened 2026-04-18
SNOW-3396010: has_aws_identity detector

sfc-gh-rsavenok:
## Summary
- Add `has_aws_identity` platform detector that calls STS `GetCallerIdentity` and accepts only IAM user / STS assumed-role ARNs suitable for workload identity federation.
- Introduce a `CallerIdentityProvider` trait and a `DetectionConfig` so detectors can be injected with fakes in tests and never hit real AWS.
- Run all detectors concurrently with a 200 ms per-detector timeout so a slow or unreachable STS endpoint cannot delay connection setup.
- Change `detect_platforms()` to `detect_platforms(&DetectionConfig)` and update the global-state caller to pass `DetectionConfig::default()`.
- Add `aws-sdk-sts = "1.86.0"` and move `tokio` into `[dev-dependencies]` with `test-util` to support paused-time tests.

## Context
Builds on `rsavenok/platform-detection-via-env` (#889), which added the env-based detector framework. This PR is the next step in SNOW-3396010: extending platform detection beyond environment variables to active identity probes, starting with AWS. The `CallerIdentityProvider` abstraction is designed so additional cloud-identity probes (e.g. GCP, Azure) can be plugged in the same way without changing `detect_platforms`.

## Test plan
- `cargo test -p sf_core telemetry::platform_detection` and `cargo test -p sf_core telemetry::aws_identity` — all pass locally.
- New tests cover:
  - `detects_has_aws_identity_when_provider_returns_wif_arn`
  - `rejects_when_sts_returns_no_arn`
  - `drops_when_provider_exceeds_timeout` (uses `#[tokio::test(start_paused = true)]` + `tokio::time::advance` to assert the 200 ms per-detector timeout fires)
  - `accepts_iam_user_and_assumed_role` and `rejects_malformed_arns` for `is_valid_arn_for_wif`.
- Existing env-detector tests were updated to use a fake provider that returns `None`, so the test suite never reaches the network.

### DIFF
--- a/sf_core/src/apis/database_driver_v1/global_state.rs
+++ b/sf_core/src/apis/database_driver_v1/global_state.rs
@@ -9,7 +9,7 @@ use super::statement::Statement;
 use crate::fs_adapter::{FsAdapter, RealFs};
 use crate::handle_manager::HandleManager;
 use crate::telemetry::os_details::detect_os_details;
-use crate::telemetry::platform_detection::detect_platforms;
+use crate::telemetry::platform_detection::{DetectionConfig, detect_platforms};
 use crate::token_cache::{KeyringTokenCache, TokenCacheError};
 
 /// Injection points for `DatabaseDriverV1`.
@@ -65,7 +65,9 @@ impl DatabaseDriverV1 {
     }
 
     pub async fn platforms(&self) -> &Vec<String> {
-        self.platforms.get_or_init(detect_platforms).await
+        self.platforms
+            .get_or_init(|| async { detect_platforms(&DetectionConfig::default()).await })
+            .await
     }
 
     pub fn os_details(&self) -> &Option<HashMap<String, String>> {
--- a/sf_core/src/telemetry/platform_detection.rs
+++ b/sf_core/src/telemetry/platform_detection.rs
@@ -1,14 +1,31 @@
-type EnvDetector = (&'static str, fn() -> bool);
+use std::sync::Arc;
+use std::time::Duration;
 
-const ENV_DETECTORS: &[EnvDetector] = &[
-    ("is_aws_lambda", is_aws_lambda),
-    ("is_azure_function", is_azure_function),
-    ("is_gce_cloud_run_service", is_gce_cloud_run_service),
-    ("is_gce_cloud_run_job", is_gce_cloud_run_job),
-    ("is_github_action", is_github_action),
-];
+use futures::FutureExt;
+use futures::future::BoxFuture;
 
-pub async fn detect_platforms() -> Vec<String> {
+use super::aws_identity::{CallerIdentityProvider, StsCallerIdentityProvider, has_aws_identity};
+
+const DETECTION_TIMEOUT: Duration = Duration::from_millis(200);
+
+#[derive(Clone)]
+pub struct DetectionConfig {
+    pub(crate) caller_identity_provider: Arc<dyn CallerIdentityProvider>,
+}
+
+impl Default for DetectionConfig {
+    fn default() -> Self {
+        Self {
+            caller_identity_provider: Arc::new(StsCallerIdentityProvider),
+        }
+    }
+}
+
+/// Run all detectors concurrently with a per-detector [`DETECTION_TIMEOUT`].
+///
+/// Returns the names of only those detectors that succeeded. Order matches
+/// the detector list below so the serialized array is stable across runs.
+pub async fn detect_platforms(config: &DetectionConfig) -> Vec<String> {
     if std::env::var("SNOWFLAKE_DISABLE_PLATFORM_DETECTION")
         .map(|value| value.eq_ignore_ascii_case("true") || value == "1")
         .unwrap_or(false)
@@ -16,10 +33,36 @@ pub async fn detect_platforms() -> Vec<String> {
         return vec!["disabled".to_string()];
     }
 
-    ENV_DETECTORS
-        .iter()
-        .filter(|(_, probe)| probe())
-        .map(|(name, _)| (*name).to_string())
+    let detectors: Vec<(&'static str, BoxFuture<'_, bool>)> = vec![
+        ("is_aws_lambda", async { is_aws_lambda() }.boxed()),
+        ("is_azure_function", async { is_azure_function() }.boxed()),
+        (
+            "is_gce_cloud_run_service",
+            async { is_gce_cloud_run_service() }.boxed(),
+        ),
+        (
+            "is_gce_cloud_run_job",
+            async { is_gce_cloud_run_job() }.boxed(),
+        ),
+        ("is_github_action", async { is_github_action() }.boxed()),
+        (
+            "has_aws_identity",
+            has_aws_identity(config.caller_identity_provider.as_ref()).boxed(),
+        ),
+    ];
+
+    let results = futures::future::join_all(detectors.into_iter().map(|(name, fut)| async move {
+        let detected = tokio::time::timeout(DETECTION_TIMEOUT, fut)
+            .await
+            .unwrap_or(false);
+        (name, detected)
+    }))
+    .await;
+
+    results
+        .into_iter()
+        .filter(|(_, detected)| *detected)
+        .map(|(name, _)| name.to_string())
         .collect()
 }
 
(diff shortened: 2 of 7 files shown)


sfc-gh-rsavenok:
* **#893** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/893?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#892** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/892?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#891** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/891?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/891?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#889** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/889?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#888** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/888?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#887** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/887?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sf_core/src/apis/database_driver_v1/global_state.rs
```diff
@@ -32,7 +32,9 @@ impl DatabaseDriverV1 {
     }
 
     pub async fn platforms(&self) -> &Vec<String> {
-        self.platforms.get_or_init(detect_platforms).await
+        self.platforms
+            .get_or_init(|| async { detect_platforms(&DetectionConfig::default()).await })
```
sfc-gh-pbulawa, line 69: As the `DetectionConfig::default` is hardcoded here, we won't be able to inject a fake provider, right?

The test helper does not clear `AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN, AWS_PROFILE, AWS_CONFIG_FILE, AWS_SHARED_CREDENTIALS_FILE, AWS_CONTAINER_CREDENTIALS_RELATIVE_URI, or AWS_WEB_IDENTITY_TOKEN_FILE`. And even if it did, `aws_config::defaults().load()` also resolves credentials from `~/.aws/credentials` and IMDS, neither of which are controlled by `temp_env`.

sfc-gh-pbulawa (changes requested):

sf_core/src/telemetry/platform_detection.rs
```diff
@@ -1,25 +1,68 @@
...
+        }
+    }
+}
 
-pub async fn detect_platforms() -> Vec<String> {
+/// Run all detectors concurrently with a per-detector [`DETECTION_TIMEOUT`].
+///
+/// Returns the names of only those detectors that succeeded. Order matches
+/// the detector list below so the serialized array is stable across runs.
+pub async fn detect_platforms(config: &DetectionConfig) -> Vec<String> {
```
sfc-gh-jszczerbinski, line 28: This should return vec of enum values:
- Also consider implementing detect method: `def detect(self, detectionConfig)`

sfc-gh-jszczerbinski (changes requested):

sf_core/src/apis/database_driver_v1/global_state.rs
sfc-gh-rsavenok: > As the DetectionConfig::default is hardcoded here, we won't be able to inject a fake provider, right?

We don't really need to test on this level. Provider is injected into the unit tests of platform detection:
```
let mut cfg = test_detection_config();
cfg.caller_identity_provider = Arc::new(FakeCallerIdentityProvider::new(None));
....
detect_platforms(&cfg).await
```

> The test helper does not clear  ....

Why does it need to clear them? It doesn't set them

> And even if it did, aws_config::defaults().load() also resolves credentials from ~/.aws/credentials and IMDS, neither of which are controlled by temp_env.

This is a good point. We have 2 integration tests which could accidentally detect cloud vm, but instead of complicating code with full dependency injection, I think it would be easier to change assertions

sfc-gh-rsavenok: Adjusted integration tests - https://github.com/snowflakedb/universal-driver/pull/891/changes/4f45898269dc06e3bcf8e7dff6d0cd9e4b974769

Note that we now don't have a coverage for `[]`, so `skip_serializing_if = "Option::is_none"` won't be caught by tests in
```
 #[serde(rename = "PLATFORM")]
pub platforms: Vec<String>,
```

But I think it's overkill to add complex dependency injection to driverv1 just to be able to test this minor edge case

sf_core/src/telemetry/platform_detection.rs
sfc-gh-rsavenok: > This should return vec of enum values

What benefit do they give us? The value in code is used only once - in the detectors vector + tests. Seems like overkill to define an enum + string serialization for something that doesn't repeat outside of this file + 1 integration test

> Also consider implementing detect method:

Do you mean something like this? Seems like much more complex code than we have now ¯\_(ツ)_/¯
```
/// Identifiers for every platform the driver can report.
///
/// The `as_str` form is what gets serialized to the wire; keep it in sync
/// with the server-side expectations.
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
pub enum Platform {
    IsAwsLambda,
    IsAzureFunction,
    IsGceCloudRunService,
    IsGceCloudRunJob,
    IsGithubAction,
    HasAwsIdentity,
}
impl Platform {
    pub fn as_str(self) -> &'static str {
        match self {
            Platform::IsAwsLambda => "is_aws_lambda",
            Platform::IsAzureFunction => "is_azure_function",
            Platform::IsGceCloudRunService => "is_gce_cloud_run_service",
            Platform::IsGceCloudRunJob => "is_gce_cloud_run_job",
            Platform::IsGithubAction => "is_github_action",
            Platform::HasAwsIdentity => "has_aws_identity",
        }
    }
}
/// Outcome of [`detect_platforms`].
///
/// Distinguishes "detection ran and found nothing" from "detection was
/// disabled by the kill switch" — both were previously squashed into the
/// same `Vec<String>`.
#[derive(Debug, Clone, PartialEq, Eq)]
pub enum DetectionOutcome {
    Disabled,
    Detected(Vec<Platform>),
}
#[derive(Clone)]
pub struct DetectionConfig {
    pub(crate) caller_identity_provider: Arc<dyn CallerIdentityProvider>,
}
impl Default for DetectionConfig {
    fn default() -> Self {
        Self {
            caller_identity_provider: Arc::new(StsCallerIdentityProvider),
        }
    }
}
/// A single platform detector.
///
/// Every detector receives the same [`DetectionConfig`] so new injectable
/// collaborators can be added without changing the detector list.
#[async_trait]
pub trait Detector: Send + Sync {
    fn platform(&self) -> Platform;
    async fn detect(&self, config: &DetectionConfig) -> bool;
}
struct AwsLambdaDetector;
#[async_trait]
impl Detector for AwsLambdaDetector {
    fn platform(&self) -> Platform {
        Platform::IsAwsLambda
    }
    async fn detect(&self, _config: &DetectionConfig) -> bool {
        env_non_empty("LAMBDA_TASK_ROOT")
    }
}
struct AzureFunctionDetector;
#[async_trait]
impl Detector for AzureFunctionDetector {
    fn platform(&self) -> Platform {
        Platform::IsAzureFunction
    }
    async fn detect(&self, _config: &DetectionConfig) -> bool {
        env_non_empty("FUNCTIONS_WORKER_RUNTIME")
            && env_non_empty("FUNCTIONS_EXTENSION_VERSION")
            && env_non_empty("AzureWebJobsStorage")
    }
}
struct GceCloudRunServiceDetector;
#[async_trait]
impl Detector for GceCloudRunServiceDetector {
    fn platform(&self) -> Platform {
        Platform::IsGceCloudRunService
    }
    async fn detect(&self, _config: &DetectionConfig) -> bool {
        env_non_empty("K_SERVICE")
            && env_non_empty("K_REVISION")
            && env_non_empty("K_CONFIGURATION")
    }
}
......
```

sfc-gh-jszczerbinski (approved):

sfc-gh-pbulawa (approved):

PR #892, opened 2026-04-18
SNOW-3396010: Metadata server detectors

sfc-gh-rsavenok:
## Summary
- Add five metadata-server–based platform detectors to `platform_detection`: `is_ec2_instance` (AWS IMDS), `is_azure_vm` and `has_azure_managed_identity` (Azure IMDS, plus an `IDENTITY_HEADER` fast path for Azure Functions), and `is_gce_vm` / `has_gcp_identity` (GCE metadata server, keyed off the `Metadata-Flavor: Google` header).
- Extend `DetectionConfig` with `ec2_metadata_url`, `azure_metadata_base_url`, `gce_metadata_root_url`, and `gce_metadata_base_url` so the real cloud endpoints are used by default but can be pointed at a `wiremock` server (or an unreachable `127.0.0.1:1`) in tests — keeping detectors fully offline under `cargo test`.
- Add `IDENTITY_HEADER` to `PLATFORM_DETECTION_ENV_KEYS` so it is scrubbed alongside the other platform-detection env vars during tests.
- Add `metadata_server_detectors` tests covering each new detector against a mocked metadata endpoint, plus a happy-path test for the Azure Functions + `IDENTITY_HEADER` shortcut.
- Unify the two per-detector timeout tests into a single `drops_detectors_when_timeout_reached` case that stalls both the STS provider and every HTTP metadata endpoint, asserting the whole batch is dropped by the shared `DETECTION_TIMEOUT`. Also tighten the `is_github_action` empty-string test to assert the full platforms list is empty.

## Context
Part of SNOW-3396010. Builds on the existing env-var and STS-based platform detection by letting the driver recognize AWS/Azure/GCP VMs and managed identities via their respective metadata services, so telemetry can report accurate hosting platforms even when no platform-specific env vars are set. The shared 200 ms `DETECTION_TIMEOUT` ensures these extra network probes can never block startup on non-cloud hosts.

## Test plan
- `cargo test -p sf_core telemetry::platform_detection` — all existing tests plus the new `metadata_server_detectors::*` cases and the unified `drops_detectors_when_timeout_reached` test pass.
- Verified detectors stay offline in tests: `test_detection_config()` points every metadata URL at `http://127.0.0.1:1`, and positive tests use `wiremock::MockServer` with strict path/header matchers (`Metadata: true` for Azure, `Metadata-Flavor: Google` for GCE).
- Verified the timeout path with `tokio::test(start_paused = true)` + `tokio::time::advance(DETECTION_TIMEOUT + 1ms)`: slow STS and slow HTTP responses (delay = `DETECTION_TIMEOUT * 4`) are all dropped and the returned platform list is empty.

### DIFF
--- a/sf_core/src/telemetry/platform_detection.rs
+++ b/sf_core/src/telemetry/platform_detection.rs
@@ -8,15 +8,27 @@ use super::aws_identity::{CallerIdentityProvider, StsCallerIdentityProvider, has
 
 const DETECTION_TIMEOUT: Duration = Duration::from_millis(200);
 
+const GCP_METADATA_FLAVOR_HEADER: &str = "Metadata-Flavor";
+const GCP_METADATA_FLAVOR_VALUE: &str = "Google";
+
 #[derive(Clone)]
 pub struct DetectionConfig {
     pub(crate) caller_identity_provider: Arc<dyn CallerIdentityProvider>,
+    pub(crate) ec2_metadata_url: String,
+    pub(crate) azure_metadata_base_url: String,
+    pub(crate) gce_metadata_root_url: String,
+    pub(crate) gce_metadata_base_url: String,
 }
 
 impl Default for DetectionConfig {
     fn default() -> Self {
         Self {
             caller_identity_provider: Arc::new(StsCallerIdentityProvider),
+            ec2_metadata_url: "http://169.254.169.254/latest/dynamic/instance-identity/document"
+                .to_string(),
+            azure_metadata_base_url: "http://169.254.169.254".to_string(),
+            gce_metadata_root_url: "http://metadata.google.internal".to_string(),
+            gce_metadata_base_url: "http://metadata.google.internal/computeMetadata/v1".to_string(),
         }
     }
 }
@@ -44,6 +56,8 @@ pub async fn detect_platforms(config: &DetectionConfig) -> Vec<String> {
         return vec!["disabled".to_string()];
     }
 
+    let http = reqwest::Client::new();
+
     let detectors: Vec<(&'static str, BoxFuture<'_, bool>)> = vec![
         ("is_aws_lambda", async { is_aws_lambda() }.boxed()),
         ("is_azure_function", async { is_azure_function() }.boxed()),
@@ -60,6 +74,14 @@ pub async fn detect_platforms(config: &DetectionConfig) -> Vec<String> {
             "has_aws_identity",
             has_aws_identity(config.caller_identity_provider.as_ref()).boxed(),
         ),
+        ("is_ec2_instance", is_ec2_instance(&http, config).boxed()),
+        ("is_azure_vm", is_azure_vm(&http, config).boxed()),
+        (
+            "has_azure_managed_identity",
+            has_azure_managed_identity(&http, config).boxed(),
+        ),
+        ("is_gce_vm", is_gce_vm(&http, config).boxed()),
+        ("has_gcp_identity", has_gcp_identity(&http, config).boxed()),
     ];
 
     let results = futures::future::join_all(detectors.into_iter().map(|(name, fut)| async move {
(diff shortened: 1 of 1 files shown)


sfc-gh-rsavenok:
* **#893** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/893?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#892** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/892?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/892?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#891** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/891?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#889** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/889?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#888** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/888?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#887** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/887?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sf_core/src/telemetry/platform_detection.rs
```diff
@@ -105,6 +127,71 @@ fn is_github_action() -> bool {
     env_non_empty("GITHUB_ACTIONS")
 }
 
+async fn is_ec2_instance(http: &reqwest::Client, config: &DetectionConfig) -> bool {
```
sfc-gh-pbulawa, line 129: This probably won't work on IMDSv2 instances. Is the IMDSv2 going to be implemented at a later phase?

sf_core/src/telemetry/platform_detection.rs
```diff
@@ -105,6 +127,71 @@ fn is_github_action() -> bool {
...
+    http.get(url)
+        .header("Metadata", "true")
+        .send()
+        .await
+        .map(|response| response.status().is_success())
+        .unwrap_or(false)
+}
+
+async fn is_gce_vm(http: &reqwest::Client, config: &DetectionConfig) -> bool {
+    http.get(&config.gce_metadata_root_url)
```
sfc-gh-pbulawa, line 192: We should send the header:  
`header(GCP_METADATA_FLAVOR_HEADER, GCP_METADATA_FLAVOR_VALUE)` here as in `has_gcp_identity` as Google requires it.

sfc-gh-pbulawa (changes requested):

sf_core/src/telemetry/platform_detection.rs
sfc-gh-rsavenok: Fixed in https://github.com/snowflakedb/universal-driver/pull/892/commits/f91ce445a47a1065b60fbab4489e97915eb761cd 

Also updated the code to match Go+Node that checks presence of isntanceId to ensure the actual response is valid ec2 response

sf_core/src/telemetry/platform_detection.rs
sfc-gh-rsavenok: I think this was a vibe mistake in JDBC, the original DBSec implementation in Python doesn't have it - https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/platform_detection.py#L328-L341

I run a query on our analytics to confirm that we have is_gce_vm entries from Python - there are millions of them

The same request with no header was used in Go and Node. According to claude, direct hit to http://metadata.google.internal doesnt require any headers and just returns list of endpoints

sfc-gh-pbulawa (approved):

PR #894, opened 2026-04-18
Injectable DiskReader interface

sfc-gh-rsavenok:
## Summary

Introduce a `DiskReader` abstraction for all filesystem reads performed by `DatabaseDriverV1`, and wire it through the driver's global state so tests can inject canned file contents instead of touching the real filesystem.

## Motivation

Driver code today reaches for `std::fs::read_to_string` directly when it needs to read files such as `/etc/hosts` or private key files. This makes integration tests hard to write deterministically and leaves no central place to add cross-cutting safeguards (ownership/permission checks, size limits, audit logging) that we will want for any disk read the driver performs.

## Changes

- Add `sf_core/src/apis/database_driver_v1/disk_reader.rs` defining:
  - `DiskReader` trait (`Send + Sync`) with a single `read_to_string(&Path) -> io::Result<String>` method.
  - `RealDiskReader`, the production implementation backed by `std::fs::read_to_string`.
  - `MockDiskReader` (test-only), an in-memory implementation built via a `with_file(path, contents)` builder; unregistered paths return `io::ErrorKind::NotFound` to match the behavior production code already expects.
- Extend `DatabaseDriverV1` (`global_state.rs`) with a lazily-initialized `OnceCell<Box<dyn DiskReader>>`:
  - `disk_reader()` returns `&dyn DiskReader`, defaulting to `RealDiskReader` on first use.
  - `set_disk_reader()` (test-only) injects a custom reader; panics if called after the cell is already initialized.
- Re-export `DiskReader` and `RealDiskReader` from `apis::database_driver_v1` (`mod.rs`).
- Unit tests cover: mock returning registered contents, mock returning `NotFound` for unknown paths, default reader behavior on the real driver, and mock injection via `set_disk_reader`.

## Notes for reviewers / follow-ups

- This PR only introduces the abstraction; no existing call sites are migrated yet. Subcomponents that read from disk should subsequently take a `&dyn DiskReader` via dependency injection rather than calling `std::fs` directly.
- The trait is intentionally minimal. Planned extensions (documented in the module docs): file ownership/permission verification, size limits, and centralized logging/error handling — all of which become possible precisely because reads now flow through a single chokepoint.

### DIFF
--- a/sf_core/src/apis/database_driver_v1/disk_reader.rs
+++ b/sf_core/src/apis/database_driver_v1/disk_reader.rs
@@ -1,0 +1,116 @@
+//! Abstraction layer for disk reads within `DatabaseDriverV1`.
+//!
+//! This module defines the interface for all filesystem reads performed by the
+//! driver. Its primary purpose is to allow integration tests to inject a
+//! [`MockDiskReader`] and avoid actual filesystem access—for example, to supply
+//! canned data for paths such as `/etc/hosts` or simulated private key files.
+//!
+//! The [`DiskReader`] trait is deliberately minimal for now but is expected
+//! to evolve with additional safety and auditing features that should apply
+//! uniformly to all disk reads, such as:
+//!
+//! - Verifying file ownership and permissions before reading;
+//! - Enforcing a file size limit;
+//! - Centralized logging and error handling.
+//!
+//! Important: To ensure all future safeguards are applied consistently, production code
+//! should never use [`std::fs`] directly. All file reads must go through a [`DiskReader`].
+//!
+//! In normal operation, a [`DiskReader`] instance is managed by
+//! [`crate::apis::database_driver_v1::DatabaseDriverV1`].
+//! For proper testability, all subcomponents of `DatabaseDriverV1` which
+//! need to read from disk should receive a reference to the `DiskReader` via
+//! dependency injection, ensuring they always use the correct mechanism.
+
+use std::io;
+use std::path::Path;
+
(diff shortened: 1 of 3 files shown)


sf_core/src/apis/database_driver_v1/disk_reader.rs
```diff
@@ -0,0 +1,116 @@
...
+//! In normal operation, a [`DiskReader`] instance is managed by
+//! [`crate::apis::database_driver_v1::DatabaseDriverV1`].
+//! For proper testability, all subcomponents of `DatabaseDriverV1` which
+//! need to read from disk should receive a reference to the `DiskReader` via
+//! dependency injection, ensuring they always use the correct mechanism.
+
+use std::io;
+use std::path::Path;
+
+pub trait DiskReader: Send + Sync {
```
sfc-gh-rsavenok, line 28: FsProvider -> FsAdaper

sf_core/src/apis/database_driver_v1/disk_reader.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn mock_returns_registered_contents() {
+        let reader = MockDiskReader::new()
+            .with_file("/etc/hosts", "127.0.0.1 localhost\n")
```
sfc-gh-rsavenok, line 94: pathBuf

PR #900, opened 2026-04-20
SNOW-32405: Prevent spurious binding dereference in JSON Binding

sfc-gh-ddas:
### TL;DR

Fix potential unsafe memory access by capping parameter serialization to the server-reported marker count and tightening the `prepared_param_count` type from `usize` to `u16`.

### What changed?

- `prepared_param_count` on `Statement` is now `Option<u16>` instead of `Option<usize>`, matching the ODBC maximum parameter count.
- `apply_parameter_bindings` now accepts a `prepared_param_count: Option<u16>` argument. For prepared statements, this value is required and used as the authoritative parameter count (`effective_count`). For `SQLExecDirect`, the effective count is derived from the APD/IPD descriptor counts as before.
- `odbc_bindings_to_json` now accepts an explicit `max_params: u16` ceiling, replacing the internally computed `max_key`. This prevents the serializer from ever dereferencing APD records beyond the server-reported marker count.
- `prepare_impl` now validates that the server-reported bind count fits in a `u16`, returning a `CountFieldIncorrect` error if it does not, and logs a warning when the server reports a negative count.
- `exec_direct_impl` resets `prepared_param_count` to `None` and passes `None` as the cap to `apply_parameter_bindings`.
- DAE parameter detection in `find_dae_params` uses `u16` comparisons throughout.
- All `ipd.records.retain` calls compare against `u16` directly.

### How to test?

- New unit tests in `statement.rs` verify that calling `apply_parameter_bindings` with `prepared=true` and `prepared_param_count=None` returns a `CountFieldIncorrect` error.
- New unit tests in `param_binding.rs` verify:
  - `max_params=0` produces an empty JSON object even when a phantom DAE binding with a dummy pointer exists, confirming the pointer is never dereferenced.
  - `max_params=1` serializes only the first parameter when two bindings are present, leaving the phantom second binding (with a dummy pointer) untouched.

### Why make this change?

When a prepared statement has fewer parameter markers than the number of APD bindings (e.g., `SELECT 1` with a bound DAE parameter), the serializer could attempt to dereference APD records that point to invalid or dummy memory addresses, causing undefined behavior or crashes. By threading the server-reported marker count through as an explicit cap, phantom bindings beyond the actual marker count are never accessed.

### DIFF
--- a/odbc/src/api/statement.rs
+++ b/odbc/src/api/statement.rs
@@ -27,11 +27,11 @@ use tokio_util::sync::CancellationToken;
 use tracing;
 
 /// Scan the APD for parameters marked as data-at-execution.
-fn find_dae_params(apd: &crate::api::ApdDescriptor, param_limit: Option<usize>) -> Vec<u16> {
+fn find_dae_params(apd: &crate::api::ApdDescriptor, param_limit: Option<u16>) -> Vec<u16> {
     let mut dae_params = Vec::new();
     for (&param_num, record) in &apd.records {
         if let Some(limit) = param_limit
-            && param_num as usize > limit
+            && param_num > limit
         {
             continue;
         }
@@ -81,6 +81,11 @@ fn exec_direct_impl(statement_handle: sql::Handle, statement_text: &str) -> Odbc
         ConnectionState::Connected { .. } => {}
     }
 
+    // exec-direct supersedes any prior SQLPrepare on this handle; clear the
+    // stale marker count before the DAE scan so it cannot leak into a later
+    // free_stmt(ResetParams) or SQLCancel → SQLExecute flow.
+    stmt.prepared_param_count = None;
+
     // No param_limit for exec-direct: enter DAE if any APD record has a DAE
     // indicator, regardless of the SQL text.  We deliberately avoid client-side
     // `?` parsing because (a) it can disagree with the server's parser on edge
@@ -112,7 +117,8 @@ fn exec_direct_impl(statement_handle: sql::Handle, statement_text: &str) -> Odbc
             db_handle: _,
             conn_handle,
         } => {
-            let (bindings, _json_owner) = apply_parameter_bindings(&stmt.apd, &stmt.ipd, false)?;
+            let (bindings, _json_owner) =
+                apply_parameter_bindings(&stmt.apd, &stmt.ipd, false, None)?;
             let stmt_handle = stmt.stmt_handle;
 
             stmt.cancel_token = CancellationToken::new();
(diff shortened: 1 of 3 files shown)


sfc-gh-ddas:
* **#990** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/990?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#911** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/911?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#903** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/903?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#900** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/900?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/900?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

odbc/src/api/statement.rs
```diff
@@ -473,49 +494,63 @@ fn create_execute_state(
...
-        if prepared {
-            let ipd_count = ipd.desc_count() as usize;
-            if ipd_count > 0 {
-                return crate::api::error::CountFieldIncorrectSnafu {
-                    reason: format!(
-                        "parameter 1 is not bound (statement has {ipd_count} parameter markers)"
-                    ),
-                }
-                .fail();
+    let effective_count: u16 = if prepared {
```
sfc-gh-makowalski, line 511: NIT Later it's passed to `odbc_bindings_to_json` as `max_params`. Maybe it's worth to unify the name to something like `marker_count`?

odbc/src/conversion/param_binding.rs
```diff
@@ -1792,7 +1791,64 @@ mod tests {
                 ..IpdRecord::default()
             },
         );
-        assert!(odbc_bindings_to_json(&apd, &ipd).is_err());
+        assert!(odbc_bindings_to_json(&apd, &ipd, apd.desc_count().max(ipd.desc_count())).is_err());
+    }
+
+    #[test]
+    fn max_params_zero_skips_phantom_dae_binding() -> TestResult {
```
sfc-gh-makowalski, line 1798: **Untested integration scenario worth adding.**

The unit tests cover `odbc_bindings_to_json` directly, which is great for the memory-safety contract. An end-to-end test would also be valuable: `SQLPrepare("SELECT 1")` → `SQLBindParameter(1, …, SQL_DATA_AT_EXEC)` → `SQLExecute` and assert it returns success (not `SQL_NEED_DATA`, not crash). That ties the whole fix together at the level the bug was actually reported.

sfc-gh-makowalski (changes requested):
Overall LGTM, just address the comments, and I'll approve. It's alright if you don't agree with some of them, I'm not pushing as I can see other PRs stacked on top.

odbc/src/conversion/param_binding.rs
sfc-gh-ddas: We already have E2E tests for this case, however they didn't fail before, probably because the DM defensively checks for this. So this unit test is the only way to ensure the driver behaviour that I can think of.

odbc/src/api/statement.rs
sfc-gh-ddas: The current names carry distinctions at each abstraction layer. Unifying them would lose clarity.

sfc-gh-makowalski (approved):
LGTM

PR #901, opened 2026-04-20
fix(odbc): reject non-finite string literals bound to SQL_REAL/DOUBLE with 22018

sfc-gh-makowalski:
## Summary

The ODBC `numeric-literal` grammar (MS ODBC spec, Appendix C) does not admit
the tokens `Infinity`, `-Infinity` or `NaN`. Rust's `f64::from_str` happens to
accept them, so `SnowflakeReal::read_odbc` used to parse such strings
successfully and then fail late with SQLSTATE `22003`, which is the wrong
state for a malformed literal.

This PR:

- Adds `JsonBindingError::InvalidNumericLiteral` mapped to SQLSTATE **22018**
  (Invalid character value for cast specification).
- Uses it in the `SQL_C_CHAR` / `SQL_C_WCHAR` arms of `real.rs` whenever the
  parsed `f64` is not finite.
- Adds six E2E tests (CHAR × 3, WCHAR × 3) asserting rejection with `22018`,
  plus Rust unit tests for the new error variant.
- Documents the intentional divergence as **BD#47** (old driver accepts these
  strings and stores non-finite `FLOAT` values; new driver rejects per spec).

Applications that need to insert non-finite IEEE-754 values should bind
`SQL_C_FLOAT` / `SQL_C_DOUBLE`, which are spec-permitted and supported end-to-end
once #899 lands.

## Test plan

- [x] `cargo test --lib` (992 tests pass)
- [x] New Rust unit tests for CHAR/WCHAR Infinity/-Infinity/NaN → `InvalidNumericLiteral`
- [x] `invalid_numeric_literal_maps_to_22018` mapping test
- [x] Pre-commit: rust-fmt / rust-clippy / rust-cargo-check / clang-format
- [x] CI: all six new E2E tests pass (new driver)
- [x] CI: old-driver reference run skips the six new tests (BD#47)


odbc/src/conversion/real.rs
```diff
@@ -416,38 +475,26 @@ impl ReadODBC for SnowflakeReal {
             }
             CDataType::Char => {
                 let s = read_char_str(binding)?;
-                let v = s.trim().parse::<f64>().map_err(|_| {
+                let trimmed = s.trim();
```
sfc-gh-pbulawa, line 478: The `Char` and `WChar` arms in `ReadODBC for SnowflakeReal` use `UnsupportedCDataTypeSnafu` when `parse::<f64>()` fails entirely (e.g., for input `"hello"`). This error maps to `RestrictedDataTypeAttributeViolation` (HY004), which means "the SQL data type is not supported." But `SQL_C_CHAR` IS a supported C type for `SQL_DOUBLE` — the problem is the string value, not the type. Per the ODBC spec ("C to SQL: Character" table), the correct SQLSTATE is **22018** (Invalid character value for cast specification).

sfc-gh-pbulawa:
Add tests:

- unparseable strings - test like `convert_char_garbage_as_real_rejected` with input `"hello"` bound to `SQL_DOUBLE`
- No WChar negative overflow test. `convert_wchar_neg_overflow_as_real_overflows` (UTF-16 `"-1e309"` → 22003).
- Limited case variation coverage. Only lowercase `"inf"` is tested beyond the canonical spellings

sfc-gh-pbulawa (approved):

PR #910, opened 2026-04-21
Migrate env handles to a safe solution

sfc-gh-jszczerbinski:
Migrate environment handles to a registry-based solution

- Introduce Env wrapper with Mutex<Environment> for thread-safe access to environment state
- Add HandleRegistry to manage environment handles by ID instead of raw pointers
- Wrap connection handles in a new Dbc struct with a Weak<Env> back-reference to the parent environment
- Track parent-child relationships between environments and connections
- Cascade cleanup of child connections when freeing an environment handle

### DIFF
--- a/odbc/src/api/handle_allocation.rs
+++ b/odbc/src/api/handle_allocation.rs
@@ -1,10 +1,16 @@
-use crate::api::error::{DisconnectedSnafu, InvalidHandleSnafu, OdbcRuntimeSnafu, Required};
+use crate::api::env_from_handle;
+use crate::api::error::{
+    ConnectionStillConnectedSnafu, DisconnectedSnafu, EnvironmentHasConnectionsSnafu,
+    InvalidHandleSnafu, OdbcRuntimeSnafu, Required,
+};
+use crate::api::handle_registry::HandleId;
 use crate::api::{
-    Connection, ConnectionState, Environment, OdbcResult, Statement, conn_from_handle,
+    Connection, ConnectionState, Dbc, Env, Environment, OdbcResult, Statement, conn_from_handle,
     diagnostic::DiagnosticInfo,
     runtime::{env_allocated, env_freed, global},
 };
 use odbc_sys as sql;
+use parking_lot::Mutex;
 use sf_core::protobuf::generated::database_driver_v1::{
     StatementNewRequest, StatementReleaseRequest,
 };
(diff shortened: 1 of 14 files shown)


odbc/src/api/types/odbc_types.rs
```diff
@@ -862,12 +869,21 @@ impl ToSqlReturn for OdbcResult<()> {
...
+// Will be removed after this PR stack is completed
+unsafe impl Send for Env {}
+unsafe impl Sync for Env {}
 
 pub struct Environment {
     pub odbc_version: sql::Integer,
     pub connection_pooling: sql::AttrConnectionPooling,
     pub connection_pool_match: sql::AttrCpMatch,
     pub diagnostic_info: DiagnosticInfo,
+    pub connections: Vec<*mut Dbc>,
```
sfc-gh-ddas: With this, we are introducing parent -> child tracking for handles. This might be necessary for our use case, but this is the opposite of Core which only has child -> parent tracking. Can we get away with tracking in just one direction to avoid risk of inconsistencies?

odbc/src/api/types/odbc_types.rs
```diff
@@ -862,12 +869,21 @@ impl ToSqlReturn for OdbcResult<()> {
         self.to_sql_return(warnings).0
     }
 }
+pub struct Env {
```
sfc-gh-ddas, line 872: I think we should have clearer names for the wrappers

odbc/src/api/handle_registry.rs
```diff
@@ -0,0 +1,171 @@
...
+
+impl From<sql::Handle> for HandleId {
+    fn from(handle: sql::Handle) -> Self {
+        Self {
+            id: handle as usize,
+        }
+    }
+}
+
+pub struct RemovalGuard<T: Send + Sync + Clone> {
```
sfc-gh-ddas, line 31: This serializes all access to the environment during a free attempt which is a solution to our problem, but can cause slowdowns and make multi-threaded ODBC operations dependent on each other. This also introduces nested locks, which should be acquired in a specific order else there will be a deadlock, we should make this clear with comments.

odbc/src/api/types/odbc_types.rs
```diff
@@ -1297,19 +1326,23 @@ impl Statement {
...
+        .context(OdbcRuntimeSnafu)?
+        .env_registry
+        .get(handle_id)?;
+    Ok(env)
 }
 
-pub fn conn_from_handle<'a>(handle: sql::Handle) -> &'a mut Connection {
-    let conn_ptr = handle as *mut Connection;
+pub fn conn_from_handle<'a>(handle: sql::Handle) -> &'a mut Dbc {
+    let conn_ptr = handle as *mut Dbc;
```
sfc-gh-ddas: Until all handles are registry managed, this pattern is vulnerable to segfaults if application passes the wrong type of handle. With environment handles now having registry generated IDs, this might try to dereference an integer as a pointer.

odbc/src/api/handle_allocation.rs
```diff
@@ -13,47 +19,57 @@ use std::sync::Arc;
...
+        }),
     });
-    Ok(Box::into_raw(env))
+    let handle = global().context(OdbcRuntimeSnafu)?.env_registry.add(env)?;
+    Ok(handle.into())
 }
 
 /// Allocate a new connection handle
-pub fn alloc_connection() -> OdbcResult<*mut Connection> {
+pub fn alloc_connection(env: Arc<Env>) -> OdbcResult<*mut Dbc> {
```
sfc-gh-ddas: There can be a race between freeing an environment and a connection allocation if the environment is freed before the connection is allocated but after the Weak pointer is created. The check `connections.is_empty()` happens without synchronization against concurrent `alloc_connection`.

sfc-gh-ddas (dismissed):

odbc/src/api/types/odbc_types.rs
sfc-gh-jszczerbinski: Almost all ODBC drivers are researched track in this direction. For a good reason I think, the ODBC suggest this hierchachical structure

odbc/src/api/types/odbc_types.rs
sfc-gh-jszczerbinski: Not sure what do you mean exactly

odbc/src/api/types/odbc_types.rs
sfc-gh-jszczerbinski: Wlll be fixed in next PR

odbc/src/api/handle_allocation.rs
sfc-gh-jszczerbinski: Will be fixed during connection handle migration

odbc/src/api/handle_registry.rs
sfc-gh-jszczerbinski: Added comments, I think locking the object I am deleting won't cause slowdowns

sfc-gh-pbulawa (approved):

PR #913, opened 2026-04-21
Add retries to sync query request

sfc-gh-makowalski:
## Summary

Follow-up to #875, which added HTTP-level retries to the login endpoint. That change only wrapped `/session/v1/login-request`; the sync query POST at `/queries/v1/query-request` was still issued via a plain `client.execute(request).await` with no retry, so transient 5xx responses from Snowflake — notably the Azure 503 flakes we've been seeing post-merge on `main` — surface directly as user-visible `HY000` errors and as red CI.

This wraps `execute_sync_query`'s POST with `execute_with_retry` using the exact same pattern as #875 (`HttpContext::new(POST, …).allow_post_retry()` + `RetryPolicy::default()`).

### Why this is safe on POST

- The `requestId` query param is generated once by the caller and is **stable** across retry-closure invocations, so Snowflake's existing request-id dedupe correctly handles the rare case where the server executed the query before returning a 5xx — the replay returns the cached result rather than running the statement twice.
- The `request_guid` and body are also identical across attempts.
- The outer app-level `is_retry` flag (used after token-refresh retries) is preserved unchanged; this change only adds HTTP-level retries underneath it.

### Evidence this was needed

Today's 503 wave on `main` (post-merge of #909) hit this exact site — `sf_core/src/rest/snowflake/mod.rs:1156` / `:1445` — causing ~112 failures in `odbc_tests (Linux x64, azure)` and ~98 in `odbc_tests (macOS ARM64, azure)`:
- https://github.com/snowflakedb/universal-driver/actions/runs/24710700754/job/72274693858
- https://github.com/snowflakedb/universal-driver/actions/runs/24710700754/job/72274693884

With #875 only covering login, those flakes never entered the retry path.

## Test plan

- [x] `cargo check -p sf_core --tests` — clean.
- [x] `cargo fmt -p sf_core -- --check` — clean.
- [x] `cargo clippy -p sf_core --tests --no-deps` — clean.
- [x] `cargo test -p sf_core --lib rest::snowflake::tests` — 25 passed, 0 failed, including the new `execute_sync_query_retry_tests::retries_on_503_then_succeeds` and the existing `send_login_request_retry_tests::retries_on_503_then_succeeds`.
- [x] Full ODBC / Python / Rust-core CI on the PR (relies on CI).

## Follow-ups

Three other endpoints in this file still use raw `client.execute(request)` and would benefit from the same treatment — happy to do them in separate PRs to keep each change reviewable:
- `mod.rs:658` — session refresh
- `mod.rs:770` — token request
- `mod.rs:1394` — abort query

### DIFF
--- a/sf_core/src/rest/snowflake/mod.rs
+++ b/sf_core/src/rest/snowflake/mod.rs
@@ -1130,25 +1130,34 @@ async fn execute_sync_query<'a>(
             path: QUERY_REQUEST_PATH,
         })?;
 
-    // Build query parameters - include retry=true if this is a retry.
-    // The `requestId` is stable across HTTP-level retries so the server can
-    // dedupe replays via its normal request-id machinery.
-    let mut query_params = vec![
+    // Base query parameters - the `requestId` is stable across HTTP-level
+    // retries so the server can dedupe replays via its normal request-id
+    // machinery. The `retry=true` flag is the documented Snowflake hint that
+    // signals "look up requestId in the dedup table"; for unknown requestIds
+    // the server simply executes the query as a fresh request.
+    let base_query_params = vec![
         ("requestId", request_id.to_string()),
         ("request_guid", uuid::Uuid::new_v4().to_string()),
     ];
-    if is_retry {
-        query_params.push(("retry", "true".to_string()));
-    }
 
     let send_start = Instant::now();
+    let attempt_counter = std::sync::atomic::AtomicU32::new(0);
     let build_request = || {
+        let n = attempt_counter.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+        // Set retry=true if either the outer caller already considers this a
+        // retry, or this is a second-or-later HTTP-level attempt within
+        // execute_with_retry. Always safe per Snowflake docs; only increases
+        // dedupe accuracy when the server has already seen this requestId.
+        let mut params = base_query_params.clone();
+        if is_retry || n >= 1 {
+            params.push(("retry", "true".to_string()));
+        }
         apply_json_content_type(apply_query_headers(
             client.post(query_url.clone()),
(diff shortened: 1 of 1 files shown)


sfc-gh-makowalski:
Flattened the retry layers (commit 0e7c0e16, pushed on top of the rebase onto `main`).

**What changed:**
- Removed the outer retry loop from `execute_sync_with_retry`. That loop only handled `RestError::Communication`, but since `execute_sync_query` wraps the `POST /queries/v1/query-request` in `execute_with_retry`, transport failures are already surfaced as `RestError::HttpRetry`. The outer loop was dead code — it could never be entered.
- Dropped the `is_retry: bool` parameter from `execute_sync_query`. The inner `build_request` closure already sets `retry=true` on every HTTP replay (`attempt >= 2`) via the `attempt_counter`, so the outer boolean was redundant.
- `execute_sync_with_retry` is now a thin facade: it mints the `requestId` (logged once up front) and delegates to `execute_sync_query` with the retry policy.

**Behaviour:**
- The request ID is still stable across all HTTP attempts (retry dedupe on the server side unchanged).
- The first attempt is still sent without `retry=true`; every replay still carries `retry=true`.
- On exhausted retries, callers now get `RestError::HttpRetry` (the honest error from `execute_with_retry`) instead of a `RestError::Communication` synthesized from a branch that could never run.

**Test coverage:** the existing wiremock test (`retries_on_503_then_succeeds_and_sets_retry_flag_on_replays`) still passes unchanged — it already asserts 3 attempts with the same `requestId` and `retry=true` on replays. Full `rest::` suite (80 tests) green locally.

Net diff: +24 / -102 lines.

sfc-gh-jszczerbinski (approved):

sfc-gh-ddas (approved):

PR #915, opened 2026-04-21
fixup json conversion for DECFLOAT

sfc-gh-pczajka:


python/tests/e2e/types/test_decfloat.py
```diff
@@ -165,6 +165,33 @@ def test_should_download_large_result_set_with_multiple_chunks_from_generator(se
...
+            cursor = conn.cursor()
+
+            # When Query "SELECT CAST('1234.56789012345678901234567890' AS DECFLOAT) AS test_value" is executed
+            cursor.execute("SELECT CAST('1234.56789012345678901234567890' AS DECFLOAT) AS test_value")
+            row = cursor.fetchone()
+
+            # Then Result should be returned as appropriate type with value 1234.56789012345678901234567890
+            assert row is not None
+            value = row[0]
+            assert isinstance(value, (Decimal, str)), f"Expected Decimal or str, got {type(value)}"
```
sfc-gh-fpawlowski: Would we be able to split those cases into 2? or do that in a loop and enforce both str and Decimal representations and make sure we are testing both paths?

sf_core/src/rest/snowflake/query_response.rs
```diff
@@ -558,17 +565,14 @@ impl TryFrom<&RowType> for query_types::RowType {
 
         match value.type_.to_uppercase().as_str() {
             "TEXT" => {
-                let length = value.length.context(MissingParameterSnafu {
```
sfc-gh-fpawlowski, line 561: nit: was path with raising here the MissingParameterSnafu tested anywhere? why didn't we break any tests with this change?

sfc-gh-fpawlowski (approved):
LGTM; left 2 nits

sfc-gh-fpawlowski (commented):
LGTM; left 2 nits

sfc-gh-fpawlowski (changes requested):
Looks good, left 2 nits 👍

python/tests/e2e/types/test_decfloat.py
```diff
@@ -165,6 +165,33 @@ def test_should_download_large_result_set_with_multiple_chunks_from_generator(se
...
+    Regression: when the server uses JSON format and doesn't recognize the client
+    as DECFLOAT-capable, it returns DECFLOAT columns as TEXT type without
+    length/byteLength metadata. The driver must handle this gracefully.
+    """
+
+    def test_should_select_decfloat_literal_with_json_result_format(self, connection_factory):
+        # Given Snowflake client is logged in
+        pass
+
+        # And Session parameter PYTHON_CONNECTOR_QUERY_RESULT_FORMAT is set to JSON
```
sfc-gh-pfus, line 181: I will have a PR that should run all tests on JSON. Maybe we can add a TODO here, since this will be run for both Arrow and JSON?

sf_core/src/rest/snowflake/query_response.rs
```diff
@@ -6,6 +6,13 @@ use snafu::{OptionExt, Snafu};
 use std::collections::HashMap;
 // TODO: Delete all unused fields when we are sure they are not needed
 
+/// Snowflake's default maximum VARCHAR length (16 MB in characters).
+/// Used as fallback when the server omits length metadata for TEXT columns.
+const DEFAULT_TEXT_LENGTH: u64 = 16_777_216;
```
sfc-gh-pfus, line 13: I'm not convinced if this is the correct approach. This number is static and may be the subject of change on the underlying Snowflake - and the truth is, there is a project to extend this number to 128MB. Why do we need this? Are we expecting to break something when length is not given?

sf_core/src/rest/snowflake/query_response.rs
```diff
@@ -6,6 +6,13 @@ use snafu::{OptionExt, Snafu};
...
+
+/// Multiplier to derive byte_length from character length (UTF-8 max 4 bytes/char).
+const DEFAULT_TEXT_BYTE_LENGTH_MULTIPLIER: u64 = 4;
```
sfc-gh-pfus: Isn't it more like x2? At least this is what I remember from max-lob-size project.

tests/definitions/shared/types/decfloat.feature
```diff
@@ -52,6 +52,17 @@ Feature: DECFLOAT type support
...
+  # =========================================================================== #
+  #                         JSON result format                                 #
+  # =========================================================================== #
+
+  @python_e2e
+  Scenario: should select decfloat literal with JSON result format
+    Given Snowflake client is logged in
+    And Session parameter PYTHON_CONNECTOR_QUERY_RESULT_FORMAT is set to JSON
+    When Query "SELECT CAST('1234.56789012345678901234567890' AS DECFLOAT) AS test_value" is executed
+    Then Result should be returned as appropriate type with value 1234.56789012345678901234567890
```
sfc-gh-pfus, line 65: If this is the real problem, I think we should fix it on the backend side, not in a driver.

sfc-gh-pfus (changes requested):

sf_core/src/rest/snowflake/query_response.rs
sfc-gh-pczajka: There is already a test breaking in Snowflake CLI (one of Python's early adopters) because of this.

sf_core/src/rest/snowflake/query_response.rs
sfc-gh-pczajka: It's worse - x1. From examples on SnowAPI docs, server always returns length = bytelength
https://docs.snowflake.com/en/developer-guide/sql-api/reference

sf_core/src/rest/snowflake/query_response.rs
sfc-gh-pczajka: Official docs say 16MB is the default for VARCHAR, I've updated the comment https://docs.snowflake.com/en/sql-reference/data-types-text

python/tests/e2e/types/test_decfloat.py
sfc-gh-pczajka: done

sf_core/src/rest/snowflake/query_response.rs
sfc-gh-pczajka: Apparently not - build are green (not Jenkins, as usual :P)

tests/definitions/shared/types/decfloat.feature
sfc-gh-pczajka: We need this working for PrPr (was working in old connector)

python/tests/e2e/types/test_decfloat.py
sfc-gh-pczajka: Should be "Decimal" only, nice catch

sf_core/src/rest/snowflake/query_response.rs
sfc-gh-fpawlowski: Okay, then could we do a follow-up PR soon after this that will add such tests? We should at least enter this path in our test suites to make sure those defaults behave as expected.

sfc-gh-fpawlowski (approved):
Approved, but let's do a follow-up with those additional tests as it is important to cover those paths quickly

sfc-gh-pfus (approved):

PR #922, opened 2026-04-21
Send OS_DETAILS in CLIENT_ENVIRONMENT

sfc-gh-rsavenok:
## Summary
- Add `OS_DETAILS` field to the login request's `CLIENT_ENVIRONMENT`, mirroring the Node.js driver's telemetry feature (snowflake-connector-nodejs `lib/telemetry/os_details`).
- New `sf_core::telemetry::os_details::detect_os_details` module that parses `/etc/os-release` on Linux via the existing `FsAdapter` and extracts an allow-listed subset of keys (`NAME`, `PRETTY_NAME`, `ID`, `BUILD_ID`, `IMAGE_ID`, `IMAGE_VERSION`, `VERSION`, `VERSION_ID`). Returns `None` on non-Linux platforms or on read/parse failure.
- Result is cached on `DatabaseDriverV1` in a `OnceCell` (same pattern as `platforms`) and wired through `ClientInfo.os_details` into `AuthRequestClientEnvironment.OS_DETAILS` with `skip_serializing_if = Option::is_none`, so the field is omitted when unavailable.
- No new crate dependency: the `os-release` format is trivial enough to hand-parse, avoiding a `regex` dep.

## Context
Follow-up to the telemetry foundations that added `FsAdapter`, SPCS token plumbing, and the `PLATFORM` field in the login request. `OS_DETAILS` is the next CLIENT_ENVIRONMENT field being rolled out across all connectors for server-side telemetry.

Node.js reference:
- https://github.com/snowflakedb/snowflake-connector-nodejs/blob/master/lib/telemetry/os_details/index.ts
- https://github.com/snowflakedb/snowflake-connector-nodejs/blob/master/lib/telemetry/os_details/linux_os_release.ts

## Test plan
- Unit tests in `sf_core/src/telemetry/os_details.rs` (8 cases): parses the Node.js Arch Linux fixture, ignores disallowed keys, handles unquoted values and malformed/commented lines, rejects lowercase keys, and covers the Linux-only detect paths (`detect_returns_none_when_file_missing_on_linux`, `detect_parses_allow_listed_keys_via_mock_fs_on_linux`, `detect_returns_none_on_non_linux`).
- Integration test `sf_core/tests/integration/authentication/os_details.rs`: compiles on every platform; on non-Linux it early-returns (`if !cfg!(target_os = "linux") { return; }`). On Linux it injects a `MockFs` holding a fake `/etc/os-release`, mounts a wiremock matcher that asserts `data.CLIENT_ENVIRONMENT.OS_DETAILS` contains the expected allow-listed keys, and exercises `connection_init` end-to-end.
- Verified locally on macOS:
  - `cargo build -p sf_core --lib` — clean
  - `cargo build -p sf_core --bin collect_chunk_data --features cli` — clean
  - `cargo test -p sf_core --lib telemetry::os_details` — 8/8 pass
  - `cargo test -p sf_core --test integration_tests authentication` — 37/37 pass (existing auth suite still green; new test early-returns on macOS as designed)
  - `cargo test -p sf_core --lib rest::snowflake` — 79/79 pass
  - `cargo clippy -p sf_core --lib --tests --all-features -- -D warnings` — clean

### DIFF
--- a/sf_core/src/telemetry/os_details.rs
+++ b/sf_core/src/telemetry/os_details.rs
@@ -1,0 +1,183 @@
+use std::collections::HashMap;
+use std::path::Path;
+
+use crate::fs_adapter::FsAdapter;
+
+const OS_RELEASE_PATH: &str = "/etc/os-release";
+const ALLOWED_KEYS: &[&str] = &[
+    "NAME",
+    "PRETTY_NAME",
+    "ID",
+    "BUILD_ID",
+    "IMAGE_ID",
+    "IMAGE_VERSION",
+    "VERSION",
+    "VERSION_ID",
+];
+
+/// Detect OS details for telemetry.
+///
+/// Returns `Some(map)` on Linux when `/etc/os-release` can be read and
+/// contains at least one allow-listed key; otherwise returns `None`.
+pub fn detect_os_details(fs: &dyn FsAdapter) -> Option<HashMap<String, String>> {
+    if !cfg!(target_os = "linux") {
+        return None;
+    }
+
+    match fs.read_to_string(Path::new(OS_RELEASE_PATH)) {
+        Ok(contents) => {
+            let parsed = parse_os_release(&contents);
+            if parsed.is_empty() {
+                None
+            } else {
+                Some(parsed)
+            }
+        }
+        Err(e) => {
+            tracing::debug!(
+                path = OS_RELEASE_PATH,
+                error = %e,
+                "Failed to read OS release file for telemetry"
+            );
+            None
+        }
+    }
+}
+
+/// Parse an `os-release(5)` file body and return the allow-listed keys.
+///
+/// Each line is expected to be `KEY=value` or `KEY="value"`. Lines that do
+/// not match or whose key is not in [`ALLOWED_KEYS`] are ignored.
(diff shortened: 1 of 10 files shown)


sf_core/src/telemetry/os_details.rs
```diff
@@ -0,0 +1,183 @@
...
+                "Failed to read OS release file for telemetry"
+            );
+            None
+        }
+    }
+}
+
+/// Parse an `os-release(5)` file body and return the allow-listed keys.
+///
+/// Each line is expected to be `KEY=value` or `KEY="value"`. Lines that do
```
sfc-gh-mcepiga: can't the value also be enclosed in single quotes?

sfc-gh-turbaszek (approved):

sfc-gh-mcepiga (approved):

sf_core/src/telemetry/os_details.rs
sfc-gh-rsavenok: Very good catch! 

> Variable
>        assignment values must be enclosed in double or single quotes if
>        they include spaces, semicolons or other special characters
>        outside of A–Z, a–z, 0–9

PR #939, opened 2026-04-23
Add GEOMETRY type e2e test suite

sfc-gh-pczajka:
## Summary
- Add comprehensive e2e tests for GEOMETRY type: type casting, literal selects, NULL handling, table operations, multi-chunk downloads, parameter binding, and JSON result format
- Add `parse_geojson()` helper to `utils.py` for GeoJSON string parsing
- Add `geometry.feature` Gherkin definitions (11 scenarios)

## Test plan
- [ ] All 11 GEOMETRY tests pass on universal driver (currently blocked by pre-existing protobuf bug)
- [ ] All 11 GEOMETRY tests pass on reference driver
- [ ] Test format validator passes

### DIFF
--- a/python/tests/e2e/types/test_geometry.py
+++ b/python/tests/e2e/types/test_geometry.py
@@ -38,7 +38,10 @@ POLYGON_GEOJSON_COORDS = [[[0.0, 0.0], [4.0, 0.0], [4.0, 3.0], [0.0, 3.0], [0.0,
 # =============================================================================
 LARGE_RESULT_SET_SIZE = 20_000
 
+SKIP_JSON = pytest.mark.skip_for_json_result_set(reason="GEOMETRY type is not supported in JSON result format")
 
+
+@SKIP_JSON
 class TestGeometryLiteral:
     """Tests for GEOMETRY type using SELECT with literals (no tables)."""
 
@@ -68,6 +71,7 @@ class TestGeometryLiteral:
         assert_geojson(result[0], expected_type, expected_coords)
 
 
+@SKIP_JSON
 class TestGeometryTypeCasting:
     """Tests for GEOMETRY type casting per output format.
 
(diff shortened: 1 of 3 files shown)


tests/definitions/shared/types/geometry.feature
```diff
@@ -0,0 +1,95 @@
...
+    When Query "SELECT * FROM <table> ORDER BY id" is executed
+    Then Result should contain [GeoJSON Point, NULL]
+
+  # =========================================================================== #
+  #                       Multiple chunks downloading                           #
+  # =========================================================================== #
+
+  @python_e2e
+  Scenario: should download geometry data in multiple chunks
+    # skip_for_json_result_set
```
sfc-gh-asolarski, line 70: we don't add it in scenarios in all other cases

python/tests/e2e/types/test_geometry.py
```diff
@@ -0,0 +1,238 @@
...
+# Simple rectangle polygon (ring must be closed)
+POLYGON_WKT = "POLYGON((0 0, 4 0, 4 3, 0 3, 0 0))"
+POLYGON_GEOJSON_COORDS = [[[0.0, 0.0], [4.0, 0.0], [4.0, 3.0], [0.0, 3.0], [0.0, 0.0]]]
+
+# =============================================================================
+# LARGE RESULT SET SIZE
+# =============================================================================
+LARGE_RESULT_SET_SIZE = 20_000
+
+SKIP_JSON = pytest.mark.skip_for_json_result_set(reason="GEOMETRY type is not supported in JSON result format")
```
sfc-gh-asolarski: if we skip all in the file, then we don't really need this decorator

sfc-gh-asolarski: I mean - we don't need to apply it to all classes

sfc-gh-asolarski (approved):

PR #940, opened 2026-04-23
Add VECTOR type e2e test suite

sfc-gh-pczajka:
## Summary
- Add comprehensive e2e tests for VECTOR type: type casting (INT and FLOAT subtypes), literal selects, NULL handling, table operations, multi-chunk downloads, and JSON result format
- Add `vector.feature` Gherkin definitions (9 scenarios)

## Test plan
- [ ] All 9 VECTOR tests pass on universal driver (currently blocked by pre-existing protobuf bug)
- [ ] All 9 VECTOR tests pass on reference driver
- [ ] Test format validator passes

### DIFF
--- a/python/tests/e2e/types/test_vector.py
+++ b/python/tests/e2e/types/test_vector.py
@@ -48,10 +48,9 @@ LARGE_RESULT_SET_SIZE = 20_000
 # VECTOR(FLOAT) uses 32-bit floats — need relaxed tolerance vs 64-bit assert_floats_equal.
 FLOAT32_APPROX = {"rel": 1e-6}
 
-SKIP_JSON = pytest.mark.skip_for_json_result_set(reason="VECTOR type is not supported in JSON result format")
+pytestmark = pytest.mark.skip_for_json_result_set(reason="VECTOR type is not supported in JSON result format")
 
 
-@SKIP_JSON
 class TestVectorTypeCasting:
     """Tests for VECTOR type casting to appropriate type."""
 
@@ -71,7 +70,6 @@ class TestVectorTypeCasting:
         assert_type(result[1], float)
 
 
-@SKIP_JSON
 class TestVectorLiteral:
     """Tests for VECTOR type using SELECT with literals (no tables)."""
 
@@ -161,7 +159,6 @@ class TestVectorLiteral:
         assert result[0] == pytest.approx(expected, **FLOAT32_APPROX)
 
 
-@SKIP_JSON
 class TestVectorTable:
     """Tests for VECTOR type using table operations."""
 
@@ -223,7 +220,6 @@ class TestVectorTable:
         assert rows[2][2] is None
 
 
-@SKIP_JSON
 class TestVectorMultipleChunks:
     """Tests for VECTOR type with multiple chunks downloading."""
 
--- a/python/tests/e2e/types/test_vector.py
+++ b/python/tests/e2e/types/test_vector.py
@@ -16,9 +16,32 @@ from .utils import assert_type
 
 
 # =============================================================================
+# INT VECTOR TEST VALUES
+# =============================================================================
+INT_VEC_3D = [1, 2, 3]
+INT_VEC_3D_B = [10, 20, 30]
+
+# =============================================================================
+# FLOAT VECTOR TEST VALUES
+# =============================================================================
+FLOAT_VEC_3D = [1.5, 2.5, 3.5]
+FLOAT_VEC_5D = [1.1, 2.2, 3.3, 4.4, 5.5]
+FLOAT_VEC_5D_B = [10.5, 20.5, 30.5, 40.5, 50.5]
+
+# =============================================================================
+# LARGE DIMENSION TEST
+# =============================================================================
+LARGE_DIMENSION_SIZE = 256
+
+# =============================================================================
 # LARGE RESULT SET SIZE
 # =============================================================================
 LARGE_RESULT_SET_SIZE = 20_000
+
+
+def _vec_sql(values: list, subtype: str) -> str:
+    """Build a VECTOR SQL literal, e.g. ``[1, 2, 3]::VECTOR(INT, 3)``."""
+    return f"{values}::VECTOR({subtype}, {len(values)})"
 
 
 class TestVectorTypeCasting:
@@ -29,13 +52,13 @@ class TestVectorTypeCasting:
         pass
 
         # When Query "SELECT [1, 2, 3]::VECTOR(INT, 3), [1.5, 2.5, 3.5]::VECTOR(FLOAT, 3)" is executed
-        sql = "SELECT [1, 2, 3]::VECTOR(INT, 3), [1.5, 2.5, 3.5]::VECTOR(FLOAT, 3)"
+        sql = f"SELECT {_vec_sql(INT_VEC_3D, 'INT')}, {_vec_sql(FLOAT_VEC_3D, 'FLOAT')}"
         result = execute_query(sql, single_row=True)
 
         # Then All values should be returned as appropriate type
         assert_type(result, list)
-        assert result[0] == [1, 2, 3]
-        assert result[1] == pytest.approx([1.5, 2.5, 3.5])
+        assert result[0] == INT_VEC_3D
+        assert result[1] == pytest.approx(FLOAT_VEC_3D)
 
 
 class TestVectorLiteral:
@@ -70,25 +93,26 @@ class TestVectorLiteral:
         pass
 
         # When Query generating a 256-dimension FLOAT vector is executed
-        values = ", ".join(str(float(i)) for i in range(256))
-        sql = f"SELECT [{values}]::VECTOR(FLOAT, 256)"
+        expected = [float(i) for i in range(LARGE_DIMENSION_SIZE)]
+        values = ", ".join(str(v) for v in expected)
+        sql = f"SELECT [{values}]::VECTOR(FLOAT, {LARGE_DIMENSION_SIZE})"
         result = execute_query(sql, single_row=True)
 
         # Then Result should contain a 256-element float vector
         assert isinstance(result[0], list)
-        assert len(result[0]) == 256
-        assert result[0] == pytest.approx([float(i) for i in range(256)])
+        assert len(result[0]) == LARGE_DIMENSION_SIZE
+        assert result[0] == pytest.approx(expected)
 
     def test_should_handle_null_vector_values_from_literals(self, execute_query):
         # Given Snowflake client is logged in
         pass
 
         # When Query "SELECT [1, 2, 3]::VECTOR(INT, 3), NULL::VECTOR(INT, 3), NULL::VECTOR(FLOAT, 3)" is executed
-        sql = "SELECT [1, 2, 3]::VECTOR(INT, 3), NULL::VECTOR(INT, 3), NULL::VECTOR(FLOAT, 3)"
+        sql = f"SELECT {_vec_sql(INT_VEC_3D, 'INT')}, NULL::VECTOR(INT, 3), NULL::VECTOR(FLOAT, 3)"
         result = execute_query(sql, single_row=True)
 
         # Then Result should contain [[1, 2, 3], NULL, NULL]
-        assert result[0] == [1, 2, 3]
+        assert result[0] == INT_VEC_3D
         assert result[1] is None
         assert result[2] is None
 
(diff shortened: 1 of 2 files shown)


python/tests/e2e/types/test_vector.py
```diff
@@ -0,0 +1,233 @@
...
+        pass
+
+        # When Query "SELECT [1, 2, 3]::VECTOR(INT, 3), [1.5, 2.5, 3.5]::VECTOR(FLOAT, 3)" is executed
+        sql = f"SELECT {_vec_sql(INT_VEC_3D, 'INT')}, {_vec_sql(FLOAT_VEC_3D, 'FLOAT')}"
+        result = execute_query(sql, single_row=True)
+
+        # Then All values should be returned as appropriate type
+        assert_type(result, list)
+        assert result[0] == INT_VEC_3D
+        assert result[1] == pytest.approx(FLOAT_VEC_3D)
```
sfc-gh-pczajka: For all tests: use assert_floats_equal from [utils.py](http://utils.py)

python/tests/e2e/types/test_vector.py
```diff
@@ -0,0 +1,233 @@
...
+
+class TestVectorLiteral:
+    """Tests for VECTOR type using SELECT with literals (no tables)."""
+
+    @pytest.mark.parametrize(
+        "query_value, expected_value",
+        [
+            ("[1, 3, -5]::VECTOR(INT, 3)", [1, 3, -5]),
+            ("[40, 1234567]::VECTOR(INT, 2)", [40, 1234567]),
+            ("[1.8, -3.4, 6.7, 0, 2.3]::VECTOR(FLOAT, 5)", [1.8, -3.4, 6.7, 0.0, 2.3]),
```
sfc-gh-pczajka: Whuy didn't you use defined constants instead?

python/tests/e2e/types/test_vector.py
```diff
@@ -0,0 +1,233 @@
...
+        )
+        rows = execute_query(sql)
+
+        # Then All 20000 rows should be fetched with valid 3-element integer vectors
+        assert len(rows) == LARGE_RESULT_SET_SIZE
+        vec_values = [row[1] for row in rows]
+        assert_type(vec_values, list)
+        for val in vec_values:
+            assert len(val) == 3
+            assert all(isinstance(v, int) for v in val)
```
sfc-gh-pczajka: This check is not exhausive. Please use assert_sequential_values from utils.py

tests/definitions/shared/types/vector.feature
```diff
@@ -0,0 +1,91 @@
+@python @core_not_needed
+Feature: VECTOR type support
+  # Snowflake VECTOR type stores fixed-size arrays of numeric values.
+  # Subtypes: INT (integer) and FLOAT (32-bit floating-point).
+  # Values are returned as Python lists (list[int] or list[float]).
+  # Maximum dimension: 4096.
```
sfc-gh-pczajka, line 5: Maximum dimension test is not covered in these tests

vector of length 0 is not covered in these tests

tests/definitions/shared/types/vector.feature
```diff
@@ -0,0 +1,91 @@
...
+    Given Snowflake client is logged in
+    When Query generating a 256-dimension FLOAT vector is executed
+    Then Result should contain a 256-element float vector
+
+  # =========================================================================== #
+  #                             NULL handling                                   #
+  # =========================================================================== #
+
+  @python_e2e
+  Scenario: should handle NULL vector values from literals
```
sfc-gh-pczajka: This scenario is covered by other tests

tests/definitions/shared/types/vector.feature
```diff
@@ -0,0 +1,91 @@
...
+    Then Result should contain <subtype> vector <expected_value>
+
+    Examples:
+      | subtype   | query_value                                      | expected_value           |
+      | INT-3d    | [1, 3, -5]::VECTOR(INT, 3)                       | [1, 3, -5]               |
+      | INT-2d    | [40, 1234567]::VECTOR(INT, 2)                    | [40, 1234567]            |
+      | FLOAT-5d  | [1.8, -3.4, 6.7, 0, 2.3]::VECTOR(FLOAT, 5)      | [1.8, -3.4, 6.7, 0, 2.3]|
+
+  @python_e2e
+  Scenario: should select large dimension vector
```
sfc-gh-pczajka: Check how test for special values for strings is constructed. Do similar one. We need to cover:  
\- NULL  
\- dimension 0  
\- max dimension

tests/definitions/shared/types/vector.feature
```diff
@@ -0,0 +1,91 @@
...
+    Given Snowflake client is logged in
+    When Query generating 20000 integer vectors is executed
+    Then All 20000 rows should be fetched with valid 3-element integer vectors
+
+  # =========================================================================== #
+  #                           Parameter binding                                 #
+  # =========================================================================== #
+
+  @python_e2e
+  Scenario: should insert and select vectors using parameter binding
```
sfc-gh-pczajka: There should be two tests: for execute and executemany. Check tests for strings for reference

python/tests/e2e/types/test_vector.py
```diff
@@ -0,0 +1,281 @@
...
+def _assert_float32_equal(actual: list[float], expected: list[float]) -> None:
+    """Assert two float32 vectors are equal within 32-bit float tolerance.
+
+    VECTOR(FLOAT) uses 32-bit floats, so values like 1.8 become 1.7999999523162842.
+    Standard assert_floats_equal uses 64-bit tolerances which are too tight.
+    """
+    assert len(actual) == len(expected), f"Length mismatch: {len(actual)} != {len(expected)}"
+    for i, (a, e) in enumerate(zip(actual, expected)):
+        assert a == pytest.approx(e, rel=1e-6), f"Mismatch at index {i}: expected {e}, got {a}"
+
```
sfc-gh-pczajka: This is unnecessary, utils.py have assert_floats_equal

python/tests/e2e/types/test_vector.py
```diff
@@ -0,0 +1,281 @@
...
+        result = execute_query(sql, single_row=True)
+
+        # Then Result should contain <subtype> vector <expected_value>
+        assert isinstance(result[0], list)
+        if is_float:
+            _assert_float32_equal(result[0], expected_value)
+            assert_type(result[0], float)
+        else:
+            assert result[0] == expected_value
+            assert_type(result[0], int)
```
sfc-gh-pczajka: These assertions should compare all vector, not only first element

python/tests/e2e/types/test_vector.py
```diff
@@ -0,0 +1,281 @@
...
+        max_dim_result = execute_query(f"SELECT [{values}]::VECTOR(FLOAT, {MAX_DIMENSION_SIZE})", single_row=True)
+
+        # Then NULL vectors should return None and max-dimension vector should be valid
+        assert result[0] == INT_VEC_3D
+        assert result[1] is None
+        assert result[2] is None
+        assert isinstance(max_dim_result[0], list)
+        assert len(max_dim_result[0]) == MAX_DIMENSION_SIZE
+        assert_type(max_dim_result[0], float)
+        _assert_float32_equal(max_dim_result[0], expected)
```
sfc-gh-pczajka: Please decouple logic of "normal" checks and max-dimension checks. SELECT "normal" -> assert normal -> SELECT max -> assert max

python/tests/e2e/types/test_vector.py
```diff
@@ -0,0 +1,281 @@
...
+    """Tests for VECTOR type using SELECT with literals (no tables)."""
+
+    LITERAL_TEST_CASES = [
+        ("INT-3d", _vec_sql(INT_VEC_3D_ALT, "INT"), INT_VEC_3D_ALT, False),
+        ("INT-2d", _vec_sql(INT_VEC_2D, "INT"), INT_VEC_2D, False),
+        ("FLOAT-5d", _vec_sql(FLOAT_VEC_5D_ALT, "FLOAT"), FLOAT_VEC_5D_ALT, True),
+    ]
+
+    @pytest.mark.parametrize(
+        "subtype, query_value, expected_value, is_float",
```
sfc-gh-pczajka: "query value" is always `_vec_sql(expected_value, "FLOAT" if is_float else "INT").  
Please refactor

tests/definitions/shared/types/vector.feature
```diff
@@ -0,0 +1,70 @@
+@python @core_not_needed
+Feature: VECTOR type support
+  # Snowflake VECTOR type stores fixed-size arrays of numeric values.
+  # Subtypes: INT (integer) and FLOAT (32-bit floating-point).
+  # Values are returned as Python lists (list[int] or list[float]).
```
sfc-gh-asolarski: this is shared gherkin definition - comments shouldn't be python specific

tests/definitions/shared/types/vector.feature
```diff
@@ -0,0 +1,70 @@
...
+
+    Examples:
+      | subtype   | vec_type | expected_value              |
+      | INT-3d    | INT      | [1, 3, -5]                  |
+      | INT-2d    | INT      | [40, 1234567]               |
+      | FLOAT-5d  | FLOAT    | [1.8, -3.4, 6.7, 0.0, 2.3] |
+
+  @python_e2e
+  Scenario: should select vector special values
+    # Special values: NULL vectors and max-dimension (4096) vector
```
sfc-gh-asolarski: I would split null and max-dimension cases

python/tests/e2e/types/test_vector.py
```diff
@@ -0,0 +1,208 @@
...
+
+# =============================================================================
+# SPECIAL VALUES TEST
+# =============================================================================
+MAX_DIMENSION_SIZE = 4096
+
+# =============================================================================
+# LARGE RESULT SET SIZE
+# =============================================================================
+LARGE_RESULT_SET_SIZE = 20_000
```
sfc-gh-asolarski, line 46: 50/100k maybe?  
Are we sure that with pretty small vectors 20k will cause multiple batches?

python/tests/e2e/types/test_vector.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+def _vec_sql(values: list, subtype: str) -> str:
```
sfc-gh-asolarski: I'm not sure if that's helpful, maybe making those SQL queries more verbose is a bit more easier to comprehend.

python/tests/e2e/types/test_vector.py
```diff
@@ -0,0 +1,208 @@
...
+        assert rows[1][2] == pytest.approx(FLOAT_VEC_3D, **FLOAT32_APPROX)
+        assert rows[2][1] is None
+        assert rows[2][2] is None
+
+
+class TestVectorMultipleChunks:
+    """Tests for VECTOR type with multiple chunks downloading."""
+
+    @pytest.mark.skip_for_json_result_set(
+        reason="Multichunk vector generates dynamic data that may not round-trip identically in JSON format"
```
sfc-gh-asolarski: this is smelly, why can't we handle that?

python/tests/e2e/types/test_vector.py
sfc-gh-pczajka: Yeah, for non-trivial types (non-float) it requires less entries. Experimentally it downloads 3 chunks (1 initial + 2 later). I'd keep it lower, test will be faster

python/tests/e2e/types/test_vector.py
sfc-gh-pczajka: JSON is not supported for VECTOR and GEO*. Will add in follow-up

tests/definitions/shared/types/vector.feature
sfc-gh-asolarski, line 1: we still miss vector of length 0 in tests
Additionally, I think it could be beneficial to add tests for max/min values of specific numbers stored in vector.
Something like "SELECT [-214748648, 2147483647]::VECTOR(INT, 2);" and similarly for float?

sfc-gh-asolarski (changes requested):

tests/definitions/shared/types/vector.feature
sfc-gh-pczajka: Vector of length 0 is invalid according to https://docs.snowflake.com/en/sql-reference/data-types-vector

Added tests with extreme values for int and float

sfc-gh-asolarski (approved):

sfc-gh-asolarski (approved):

PR #941, opened 2026-04-23
Performance smoke tests on branch

sfc-gh-akolodziejczyk:
Performance regression detection for PRs

  Adds automated performance smoke tests that run on every PR. The CI runs performance benchmarks with recorded HTTP traffic for selected SELECT tests and compares results against recent main branch baselines from Benchstore. If a regression >5% is detected, it's re-run to confirm it's not noise.

  When a confirmed regression is found, the performance/regression status blocks the PR. To accept a performance change, add the ACCEPT_PERFORMANCE_CHANGES label — a GitHub Actions workflow immediately overrides the status to success,
  unblocking the merge gate. The label is automatically removed on the next push, so each new commit must be explicitly accepted.

### DIFF
--- a/.github/workflows/accept-performance-changes.yml
+++ b/.github/workflows/accept-performance-changes.yml
@@ -1,0 +1,55 @@
+name: Accept Performance Changes
+
+on:
+  pull_request:
+    types: [labeled, synchronize]
+
+jobs:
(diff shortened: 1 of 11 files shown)


sfc-gh-pcyrek (approved):
lgtm

PR #948, opened 2026-04-23
SNOW-3264508: python - error handler

sfc-gh-asolarski:


sfc-gh-asolarski:
* **#948** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/948?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/948?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python/src/snowflake/connector/cursor/_base.py
```diff
@@ -282,12 +286,12 @@ def arraysize(self, value: int) -> None:
 
     @property
     @pep249
-    def messages(self) -> list[tuple[type[Exception], dict[str, str | bool]]]:
+    def messages(self) -> list[tuple[type[Exception], ErrorValue]]:
```
sfc-gh-turbaszek, line 289: Is this backward compatible? Won't it break linters/type checkers?

sfc-gh-turbaszek (approved):
I'm still skeptical about using `self._handle_error` instead of plain raise.

python/src/snowflake/connector/cursor/_base.py
sfc-gh-asolarski: This is backward compatible — `ErrorValue` resolves to `dict[str, Any]`, which is strictly wider than the old `dict[str, str | bool]`. Existing code that reads or writes `messages` will type-check fine.

The old annotation was actually inaccurate: the dict also carries `errno` (int) and `sqlstate` (str | None), which don't fit `str | bool`. So the new alias is more correct while remaining compatible.

PR #956, opened 2026-04-24
fix(sf_core): map known Snowflake error codes to SQLSTATE in core

sfc-gh-makowalski:
## Summary

Move SQLSTATE classification for `GenericError`/`InternalError` payloads down from the ODBC layer into `sf_core`, so every consumer (ODBC today, JDBC/ADBC tomorrow) gets a consistent `sql_state` on the wire instead of having to re-derive it from English error text.

## Why

Today `odbc/src/api/error.rs` carries `sql_state_from_server_message`, a substring-match heuristic on the server's English message:

```rust
fn sql_state_from_server_message(message: &str) -> SqlState {
    if message.contains("SQL compilation error") { ... }
    else if message.contains("out of representable range") { ... }
    else if message.contains("too long and would be truncated") { ... }
    else { SqlState::GeneralError }
}
```

It exists because some core paths drop the server-supplied `sqlState`:

- `RestError::AsyncQuery { source: SfError::SnowflakeBody { code, .. } }` is mapped in `extract_vendor_info` to `(Some(*code), None)` — code preserved, SQLSTATE thrown away.
- `snowflake_query_status` constructs `QueryFailed { sql_state: None, .. }` even when it has a numeric `code`.

Sniffing English text is fragile (any docs/i18n tweak silently regresses the SQLSTATE) and pushes the same lookup table into every future driver binding.

## What

- New `sf_core::rest::snowflake::sql_state::sql_state_from_code(code: i32) -> Option<&'static str>` mapping the small set of well-known Snowflake error codes to ANSI SQLSTATE strings (1003 → `42000`, 100038 → `22003`, 100078 → `22001`). Documented as a fallback table; codes are added incrementally with tests.
- `extract_vendor_info` now falls back to that lookup whenever the wire response carried a recognised numeric code but no `sqlState`. Server-supplied `sqlState` still wins — the lookup only fills the gap.
- `odbc/src/api/error.rs`: delete `sql_state_from_server_message` and its branch; the ODBC layer now trusts the SQLSTATE forwarded by `sf_core` and returns `HY000` for genuinely opaque errors.
- Tests updated to encode the new contract (sf_core supplies the SQLSTATE, ODBC trusts it) plus a regression guard ensuring the ODBC layer never re-introduces message-text sniffing.

## Files

- `sf_core/src/rest/snowflake/sql_state.rs` — new lookup module + unit tests.
- `sf_core/src/rest/snowflake/mod.rs` — declare the module.
- `sf_core/src/protobuf/apis/database_driver_v1/converter.rs` — wire lookup into `extract_vendor_info`; add tests for sync/async paths and override semantics.
- `odbc/src/api/error.rs` — drop heuristic, update tests, add regression guard.

## Test plan

- [x] `cargo test -p sf_core --lib` — 572 passed.
- [x] `cargo test -p odbc --lib` — 1021 passed.
- [x] `cargo clippy -p sf_core -p odbc --tests -- -D warnings` — clean.
- [ ] CI green.
- [ ] Spot-check on an integration test that intentionally trips a `100038` / `100078` server error via the async-poll path and verify `SQLGetDiagRec` now returns `22003` / `22001` instead of `HY000`.

## Follow-ups

- Add more codes to `sql_state_from_code` as production telemetry surfaces them (e.g. lock-timeout, statement-cancelled, division-by-zero). Each addition gets a unit test alongside.
- Consider plumbing `vendor_code` from `DriverException` into `CoreProtobufError::Application` so the ODBC layer can also return it via `SQLGetDiagField(SQL_DIAG_NATIVE)` for non-login errors. Out of scope here.

### DIFF
--- a/sf_core/src/rest/snowflake/sql_state.rs
+++ b/sf_core/src/rest/snowflake/sql_state.rs
@@ -1,0 +1,68 @@
+//! Fallback Snowflake error-code → ANSI SQLSTATE lookup.
+//!
+//! The Snowflake server normally returns an explicit `sqlState` field on the
+//! query-failure response (see `query_response::Data::sql_state`), and that
+//! value is forwarded verbatim through `RestError::QueryFailed::sql_state`
+//! and onto `DriverException::sql_state`.
+//!
+//! However, a few code paths receive only the numeric error code and never an
+//! `sqlState` string — for example:
+//!
+//! - The async polling path constructs `RestError::QueryFailed { sql_state:
+//!   None, .. }` from `SfError::SnowflakeBody { code, .. }` (see
+//!   `apis::database_driver_v1::error::map_async_query_error`).
+//! - The query-monitoring response (`snowflake_query_status`) emits
+//!   `error_code` without a SQLSTATE.
+//!
+//! Without a SQLSTATE on the wire, downstream consumers (ODBC, JDBC, ADBC)
+//! lose the ability to classify the error and fall back to a generic
+//! `HY000` / `08000`. This table fills that gap by mapping the small set of
+//! well-known Snowflake error codes whose ANSI SQLSTATE is stable to the
+//! corresponding string, so the rest of the stack can keep treating
+//! `sql_state` as the single source of truth.
+//!
+//! The table is intentionally narrow: only codes whose mapping is unambiguous
+//! and stable across server versions belong here. Add a new entry when a new
+//! code is observed in production telemetry and confirmed against the
+//! Snowflake error registry — and add a unit test alongside it.
+
+/// Look up the ANSI SQLSTATE string for a Snowflake server error code.
+///
+/// Returns `None` for codes that have no canonical SQLSTATE mapping; callers
+/// should treat that as "leave `sql_state` unset" and let the consumer apply
+/// its own default (typically `HY000`).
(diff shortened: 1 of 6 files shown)


sf_core/src/rest/snowflake/sql_state.rs
```diff
@@ -0,0 +1,68 @@
...
+        // a safety net for paths (async poll, monitoring) where the server
+        // omits the field.
+        1003 => Some("42000"),
+        // Numeric value out of representable range for the target column type.
+        // e.g. "Number out of representable range: type FIXED[SB2](3,0), value 99999".
+        100038 => Some("22003"),
+        // String length exceeds the column's declared maximum and would be
+        // truncated. e.g. "String 'hello world' is too long and would be truncated".
+        100078 => Some("22001"),
+        _ => None,
```
sfc-gh-pbulawa: We should catch the unknown error codes, e.g. with tracing, so there will be a mention at least in the logs about the missing error code.

sfc-gh-pbulawa:
There's no integration-level test that verifies the full chain: `extract_vendor_info` populates the SQLSTATE → protobuf conversion → ODBC `to_sql_state()` renders it

sfc-gh-pbulawa (changes requested):

sfc-gh-pbulawa (approved):

PR #966, opened 2026-04-24
SNOW-3264508: python - ErrorHandlerMixin for PEP 249 error routing

sfc-gh-asolarski:
### TL;DR

Replaces the ad-hoc `_handle_error` pattern on `Connection` and `SnowflakeCursorBase` with a centralized `ErrorHandlerMixin` that automatically routes PEP 249 errors from public methods through the errorhandler chain.

### What changed?

A new `ErrorHandlerMixin` class (in `_internal/errorhandler.py`) wraps public methods of any subclass at class-creation time via `__init_subclass__`. When a public method raises a `snowflake.connector.errors.Error`, the mixin intercepts it, calls `route_exception` to invoke the PEP 249 errorhandler chain (populating `messages`, calling any user-supplied `errorhandler`), and then re-raises the original exception. A `ContextVar` (`_errorhandler_active`) prevents double-routing when a wrapped public method calls another.

- `Connection`, `SnowflakeCursorBase`, and `ResultBatch` now inherit `ErrorHandlerMixin` and expose `_errorhandler_connection` / `_errorhandler_cursor` properties instead of implementing `_handle_error`.
- All `self._handle_error(SomeError, ...)` call sites are replaced with plain `raise SomeError(msg=..., errno=...)`.
- `write_pandas` (a free function) wraps its body in `try/except Error` and calls `route_exception(conn, None, exc)` directly.
- The `_handle_error` methods on `Connection` and `SnowflakeCursorBase` are removed entirely.
- Generator methods, private methods, properties, classmethods, and staticmethods are intentionally excluded from wrapping.

### Why make this change?

The previous `_handle_error` helper was duplicated on every class that needed PEP 249 error routing, required callers to pass error metadata through an intermediate dict, and made it easy to forget to route errors correctly. Centralizing the logic in `ErrorHandlerMixin` ensures consistent behavior across all public API surfaces, removes boilerplate, and makes it impossible to accidentally bypass the errorhandler chain when raising from a public method.

### DIFF
--- a/python/src/snowflake/connector/_internal/errorhandler.py
+++ b/python/src/snowflake/connector/_internal/errorhandler.py
@@ -1,8 +1,9 @@
-"""PEP 249 error-handler routing for ``@with_errorhandler``-decorated classes.
+"""PEP 249 error-handler routing via ``ErrorHandlerMixin``.
 
 How to raise errors in the driver:
-- Classes (Connection, Cursor, ResultBatch): decorate with ``@with_errorhandler``
-  and inherit ``ErrorHandlerMixin``.  Then use plain ``raise``.
+- Classes (Connection, Cursor, ResultBatch): inherit ``ErrorHandlerMixin``.
+  Public methods are wrapped automatically via ``__init_subclass__``.
+  Then use plain ``raise``.
 - Free functions with a ``conn`` argument (e.g. ``write_pandas``): wrap the body
   in ``try/except Error`` and call ``route_exception(conn, None, exc)``.
 """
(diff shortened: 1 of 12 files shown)


sfc-gh-asolarski:
* **#966** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/966?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/966?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python/src/snowflake/connector/_internal/errorhandler.py
```diff
@@ -0,0 +1,113 @@
...
+    return {
+        "msg": exc.raw_msg,
+        "errno": exc.errno,
+        "sqlstate": exc.sqlstate,
+        "sfqid": exc.sfqid,
+        "query": exc.query,
+    }
+
+
+class ErrorHandlerMixin:
```
sfc-gh-turbaszek: Given we have the mixin, can we use it to perform the logic of `with_errorhandler`? This way we will need only one thing instead of two. WDYT?

sfc-gh-asolarski: good idea - simplified

sfc-gh-turbaszek (approved):

PR #970, opened 2026-04-25
Add generated ConnectionConfig dataclass from Rust PARAM_DEFS

sfc-gh-turbaszek:
## Summary
- New Rust binary `param_defs_export` generates a Python `ConnectionConfig` dataclass from the canonical `PARAM_DEFS` registry in `sf_core`, establishing a single source of truth for connection parameters
- `ConnectionConfig` handles all config building, validation, and normalisation (`from_kwargs`, `from_connection_args`, `to_options`, `to_proto_options`, `redacted_options`)
- `Connection.__init__` refactored to delegate to `ConnectionConfig`, storing `self.config` — backward compatible via `Connection(**kwargs)`

## Test plan
- [ ] `hatch run dev.py3.13:unit tests/unit/test_connection_config.py` — 24 new tests for ConnectionConfig
- [ ] `hatch run dev.py3.13:unit tests/unit/test_connection.py` — existing connection tests pass
- [ ] `hatch run dev.py3.13:unit tests/unit/test_cursor.py` — cursor numpy flag tests updated
- [ ] `hatch run precommit:fix` — ruff + mypy pass clean


python/src/snowflake/connector/connection.py
```diff
@@ -878,11 +795,20 @@ def use_openssl_only(self) -> bool:
     @property
     def arrow_number_to_decimal(self) -> bool:
         """Whether to convert Arrow numeric types to Python ``Decimal`` instead of ``float``."""
+        # Read the backward-compat private attribute so legacy code that mutates
+        # ``self._arrow_number_to_decimal`` directly (rather than via the setter)
+        # is still observed.  ``self.config.arrow_number_to_decimal`` is the
+        # canonical source set in ``__init__``; the two are kept in sync by the
+        # setter below.
         return self._arrow_number_to_decimal
```
sfc-gh-asolarski: aren't we overcomplicate stuff there?
isn't it enough to have `self.config.arrow_number_to_decimal`, anyway property `_arrow_number_to_decimal` was marked private and setting it directly (`conn._arrow_number_to_decimal  = True`) would be considered violation 🤔

sfc-gh-turbaszek: Good point — agreed it was over-engineered. Dropped the `_arrow_number_to_decimal` mirror, the property now reads/writes `self.config.arrow_number_to_decimal` directly (with `bool()` to coerce `None` → `False`), matching the pattern we already use for `numpy`. Test that introspected the private attr is gone; `test_arrow_number_to_decimal_default_is_false` keeps the regression coverage via the public property.

sfc-gh-asolarski (approved):

PR #982, opened 2026-04-27
NO-SNOW: Account for intermittent failures in reference driver cross thread cancel behaviour

sfc-gh-ddas:
### TL;DR

Documents and adds test coverage for the behavior difference where cross-thread `SQLCancel`/`SQLCancelHandle` intermittently returns `SQL_ERROR` with `HY008` on the old driver instead of `SQL_SUCCESS` as required by the ODBC spec.

### What changed?

A new behavior difference entry (BD#47) was added to `BehaviorDifferences.yaml` describing the old driver bug where `SQLCancel` and `SQLCancelHandle`, when called from a different thread to cancel an in-progress `SQLExecDirect`, could intermittently return `SQL_ERROR` with SQLSTATE `HY008` instead of `SQL_SUCCESS`. This occurs because the Simba SDK DSII's `CancelExecute()` calls `Statement::cancel()` without a try-catch, allowing exceptions to propagate when the server's abort-request endpoint responds with a non-success status.

The cross-thread cancel tests in `cancel_tests.cpp` and `cancel_handle_tests.cpp` were updated to use `OLD_DRIVER_ONLY`/`NEW_DRIVER_ONLY` guards. The old driver assertion accepts either `SQL_SUCCESS` or `SQL_ERROR` (verifying `HY008` when it is an error), while the new driver assertion strictly requires `SQL_SUCCESS` per the ODBC specification.

### How to test?

Run the existing cross-thread cancel test cases:
- `SQLCancel: Cross-thread cancel interrupts` in `cancel_tests.cpp`
- `SQLCancelHandle: Cross-thread cancel in` in `cancel_handle_tests.cpp`

On the new driver, both tests should pass with `SQL_SUCCESS` returned from the cancel call. On the old driver, the tests accept either `SQL_SUCCESS` or `SQL_ERROR` with `HY008`.

### Why make this change?

The ODBC specification explicitly states that when `SQLCancel` successfully cancels a function running in another thread, the driver must not return `SQL_ERROR` with `HY008` — only the canceled function itself should report that error. The old driver violated this contract intermittently. This change documents the known deviation and ensures the new driver enforces the correct behavior.

### DIFF
--- a/odbc_tests/BehaviorDifferences.yaml
+++ b/odbc_tests/BehaviorDifferences.yaml
@@ -497,3 +497,17 @@ behavior_differences:
       Per the MS ODBC specification ("C to SQL: Binary" appendix) the driver checks that BufferLength equals the SQL data length (sizeof(SQL_NUMERIC_STRUCT) = 19 for DECIMAL/NUMERIC, sizeof(SQL_TIME_STRUCT) = 6 for TIME, sizeof(SQL_TIMESTAMP_STRUCT) = 16 for TIMESTAMP) and forwards the raw bytes to the server. Buffers of any other length are rejected with SQLSTATE 22003, matching the spec's "Byte length of data <> SQL data length -> 22003" rule.
     impact: |
       Applications that bind a typed C struct as SQL_C_BINARY now have a portable way to insert these SQL types without separate SQL_C_NUMERIC / SQL_C_TYPE_TIMESTAMP / etc. bindings. See SNOW-3031774 (PR #833 for the Rust implementation, PR #834 for E2E coverage).
+
+  47:
+    name: "Cross-thread SQLCancel/SQLCancelHandle returns SQL_ERROR with HY008 instead of SQL_SUCCESS"
+    status: unknown
+    type: bug
+    reviewed: false
+    old_driver_behavior: |
+      When SQLCancel or SQLCancelHandle is called from a different thread to cancel an in-progress SQLExecDirect, the cancel function intermittently returns SQL_ERROR with SQLSTATE HY008 (NativeError 604, "SQL execution canceled") instead of SQL_SUCCESS. This happens because the driver's internal CancelExecute() calls Statement::cancel() without a try-catch. When the server's abort-request endpoint responds with a non-success status (e.g. the query was already canceled before the REST abort completed), cancel() throws and the exception propagates to the caller.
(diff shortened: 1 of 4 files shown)


sfc-gh-ddas:
* **#982** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/982?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/982?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

odbc_tests/BehaviorDifferences.yaml
```diff
@@ -497,3 +497,17 @@ behavior_differences:
...
     impact: |
       Applications that bind a typed C struct as SQL_C_BINARY now have a portable way to insert these SQL types without separate SQL_C_NUMERIC / SQL_C_TYPE_TIMESTAMP / etc. bindings. See SNOW-3031774 (PR #833 for the Rust implementation, PR #834 for E2E coverage).
+
+  47:
+    name: "Cross-thread SQLCancel/SQLCancelHandle returns SQL_ERROR with HY008 instead of SQL_SUCCESS"
+    status: unknown
+    type: bug
+    reviewed: false
+    old_driver_behavior: |
+      When SQLCancel or SQLCancelHandle is called from a different thread to cancel an in-progress SQLExecDirect, the cancel function intermittently returns SQL_ERROR with SQLSTATE HY008 (NativeError 604, "SQL execution canceled") instead of SQL_SUCCESS. This happens because the Simba SDK DSII's CancelExecute() calls Statement::cancel() without a try-catch. When the server's abort-request endpoint responds with a non-success status (e.g. the query was already canceled by the SDK's internal cancel flag before the REST abort completed), cancel() throws and the exception propagates to the caller.
```
sfc-gh-makowalski: I don't think we should be mentioning Simba specifically here. Lets just drop it.

sfc-gh-makowalski (dismissed):

sfc-gh-merbel (approved):

odbc_tests/BehaviorDifferences.yaml
sfc-gh-ddas: Removed

PR #998, opened 2026-04-28
SNOW-3098562: Default statement classifier to NoResult to match snowflake-odbc

sfc-gh-rkowalski:
Align `QueryType::result_kind()` with the reference driver (`snowflake-odbc SFResults::SetQueryResultSet`) by defaulting to `NoResult` instead of `Cursor` for any statement type that is not whitelisted for a cursor and not DML.

Two cases were incorrectly returning Cursor:
  - MULTI_STMT parents (0xA000) — regressed in PR #974 commit c3e5b07 when is_known_no_result was inlined and its MULTI_STMT branch silently dropped.
  - SYSCMD subtypes (ALTER SESSION SET, USE, …) and unknown / future ids — pre-existing: the function's final else returned Cursor, the reverse of the reference driver.

Both produced a phantom cursor over an empty reader, which crashed ODBC consumers that trust the cursor shape (pyodbc segfaults; DTM odbc_beta → Erlang tdc_msg_utils:row_to_binary/6 function_clause).


odbc_tests/tests/e2e/query/multistatement_cursor_shape.cpp
```diff
@@ -0,0 +1,117 @@
...
+  CursorShape shape{};
+  SQLRETURN ret = SQLNumResultCols(stmt.getHandle(), &shape.num_cols);
+  REQUIRE_ODBC(ret, stmt);
+  ret = SQLRowCount(stmt.getHandle(), &shape.row_count);
+  REQUIRE_ODBC(ret, stmt);
+  return shape;
+}
+}  // namespace
+
+TEST_CASE("should report correct cursor shape for each result set in a DDL + DML + DDL batch",
```
sfc-gh-pbulawa, line 35: Missing gherkin files. Please add them

sfc-gh-pbulawa (changes requested):

odbc_tests/tests/e2e/query/multistatement_cursor_shape.cpp
sfc-gh-rkowalski: I've added missing gherkin file.

sfc-gh-pbulawa (approved):

PR #965, opened 2026-04-24
SNOW-3282031: add JSON chunks support for distributed fetch

sfc-gh-asolarski:

PR #1000, opened 2026-04-28
SNOW-3282031: support distributed fetch for multistatements & clean up python wrapper

sfc-gh-asolarski:


sfc-gh-asolarski:
* **#1000** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1000?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1000?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#965** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/965?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

python/src/snowflake/connector/_internal/statement_utils.py
```diff
@@ -20,8 +20,26 @@
     from ..connection import Connection
 
 
+def new_stmt(connection: Connection) -> StatementHandle:
```
sfc-gh-turbaszek, line 24: How about wrapping `StatementHandle` into `StatementManager`. This way we don't need to push connection handle every time. WDYT?

python/src/snowflake/connector/_internal/statement_utils.py
```diff
@@ -115,7 +108,28 @@ def get_stream_ptr(result: DatabaseFetchChunkResponse | PrepareResult | ResultSe
...
+    Args:
+        result: A ``PrepareResult`` or ``ResultSetDescriptor``,
+            or ``None`` if no result is available.
+
+    Returns:
+        A five-character SQLSTATE string for warnings/errors, or ``None``
+        on success or when *result* is ``None``.
+    """
+    sql_state = result.sql_state if result else None
+    if sql_state and sql_state != "00000":
```
sfc-gh-turbaszek: Can we use enum/constant for `00000`?

sfc-gh-turbaszek (approved):

python/src/snowflake/connector/_internal/statement_utils.py
sfc-gh-asolarski: done

python/src/snowflake/connector/_internal/statement_utils.py
sfc-gh-asolarski: That's a good idea. However, the way we handle statements is currently under evaluation and may be changed in the future (which may cause storing statement handle in cursor unnecessary).  
So, it'll be done eventually as a follow up.

PR #1004, opened 2026-04-29
feat(bindparam): route Snowflake vendor TIMESTAMP type codes (LTZ/NTZ) for explicit subtype binding

sfc-gh-makowalski:
## Summary

Adds the Snowflake-specific vendor TIMESTAMP type codes as accepted `ParameterType` values for `SQLBindParameter`, mirroring the legacy 3.16.0 driver (`Source/sf_odbc.h`). The standard `SQL_TYPE_TIMESTAMP` (93) cannot distinguish the three subtypes on the bind side, so applications that need explicit LTZ or TZ routing have no spec-clean alternative; the vendor codes give them an additive opt-in.

| Code | Constant | Routes to | Notes |
|---|---|---|---|
| 93   | `SQL_TYPE_TIMESTAMP`     | `TIMESTAMP_NTZ` | Default; back-compat with Tableau / Excel / Power BI |
| 2002 | `SQL_SF_TIMESTAMP_NTZ`   | `TIMESTAMP_NTZ` | Explicit vendor opt-in |
| 2000 | `SQL_SF_TIMESTAMP_LTZ`   | `TIMESTAMP_LTZ` | Server interprets wall-clock in session timezone |
| 2001 | `SQL_SF_TIMESTAMP_TZ`    | (rejected, SQLSTATE `07006`) | Lifted in #1005 once offset round-trip lands |

The `SQL_SF_TIMESTAMP_TZ` (2001) rejection is intentional for now: faithfully binding `TIMESTAMP_TZ` requires preserving the offset on the wire, which lands in #1005. Returning a clear `Restricted data type attribute violation` error is safer than silently dropping the offset and storing the wrong instant.

## Notes on the values

- These codes are **input-only**. `SQLDescribeCol` and `SQLColAttribute(SQL_DESC_CONCISE_TYPE)` continue to return the standard `SQL_TYPE_TIMESTAMP` (93) for all three subtypes per the MS ODBC 3.x spec; applications distinguish NTZ/LTZ/TZ via `SQL_DESC_TYPE_NAME`.
- Strictly per the MS spec, driver-specific SQL data types should fall in `0x4000`–`0x7FFF`. The legacy 3.16.0 driver picked `2000`–`2002` (which sit in the ODBC-reserved range) before that guidance was followed; we accept those values for application compatibility but do not introduce new codes outside the reserved range.

## Test plan

- Rust unit coverage for the new `make_converter` arms.
- e2e `timestamp_vendor_bindparam.cpp` covers NTZ and LTZ positive round-trips and the explicit TZ rejection (`07006`). The TZ test is `SKIP`'d pending #1005 (which adds offset-preserving TZ binding and unskips it).
- [ ] CI green


sfc-gh-makowalski:
* **#1004** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1004?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1004?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

odbc_tests/tests/e2e/types/timestamp_vendor_bindparam.cpp
sfc-gh-rkowalski, line 1: Aren't we missing feature file?

odbc/src/conversion/param_binding.rs
```diff
@@ -231,6 +237,30 @@ fn make_converter(
             },
         })),
 
+        // Snowflake-specific LTZ vendor code: drives the binding as
+        // TIMESTAMP_LTZ on the wire so the server interprets a naive datetime
+        // in the session timezone instead of as a wall-clock NTZ value.
+        SQL_SF_TIMESTAMP_LTZ => Ok(Box::new(JsonParamConverter {
+            snowflake_type: SnowflakeTimestampLtz { scale: 9 },
+        })),
```
sfc-gh-rkowalski: I think this arm doesn't actually deliver what the commit message promises. The logical type tag flips to `TimestampLtz`, but `SnowflakeTimestampLtz` shares `write_json` with the NTZ variant via `impl_snowflake_timestamp!(... standard ...)`, and that path emits `value.and_utc().timestamp_nanos_opt()` — i.e. it treats the wall-clock as if it were already UTC.

So if an app binds `2024-03-15 14:30:45` with session `TIMEZONE='America/Los_Angeles'`, we send the instant `2024-03-15T14:30:45Z` and the server renders it as `2024-03-15 07:30:45-07:00`. That's the opposite of what we want, and not what the legacy 3.16.0 driver does (legacy sends a wall-clock string and lets the server parse it in the session TZ).

The e2e test doesn't catch this because it pins `TIMEZONE='UTC'`, which is exactly the configuration where LTZ and NTZ collapse to the same observable — `CHECK(ts_out.hour == 14)` would still pass if this arm silently routed through `TimestampNtz`.

Two ways out, pick one:

1. Make the bind actually preserve the wall-clock — emit a literal string (`"YYYY-MM-DD HH:MM:SS.fff..."`) so the server parses it in the session TZ, matching legacy. Then add an e2e test that runs with a non-UTC `TIMEZONE` and asserts the stored epoch matches the local-time-in-session-TZ interpretation.
2. Keep the current behavior, but be honest in the commit message and the comment on this arm: "binds as the same UTC instant the wall-clock denotes; the server displays it in session TZ". If we go this route, LTZ becomes observably identical to NTZ for non-UTC sessions, which makes me question whether we should ship the LTZ vendor code at all in this PR.

Leaning toward (1) — that's the contract the commit message is selling and the one downstream apps are going to expect.

odbc/src/api/types/odbc_types.rs
```diff
@@ -608,6 +626,9 @@ impl TryFrom<sql::SmallInt> for SqlType {
             111 => Ok(SqlType::IntervalHourToMinute),
             112 => Ok(SqlType::IntervalHourToSecond),
             113 => Ok(SqlType::IntervalMinuteToSecond),
+            2000 => Ok(SqlType::SqlSfTimestampLtz),
+            2001 => Ok(SqlType::SqlSfTimestampTz),
+            2002 => Ok(SqlType::SqlSfTimestampNtz),
```
sfc-gh-rkowalski, line 631: Once 2000/2001/2002 round-trip through `try_from` they get stored verbatim into `IpdRecord::sql_data_type` (in `bind_parameter`), and `SQLDescribeParam` then echoes them back to the app via `*data_type_ptr = ipd_rec.sql_data_type.0`.

That directly contradicts the commit message's "input-only" framing. An app that bound with `SQL_SF_TIMESTAMP_LTZ` and then calls `SQLDescribeParam` will see `2000` instead of the spec-mandated `93`, and any consumer that switches on 93 in describe-param falls through to an unknown-type branch.

Two options, in order of preference:

- **Preferred:** in `bind_parameter`, normalize all three vendor codes to `SQL_TYPE_TIMESTAMP` (93) when writing `IpdRecord::sql_data_type`, and stash the subtype on a separate field (something like `IpdRecord::sf_subtype: Option<{Ntz,Ltz,Tz}>`) for the converter dispatch. That keeps `SQLGetDescField(IPD, SQL_DESC_TYPE)` honest too.
- **Cheaper:** map 2000/2001/2002 → 93 on the way out of `describe_param`.

Either way please add a unit test that binds with each vendor code and asserts `describe_param` returns 93. Right now we have nothing pinning the input-only contract.

odbc_tests/tests/e2e/types/timestamp_vendor_bindparam.cpp
```diff
@@ -0,0 +1,165 @@
...
+  REQUIRE_ODBC_SUCCESS(ret, stmt);
+
+  SQL_TIMESTAMP_STRUCT ts_in = {};
+  ts_in.year = 2024;
+  ts_in.month = 3;
+  ts_in.day = 15;
+  ts_in.hour = 14;
+  ts_in.minute = 30;
+  ts_in.second = 45;
+  ts_in.fraction = 0;
```
sfc-gh-rkowalski: Both the NTZ test here and the LTZ test below use `fraction = 0`. The arm in `make_converter` hardcodes `scale: 9`, so a regression that silently truncated to whole seconds would still pass these tests — `timestamp_bind_fetch.cpp` uses `123456789` for exactly that reason.

Two-line change: set `fraction = 123456789` in both tests and assert it round-trips. Locks in the scale-9 contract at the integration level.

sfc-gh-rkowalski (changes requested):
Few commens left

odbc_tests/tests/e2e/types/timestamp_vendor_bindparam.cpp
sfc-gh-makowalski: Nope, we should add feature files only when the tests/features are shared between wrappers. Here it's only ODBC

sfc-gh-makowalski: No feature file for this one. Per `.cursor/rules/datatype_gherkins.mdc`, Gherkin features under `tests/definitions/shared/types` document **shared cross-wrapper** datatype behaviour. The `SQL_SF_TIMESTAMP_*` vendor codes are an ODBC-only binding mechanism (no JDBC/Python equivalent — those wrappers expose the same routing through `setObject(idx, val, JDBCType.TIMESTAMP_*)` etc.) and the diff here is contained to `odbc/` + `odbc_tests/` (zero core-crate touches). So this stays ODBC-local.

I extended the existing `timestamp_vendor_bindparam.cpp` with the cases that actually move us forward: a `SQLDescribeParam` round-trip per vendor code (pinning the input-only contract) and a non-UTC LTZ wall-clock test (Asia/Kolkata, UTC+05:30 no DST). Pushed in fa107b963.

odbc/src/conversion/param_binding.rs
sfc-gh-makowalski: Good catch — and you're absolutely right, the macro-shared `write_json` was treating the bound naive datetime as already-UTC, which is the bug you described.

Implemented option (1) — wall-clock string emission — in fa107b963. Concretely:

- Split `impl_snowflake_timestamp!` into a new `ltz_wallclock_string` arm that generates the shared trait impls but **skips** `WriteJson`.
- Hand-wrote `impl WriteJson for SnowflakeTimestampLtz` to emit a wall-clock literal string via the existing `format_timestamp_string_into` helper (same one the fetch-side CHAR conversion uses, so format correctness is shared).
- NTZ and TZ continue through the renamed `write_timestamp_json_epoch_nanos` helper — no behavioural change for those.
- Three new unit tests in `timestamp_tests.rs` pin all three encoders (LTZ wall-clock with and without fractional digits, NTZ epoch_ns), so a future macro refactor that swaps an encoder is caught before it ships.
- New e2e `SQL_SF_TIMESTAMP_LTZ preserves wall-clock interpreted in non-UTC session timezone` runs with `Asia/Kolkata` (UTC+05:30, no DST so the assertion is DST-immune) and asserts both observables: (a) wall-clock displayed in session TZ is unchanged, (b) underlying UTC instant is shifted by -05:30 from the bound value. A regression to epoch_ns would fail both.

Given the round-trip is now correct end-to-end I dropped `BD#48` from `BehaviorDifferences.yaml` mentally — it was the placeholder for exactly this gap. Let me know if you want me to add a passing-as-of-this-PR note instead.

odbc/src/api/types/odbc_types.rs
sfc-gh-makowalski: Did the full IPD refactor in fa107b963. Changes:

- New `TimestampSubtype { Ntz, Ltz, Tz }` enum + `from_parameter_type` classifier sitting next to the `SQL_SF_TIMESTAMP_*` constants in `odbc_types.rs`. Two new unit tests pin the vendor-code/standard-code contract.
- Added `IpdRecord::sf_subtype: Option<TimestampSubtype>` and matching field on `ParameterBinding`, propagated by `from_apd_ipd`. Default for the untyped `?` markers and `with_varchar_size` is `None`.
- `bind_parameter` now normalises 2000/2001/2002 to `SQL_TYPE_TIMESTAMP` (93) on the IPD and stashes the chosen subtype on `sf_subtype`. `SQLDescribeParam` and `SQLGetDescField(IPD, SQL_DESC_TYPE)` now naturally return spec-mandated 93 with **zero further code change** in those paths.
- `make_converter` now takes `&ParameterBinding` and dispatches `SQL_TYPE_TIMESTAMP` based on `sf_subtype`: default/None and `Some(Ntz)` route to NTZ (preserving Tableau/Excel/Power BI backward compat); `Some(Ltz)` to LTZ; `Some(Tz)` to the explicit "not yet supported" rejection.
- The in-module `make_binding` test helper mirrors the same normalisation so the dispatch tests exercise the production shape — all 4 existing vendor-code unit tests still pass unchanged.
- New e2e `SQLDescribeParam returns SQL_TYPE_TIMESTAMP (93) after binding with vendor codes` binds parameter #2 with each of the three vendor codes in turn and asserts `SQLDescribeParam` reports 93 — the missing pin you flagged.

The 1115 lib tests still pass and clippy is clean.

odbc_tests/tests/e2e/types/timestamp_vendor_bindparam.cpp
sfc-gh-makowalski: Done in fa107b963 — bumped both `fraction` to `123456789` and added `CHECK(ts_out.fraction == 123456789)` to the round-trip in both the NTZ and LTZ vendor tests. A regression that silently truncates to whole seconds now fails the test. The comments above each test note the scale-9 contract being pinned.

sfc-gh-rkowalski (approved):

PR #1005, opened 2026-04-29
fix(odbc): support TIMESTAMP_TZ parameter binding with offset round-trip

sfc-gh-makowalski:
## Summary

Lifts the explicit `SQL_SF_TIMESTAMP_TZ` rejection introduced in the previous PR and routes the vendor code to a new `SnowflakeTimestampTz` binding path that **preserves the original timezone offset** on the wire.

### What changes

- `SnowflakeTimestampTz` now uses a `TzInstant { utc, offset_minutes }` representation so the offset survives the read/bind boundary.
- The TZ Arrow read path (struct + flat layouts) extracts `tz_offset_min` and applies the legacy `+1440` bias.
- The bind path accepts both `SQL_C_TYPE_TIMESTAMP` (treated as UTC, offset = 0; matches the legacy Python connector for naive `datetime`) and `SQL_C_CHAR` / `SQL_C_WCHAR` strings with an optional `+/-HH:MM` suffix.
- `WriteJson` for TZ emits the legacy two-token wire format `"<epoch_ns> <offset_minutes_plus_1440>"` so the server stores the correct instant **and** offset.

### Out of scope (intentionally)

- TZ -> `SQL_C_CHAR/WCHAR` **fetch** rendering. The driver continues to emit the bare UTC wall-clock with no offset suffix, matching the legacy 3.16.0 driver's default `TIMESTAMP_TZ_OUTPUT_FORMAT` and the ODBC spec rule "datetime with timezone -> datetime without timezone drops the offset". A future PR can add session-format-aware rendering when the format string contains TZH/TZM tokens.

## Test plan

- Unit tests cover the new `TzInstant`-based read/write/JSON helpers (offset extraction, JSON wire format with `+1440` bias, CHAR/WCHAR bind parsing, struct-as-UTC fallback).
- The e2e `timestamp_vendor_bindparam.cpp` TZ tests are unskipped and flipped to positive round-trips:
  - Naive `SQL_C_TYPE_TIMESTAMP` bind stores the wall-clock as UTC.
  - `SQL_C_CHAR` bind with a `+05:30` suffix has the offset applied before storage and round-trips correctly through `SELECT`.
- [ ] CI green

### DIFF
--- a/odbc/src/conversion/timestamp.rs
+++ b/odbc/src/conversion/timestamp.rs
@@ -9,8 +9,9 @@ use serde_json::Value;
 use crate::api::CDataType;
 use crate::api::ParameterBinding;
 use crate::conversion::error::{
-    BindingNumericOutOfRangeSnafu, InvalidDatetimeValueSnafu, JsonBindingError,
-    NumericValueOutOfRangeSnafu, UnsupportedCDataTypeSnafu,
+    BindingNumericOutOfRangeSnafu, DatetimeFieldOverflowSnafu, InvalidCharacterValueForCastSnafu,
+    InvalidDatetimeValueSnafu, JsonBindingError, NumericValueOutOfRangeSnafu,
+    UnsupportedCDataTypeSnafu,
 };
 use crate::conversion::error::{
     InvalidArrowValueSnafu, ReadArrowError, UnsupportedOdbcTypeSnafu, WriteOdbcError,
@@ -617,6 +618,16 @@ fn write_timestamp_json_wallclock(value: NaiveDateTime) -> Result<Value, JsonBin
 // -> datetime without timezone converts to UTC".
 // =============================================================================
 
+/// Static template used in both diagnostics and unit tests, so a future
+/// change to the accepted grammar updates the user-facing message and the
+/// pinning test in lockstep.
+const TZ_CHAR_EXPECTED_FORMAT: &str = "YYYY-MM-DD HH:MM:SS[.fff] +/-HH:MM";
+
+/// Cap a (potentially adversarial / huge) bound string before stashing it on
+/// a diagnostic record. ODBC diagnostic message buffers are bounded and
+/// callers shouldn't be able to blow them up by binding a 1 MB literal.
+const TZ_CHAR_MAX_DIAG_LEN: usize = 64;
+
 /// Read a TIMESTAMP_TZ value from a parameter binding. Captures both the UTC
 /// instant and the offset so `write_timestamp_tz_json` can emit the legacy
 /// two-token wire format.
@@ -628,8 +639,13 @@ fn write_timestamp_json_wallclock(value: NaiveDateTime) -> Result<Value, JsonBin
 /// - `SQL_C_CHAR` / `SQL_C_WCHAR`: parse `YYYY-MM-DD HH:MM:SS[.fff] +/-HH:MM`;
 ///   if no offset suffix is present, fall back to the offset-less parser and
 ///   treat as UTC (offset = 0). A genuinely unparseable string surfaces as
-///   `UnsupportedCDataType` (mapped to SQLSTATE HY000) so the caller learns
-///   the bind failed instead of silently storing the Unix epoch.
+///   `InvalidCharacterValueForCast` (mapped to SQLSTATE 22018), carrying a
+///   truncated copy of the input plus the expected format so the caller
+///   learns *what* was rejected and *why*. This is distinct from the 07006
+///   ("Restricted data type attribute violation") that signals an
+///   unsupported binding *shape*, and from 22008 ("Datetime field overflow")
+///   that the JSON writer emits when the parsed instant exceeds the
+///   nanosecond epoch range.
 fn read_timestamp_tz_odbc(binding: &ParameterBinding) -> Result<TzInstant, JsonBindingError> {
     match binding.value_type {
         CDataType::Char => {
@@ -654,8 +670,9 @@ fn read_timestamp_tz_odbc(binding: &ParameterBinding) -> Result<TzInstant, JsonB
 }
 
 /// Try `YYYY-MM-DD HH:MM:SS[.fff] +/-HH:MM` first, then fall back to the
-/// offset-less formats (treated as UTC). Returns `UnsupportedCDataType` if
-/// neither parse succeeds.
+/// offset-less formats (treated as UTC). Returns
+/// `InvalidCharacterValueForCast` (SQLSTATE 22018) if neither shape parses,
+/// carrying a truncated copy of the input and the expected format template.
 fn parse_tz_string_with_fallback(
     s: &str,
     c_type: CDataType,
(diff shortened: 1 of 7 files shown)


sfc-gh-makowalski:
* **#1005** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1005?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1005?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

odbc/src/conversion/timestamp.rs
```diff
@@ -527,50 +570,114 @@ fn write_timestamp_json(value: NaiveDateTime) -> Result<Value, JsonBindingError>
...
+            });
+        }
+    }
+    NaiveDateTime::parse_from_str(s, "%Y-%m-%d %H:%M:%S")
+        .or_else(|_| NaiveDateTime::parse_from_str(s, "%Y-%m-%d %H:%M:%S%.f"))
+        .map(|utc| TzInstant {
+            utc,
+            offset_minutes: 0,
+        })
+        .map_err(|_| UnsupportedCDataTypeSnafu { c_type }.build())
```
sfc-gh-rkowalski: Every parse failure on the new TZ CHAR/WCHAR path collapses to `UnsupportedCDataType` and surfaces as SQLSTATE 07006 ("Restricted data type attribute violation"). That's the wrong category — the C data type *is* supported, it's the value content that failed. The spec class for that is 22018 ("Invalid character value for cast specification"), or 22008 for the `timestamp_nanos_opt() == None` overflow path on line 647-651.

The right tools are already in this module — `error.rs:619-621` maps `InvalidNumericLiteral`/`InvalidBooleanValue` → 22018, and `error.rs:613` maps `InvalidDatetimeValue` → 22007. So apps that switch on SQLSTATE class to distinguish "unsupported binding shape" from "bad input data" silently get the wrong answer here.

The Display message also hides which value was rejected: `"Unsupported C data type for JSON binding: Char"` says nothing about the offending string or the expected format.

Suggested:
- Route TZ-string parse failures through `InvalidDatetimeValue` (or a new `InvalidCharacterValueForCast` variant mapped to 22018), carrying a truncated copy of the input plus the expected format in the Display.
- Route the `timestamp_nanos_opt()` overflow on line 647 to `DatetimeFieldOverflow` → 22008 (already wired in `error.rs:614`).

This pairs naturally with H2 (the doc comment a few lines up, on line 597, also lies about the resulting SQLSTATE — claims HY000, actually 07006). Fix both in the same touch.

odbc/src/conversion/timestamp.rs
```diff
@@ -527,50 +570,114 @@ fn write_timestamp_json(value: NaiveDateTime) -> Result<Value, JsonBindingError>
...
+/// two-token wire format.
+///
+/// Bind paths:
+/// - `SQL_C_TYPE_TIMESTAMP` / `SQL_C_BINARY`: the struct has no offset field,
+///   so we treat the wall-clock as UTC (offset = 0). Matches the legacy
+///   Python connector's treatment of a naive `datetime` bound to TIMESTAMP_TZ.
+/// - `SQL_C_CHAR` / `SQL_C_WCHAR`: parse `YYYY-MM-DD HH:MM:SS[.fff] +/-HH:MM`;
+///   if no offset suffix is present, fall back to the offset-less parser and
+///   treat as UTC (offset = 0). A genuinely unparseable string surfaces as
+///   `UnsupportedCDataType` (mapped to SQLSTATE HY000) so the caller learns
```
sfc-gh-rkowalski: Doc bug: this says `"UnsupportedCDataType (mapped to SQLSTATE HY000)"` but the actual mapping at `error.rs:615-617` produces 07006, not HY000. Tempts a downstream test to assert HY000 and just confuses the next reader.

Fix alongside the SQLSTATE-mapping correction in the comment below — once `UnsupportedCDataType` here gets re-routed to a `22018`/`22008` variant, this doc line should describe the new mapping accurately.

odbc_tests/tests/e2e/types/timestamp_vendor_bindparam.cpp
```diff
@@ -152,14 +153,72 @@ TEST_CASE("SQL_SF_TIMESTAMP_TZ binding is rejected with SQLSTATE 07006",
...
+  // Then The fetched UTC wall-clock matches the offset-applied instant: the
+  // `+05:30` suffix moved 14:30:45 back to 09:00:45 UTC before storage.
+  auto select_stmt = conn.execute_fetch("SELECT ts FROM ts_tz_char_bind WHERE id = 1");
+  auto ts_out = check_no_truncation<SQL_C_TYPE_TIMESTAMP>(select_stmt, 1);
+  CHECK(ts_out.year == 2024);
+  CHECK(ts_out.month == 3);
+  CHECK(ts_out.day == 15);
+  CHECK(ts_out.hour == 9);
+  CHECK(ts_out.minute == 0);
+  CHECK(ts_out.second == 45);
```
sfc-gh-rkowalski, line 302: This is the only e2e test for the offset-bind path, and as written it doesn't actually prove what the headline feature promises. The bind sends `"2024-03-15 14:30:45 +05:30"`, but the readback uses `SQL_C_TYPE_TIMESTAMP`, which **intentionally drops the offset on the fetch side** (see `SnowflakeTimestampTz::write_odbc_type`). All the assertions confirm is the UTC instant — they don't confirm the server stored the offset rather than just deriving UTC via the session TZ.

A buggy implementation that always emitted `1440` (offset=0) on the wire would pass this test. The unit-level wire-format pin at `param_binding.rs:1906` does catch that, but that's the only thing standing between us and a silent regression.

One extra `SELECT` in the same TEST_CASE closes it. Either:

```cpp
auto tz_stmt = conn.execute_fetch(
    "SELECT EXTRACT(TIMEZONE_HOUR FROM ts), EXTRACT(TIMEZONE_MINUTE FROM ts) "
    "FROM ts_tz_char_bind WHERE id = 1");
CHECK(get_data<SQL_C_SLONG>(tz_stmt, 1) == 5);
CHECK(get_data<SQL_C_SLONG>(tz_stmt, 2) == 30);
```

or

```cpp
auto str_stmt = conn.execute_fetch(
    "SELECT TO_VARCHAR(ts, 'YYYY-MM-DD HH24:MI:SS TZH:TZM') FROM ts_tz_char_bind WHERE id = 1");
CHECK(get_data<SQL_C_CHAR>(str_stmt, 1) == "2024-03-15 09:00:45 +05:30");
```

Whichever you prefer.

sfc-gh-rkowalski (changes requested):
Wire format and TzInstant refactor look solid — the +1440 bias is symmetric on encode/decode, the `parse_from_str` → `naive_utc()` chain produces a true UTC instant (no wall-clock-as-UTC trap), and the literal pin `"1705309245000000000 1770"` locks the contract with the server.

Three things I'd like to see before approving though, all clustered around diagnostics and test coverage. Details inline.

odbc/src/conversion/timestamp.rs
sfc-gh-makowalski: Done in 8bd50169e. Routed TZ char/wchar parse failures through a new `JsonBindingError::InvalidCharacterValueForCast { c_type, value, expected_format }` variant mapped to SQLSTATE 22018 (alongside `InvalidNumericLiteral`/`InvalidBooleanValue`), and routed `timestamp_nanos_opt() == None` overflows through `DatetimeFieldOverflow` (22008). Diagnostic value is capped at 64 chars to keep diag-record buffers bounded. Pinning tests live in `odbc/src/api/error.rs` (variant -> SqlState) and `odbc/src/conversion/timestamp.rs` (parse-fn metadata + write-fn overflow). Resolving.

odbc/src/conversion/timestamp.rs
sfc-gh-makowalski: Done in 8bd50169e. Doc on `read_timestamp_tz_odbc` now reflects the new mapping: 22018 for char/wchar parse failures, 22008 for nanosecond-epoch overflow, 07006 reserved for genuinely unsupported binding shapes. Resolving.

odbc_tests/tests/e2e/types/timestamp_vendor_bindparam.cpp
sfc-gh-makowalski: Done in 8bd50169e. Added a follow-up `SELECT TO_VARCHAR(ts, 'YYYY-MM-DD HH24:MI:SS TZH:TZM')` readback to the same TEST_CASE that pins the wall-clock *and* the original `+05:30` offset on the wire. A regression that drops the offset (and lets session-TZ derive UTC) reverts to `+00:00` and fails the assertion. Kept the `SQL_C_TYPE_TIMESTAMP` readback above for the UTC-instant proof. Resolving.

sfc-gh-rkowalski (approved):

PR #1031, opened 2026-04-30
Extend OTel instrumentation with child spans from sf_core

sfc-gh-turbaszek:
## Summary

Bring sf_core telemetry in line with OpenTelemetry's per-operation span model.
Each FFI entry-point opens its own bounded span tagged with `db.system="snowflake"`
and `snowflake.session.id`. The exporter routes spans to the correct session
by reading that attribute — no parent-child registries, no shared connection
trace_id.

This unblocks distributed context propagation: when a wrapper caller has an
active OTEL context (e.g. inside an HTTP handler span), our spans attach to
the caller's trace via standard tracing-subscriber semantics, instead of
forcing every operation under a long-lived connection-scoped trace.

### Span surface

- `session_init` — event-carrier; opens after login, attaches the
  `session_init` OTEL event with environment metadata.
- `statement_execute_query`, `statement_prepare`
- `connection_get_query_result`, `connection_abort_query`, `connection_heartbeat`
- `wrapper_api_usage`, `wrapper_error` — per-call spans for the FFI methods
  that record `api_call` / `exception` OTEL events from the wrapper layer.

### Implementation

- **`snowflake_op_span!` macro** — takes `Option<i64>` and returns
  `Span::none()` when no session id is resolvable, so unknown-handle telemetry
  is silently dropped instead of stamped with a sentinel that could collide
  with a real session id and route to the wrong tenant.
- **`SnowflakeSpanProcessor`** — buffers spans per session, flushes at 50 per
  session or on `flush_connection_telemetry` (called during connection release
  with a bounded 5 s timeout).
- **`SnowflakeInBandExporter`** — groups spans by `snowflake.session.id`
  attribute and POSTs each session's batch to `/telemetry/send` with the
  active session token.
- **`Connection::session_id_cache: AtomicI64`** — populated on login; entry
  points read it lock-free via `DatabaseDriverV1::session_id_for_conn` /
  `session_id_for_stmt`. `0` is a sentinel for "not populated" — never a real id.
- **Serialization** — span events carry `event_kind="event"`; empty operation
  spans carry `event_kind="span"` plus `duration_ms` and
  `status`/`status_description`. Downstream consumers can filter on `event_kind`
  rather than enumerating every span name.

### What the Snowflake backend receives

Same flat JSON shape; `event_kind` distinguishes span records from event records:

```json
{"logs": [
  {"message": {"type": "session_init", "event_kind": "event", "service.name": "snowflake-python", "snowflake.session.id": 12345, ...}, "timestamp": "..."},
  {"message": {"type": "api_call", "event_kind": "event", "api_method": "Connection.cursor", "snowflake.session.id": 12345, ...}, "timestamp": "..."},
  {"message": {"type": "statement_execute_query", "event_kind": "span", "duration_ms": 342, "snowflake.session.id": 12345, ...}, "timestamp": "..."},
  {"message": {"type": "connection_heartbeat", "event_kind": "span", "duration_ms": 2, "snowflake.session.id": 12345, ...}, "timestamp": "..."}
]}
```

### Not yet instrumented

- `statement_execute_async` — lifetime `'a` on `bindings` prevents the
  async-block-capture pattern; needs core logic extraction.
- `connection_set_autocommit`, `connection_use_database`,
  `connection_token_request`, `connection_get_query_status`,
  `connection_send_http_request`.
- The `session_init` span only wraps the event recording — it does **not**
  capture login duration. Instrumenting the full `connection_init` future
  (so login latency / failures become part of the span) is left as a follow-up.
- Accept caller OTEL context from wrappers so driver spans nest under the
  caller's app span.

## Test plan

- [x] sf_core `cargo test --lib` — 891 passed
- [x] sf_core telemetry integration tests — 26 passed
- [x] Python integ against local SUT — 439 passed, 1 unrelated env failure
      (`TIMESTAMP_NTZ` cursor-description; reproduces on main)
- [x] Python e2e — 744 passed; remaining failures are env (numpy/pandas
      missing, `SNOWFLAKE` db absent on local SUT)
- [x] `test_api_usage_telemetry_sent_on_cursor_creation` updated to assert
      the new `db.system` and `event_kind` attributes; no longer pins
      `code.filepath` / `code.namespace` to a specific path.
- [x] All pre-commit hooks pass (rustfmt, clippy, cargo check, ruff)

### DIFF
--- a/sf_core/src/apis/database_driver_v1/connection.rs
+++ b/sf_core/src/apis/database_driver_v1/connection.rs
@@ -1,10 +1,11 @@
 use snafu::{OptionExt, ResultExt};
 use std::future::Future;
 use std::sync::RwLock;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicI64, Ordering};
 use std::{collections::HashMap, sync::Arc};
 use tokio::sync::Mutex;
 use tokio::sync::RwLock as AsyncRwLock;
+use tracing::Instrument;
 
 use super::Setting;
 use super::async_query_registry::AsyncQueryRegistry;
@@ -345,6 +346,11 @@ impl DatabaseDriverV1 {
                     )
                     .await;
 
+                    // Populate the lock-free session id cache so entry-point
+                    // methods can stamp `snowflake.session.id` on their spans
+                    // without taking the connection mutex.
+                    conn.session_id_cache.store(session_id, Ordering::Relaxed);
+
                     // Telemetry setup: check if the server has opted this session
                     // into in-band telemetry and a session registry is configured,
                     // then register the session so spans tagged with this
@@ -389,21 +395,18 @@ impl DatabaseDriverV1 {
                             .as_ref()
                             .map(crate::telemetry::environment::EnvironmentInfo::with_wrapper)
                             .unwrap_or_else(crate::telemetry::environment::EnvironmentInfo::detect);
-                        // Long-lived connection span carries all telemetry events
-                        // (session_init, api_usage, wrapper_error) for this session.
-                        // Events are exported when the span ends on connection release.
-                        let conn_span =
-                            tracing::info_span!("connection", "snowflake.session.id" = session_id,);
-
-                        // Record session_init as an event on the connection span.
-                        // We enter the span so the OpenTelemetryLayer captures the
-                        // tracing event as an OTel span event.
-                        {
-                            let _guard = conn_span.enter();
-                            crate::telemetry::record_session_init(&env_info);
-                        }
-
-                        conn.telemetry_span = Some(conn_span);
+                        // Bounded `session_init` span: an event-carrier span whose
+                        // sole purpose is to attach `snowflake.session.id` to the
+                        // session_init OTEL event. It does NOT measure connection_init
+                        // duration — login has already completed by the time this span
+                        // opens. Each subsequent driver operation emits its own bounded
+                        // span tagged with the same session id; they share no trace_id
+                        // with this span.
+                        // TODO: instrument the surrounding connection_init future so we
+                        // capture login duration + Status::Error on failure.
+                        let init_span = crate::snowflake_op_span!("session_init", Some(session_id));
+                        let _guard = init_span.enter();
+                        crate::telemetry::record_session_init(&env_info);
                     }
 
                     if keep_alive {
@@ -440,45 +443,35 @@ impl DatabaseDriverV1 {
         }
     }
 
-    /// Return the per-connection telemetry span if available.
+    /// Flush all telemetry for this connection and clean up the session registry.
     ///
-    /// Callers enter this span and record OTel events (via
-    /// `telemetry::record_api_call` / `record_exception`). The span
-    /// carries `snowflake.session.id` for routing by the shared exporter.
-    pub async fn telemetry_span(&self, handle: Handle) -> Option<tracing::Span> {
-        let conn_ptr = self.connections.get_obj(handle)?;
-        let conn = conn_ptr.lock().await;
-        conn.telemetry_span.clone()
-    }
-
-    /// End the connection span and deregister from the shared telemetry
-    /// session registry. Must be called while session tokens are still alive
-    /// (before logout/cleanup) so the exporter can authenticate the POST.
-    /// Idempotent: no-op if the span was already taken.
+    /// Order matters:
+    /// 1. Flush per-session buffered spans — awaited export to `/telemetry/send`
+    ///    while session tokens are still alive for authentication.
+    /// 2. Remove the session from the exporter registry so it no longer resolves.
+    ///
+    /// Must be called while session tokens are still alive (before logout).
+    /// Idempotent. Each driver operation emits a bounded span that ends when the
+    /// operation returns, so there is no long-lived parent span to drop here.
     pub async fn flush_connection_telemetry(&self, conn_handle: Handle) {
         let Some(conn_ptr) = self.connections.get_obj(conn_handle) else {
             return;
         };
-        let (span, tokens_arc) = {
-            let mut conn = conn_ptr.lock().await;
-            (conn.telemetry_span.take(), conn.tokens.clone())
+        let tokens_arc = {
+            let conn = conn_ptr.lock().await;
+            conn.tokens.clone()
         };
-        // Mutex dropped before reading the tokens RwLock to avoid deadlock.
         let session_id = tokens_arc.read().await.as_ref().map(|t| t.session_id);
 
-        // Drop the tracing span — this ends the underlying OTel span.
-        // SimpleSpanProcessor calls the exporter synchronously via
-        // futures_executor::block_on.  The exporter spawns the HTTP
-        // POST on the tokio runtime and awaits the JoinHandle, so by
-        // the time drop() returns the telemetry has been sent.
-        drop(span);
-        if let Some(id) = session_id
-            && let Some(sessions) = self.telemetry_sessions()
-        {
-            sessions
-                .write()
-                .unwrap_or_else(|e| e.into_inner())
-                .remove(&id);
+        if let Some(id) = session_id {
+            self.flush_telemetry_session(id).await;
+
+            if let Some(sessions) = self.telemetry_sessions() {
+                sessions
+                    .write()
+                    .unwrap_or_else(|e| e.into_inner())
+                    .remove(&id);
+            }
         }
     }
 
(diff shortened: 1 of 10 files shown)


sfc-gh-asolarski:
* **#1200** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1200?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1031** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1031?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1031?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sf_core/src/apis/database_driver_v1/connection.rs
```diff
@@ -1664,51 +1699,58 @@ impl DatabaseDriverV1 {
...
+                    Err(_) => return Ok(false),
                 }
-                Err(_) => return Ok(false),
             }
         }
+        .instrument(tracing::info_span!(
+            parent: self.telemetry_parent(conn_handle),
+            "connection_heartbeat"
+        ))
+        .await
```
sfc-gh-turbaszek: @sfc-gh-mhofman @sfc-gh-jszczerbinski wdyt?

sf_core/src/logging/log_manager.rs
```diff
@@ -221,20 +243,21 @@ impl LogManager {
...
             let exporter =
                 crate::telemetry::snowflake_exporter::SnowflakeInBandExporter::new(sessions);
+            let (processor, flush_handle) =
+                crate::telemetry::snowflake_processor::SnowflakeSpanProcessor::new(exporter);
             let provider = opentelemetry_sdk::trace::SdkTracerProvider::builder()
-                .with_simple_exporter(exporter)
+                .with_span_processor(processor)
                 .build();
             let tracer = provider.tracer("snowflake.telemetry");
-            let layer =
```
sfc-gh-asolarski, line 231: without the filter
> "Every span from tokio, hyper, tower, tonic, etc. flows through SnowflakeSpanProcessor::on_end → acquires buffers mutex → calls extract_session_id → linear scan of attributes → returns None → drops. Under high concurrency with many library-generated spans, this is unnecessary lock contention on the hot path."

so I guess we should have some basic filtering like by name starting with "sf_core" or something like that

sf_core/src/apis/database_driver_v1/connection.rs
```diff
@@ -781,10 +774,12 @@ pub struct Connection {
...
-    /// Long-lived connection span carrying `snowflake.session.id`.
-    /// Telemetry events (session_init, api_usage, wrapper_error) are
-    /// recorded as OTel events on this span; it is ended on release.
-    pub(crate) telemetry_span: Option<tracing::Span>,
+    /// Snowflake session id, populated from `tokens.session_id` immediately after
+    /// `connection_init` succeeds. Stored as `AtomicI64` so entry-point methods
+    /// can resolve it lock-free when stamping `snowflake.session.id` on their
+    /// telemetry spans. `0` means "not yet populated"; resolvers must treat `0`
+    /// as `None` to avoid colliding with a real session id of zero.
+    pub(crate) session_id_cache: AtomicI64,
```
sfc-gh-asolarski: > Stored as `AtomicI64` so entry-point methods can resolve it lock-free when stamping `snowflake.session.id` on their telemetry spans.

But the mutex on a connection is always locked before accessing this.

In theory it could be decoupled and stored on handle manager and accessed without locking I assume, but is it worth it?

sf_core/src/apis/database_driver_v1/global_state.rs
```diff
@@ -112,6 +113,43 @@ impl DatabaseDriverV1 {
...
+        match conn.session_id_cache.load(Ordering::Relaxed) {
+            0 => None,
+            id => Some(id),
+        }
+    }
+
+    /// Resolve the Snowflake session id for a statement handle by traversing
+    /// to its owning connection's cached session id (see
+    /// [`Self::session_id_for_conn`]).
+    pub(crate) async fn session_id_for_stmt(&self, stmt_handle: Handle) -> Option<i64> {
```
sfc-gh-asolarski, line 129: to avoid this double locking (first on connection than on statement) we could cache session id on statement, but I'm not sure is it worth it

sfc-gh-asolarski: anyway, when it's used in stmt execute we even have more locking:
1. first lock stmt to get connection, release
2. lock connection to get session id, release
3. start execute, inside execute again we lock connection and stmt

and we claimed that `pub(crate) session_id_cache: AtomicI64` is intended to be lock-free

so maybe it's worth making usage of it really lock-free 🤔

sf_core/src/apis/database_driver_v1/statement.rs
```diff
@@ -410,52 +421,69 @@ impl DatabaseDriverV1 {
...
-        with_valid_session(&conn_ptr, |token| {
-            let http_client = &http_client;
-            let query_parameters = &query_parameters;
-            let query_id = &query_id;
-            async move {
-                snowflake_abort_query(http_client, query_parameters, token.reveal(), query_id).await
-            }
-        })
+        let session_id = self.session_id_for_conn(conn_handle).await;
+        async {
```
sfc-gh-asolarski, line 498: this looks like repeatable pattern - get session id, async {code}.instrument.await, maybe we could simplify
maybe introduce `#[snowflake_instrument("connection_abort_query", stmt_handle)]` or `#[snowflake_instrument("connection_abort_query", conn_handle)]`  or something like that?

sfc-gh-asolarski (changes requested):

sf_core/src/apis/database_driver_v1/connection.rs
sfc-gh-turbaszek: Done

sf_core/src/apis/database_driver_v1/global_state.rs
sfc-gh-turbaszek: Fixed

sf_core/src/apis/database_driver_v1/statement.rs
sfc-gh-turbaszek: Reasonable suggestion but I'd defer it. A proc macro needs its own crate (sf_core_macros) plus a build-time dependency, and the current pattern is:                                                          
```rust                                                                                                                                                    
let session_id = self.session_id_for_X(handle);                                                                                                                                                              
async { ... }.instrument(crate::snowflake_op_span!("name", session_id)).await                                                                                                                                
```                                                                                                                                                                                                  
 …three lines per method. With the lock-free fix the resolution is sync, so the duplication is even smaller. Filing as a follow-up — happy to revisit if many more entry-points get instrumented.

sfc-gh-asolarski (approved):

sf_core/src/apis/database_driver_v1/global_state.rs
```diff
@@ -75,6 +75,18 @@ pub struct DatabaseDriverV1 {
     pub(super) connections: HandleManager<Mutex<Connection>>,
     pub(super) statements: HandleManager<Mutex<Statement>>,
     pub(super) results: HandleManager<Mutex<ResultSet>>,
+    /// Cached `connection_handle.id → Snowflake session_id` populated on login
+    /// and removed on connection release. Keeping this off the `Connection`
+    /// struct (which is behind a `tokio::Mutex`) lets entry-point methods
+    /// resolve the session id without taking the connection mutex — the only
+    /// writers are login / logout, so reads under the parallel `RwLock` are
+    /// uncontended in normal operation.
+    pub(super) session_ids: RwLock<HashMap<u64, i64>>,
```
sfc-gh-jszczerbinski: This is an antipattern IMHO, we have global RwLock instead of connection scoped ones. This maximizes contention with many connections. Moreover let's not operate on raw handles outside HandleManager -> This seems like bad idea.

Currently we have HandleManager that holds Mutex<Connection>, this optimized for correctness not efficiency.

We don't want to optimize prematurely - but at some point we probably will. The way to optimize it is by  HandleManager holding immutable Connection struct instead of Mutex<Connection>. We will allow for interior mutability with finer grained locks for particular fields: you can see that we are preparing for it already as we have "unnecessary locks" (the mutability is already handled by top level mutex) on some fields (session token for example) already.

sf_core/src/telemetry/mod.rs
```diff
@@ -6,18 +6,53 @@ pub mod types;
...
+/// per-operation span records from event records (`session_init`, `api_call`,
+/// `exception`) that share the same `/telemetry/send` payload shape.
+#[macro_export]
+macro_rules! snowflake_op_span {
+    ($name:expr, $session_id:expr) => {
+        match $session_id {
+            ::std::option::Option::Some(id) => ::tracing::info_span!(
+                $name,
+                "db.system" = "snowflake",
+                "snowflake.session.id" = id,
```
sfc-gh-jszczerbinski, line 45: Use SESSION_ID_FIELD?

sfc-gh-jszczerbinski (changes requested):
In general LGTM, but I don't like how we hold and lookup session ids. Seems like premature optimization done wrong way

sfc-gh-jszczerbinski (approved):

PR #1035, opened 2026-05-04
NO-SNOW: Fix clippy result_large_err by heap allocating error source enums

sfc-gh-ddas:
### TL;DR

Box large error variants in `ApiError` to reduce enum size and satisfy Clippy's `result_large_err` lint.

### What changed?

Several `source` fields in `ApiError` variants (`Login`, `TlsClientCreation`, `SessionRefresh`, `Query`, `TokenRequest`) have been changed from inline types (`RestError`, `TlsError`) to `Box<RestError>` and `Box<TlsError>`. The `snafu` attribute `#[snafu(source(from(..., Box::new)))]` is used to preserve ergonomic error conversion. Pattern matches against these variants in `converter.rs` have been updated to dereference the boxed sources via `.as_ref()`. A `#[allow(clippy::result_large_err)]` annotation was added to `bump_batch_idx` in `data.rs` where suppression is intentional.

### How to test?

Run the existing test suite, particularly the tests in `converter.rs` (e.g., `query_failed` and `async_query` helpers), to confirm error conversion and pattern matching behavior is unchanged. Verify that the build produces no new Clippy warnings related to `result_large_err`.

### Why make this change?

`RestError` and `TlsError` are large types, and embedding them directly in `ApiError` variants causes the overall enum size to grow significantly. Boxing these sources reduces the size of the `ApiError` enum, addressing the `clippy::result_large_err` lint and improving stack usage when these errors are propagated.

PR #1036, opened 2026-05-04
SNOW-3447929: Custom Root store not ignored when CRL is disabled

sfc-gh-ddas:
### TL;DR

Fix custom root store being silently ignored when CRL validation is disabled.

### What changed?

When CRL validation is disabled and a custom root store path is provided, the TLS client now correctly applies the custom root certificates instead of falling back to system roots. Previously, the custom root store was ignored in this case, and a warning was logged. Now the client branches on whether a custom root store is present: if one is provided, it builds a `ClientConfig` using those roots; otherwise, it falls back to the default system roots.

The existing test for custom PEM roots has been strengthened to assert a successful response rather than silently discarding the result.

A new end-to-end test (`should_trust_custom_root_store_when_crl_disabled`) has been added to verify this behavior. It spins up a local TLS proxy using a self-signed certificate not in any system trust store, writes the certificate to a temp file, and confirms that a client configured with that cert as a custom root can complete the TLS handshake successfully.

### How to test?

Run the TLS handshake end-to-end tests:

```
cargo test -p sf_core --test e2e tls::handshake
```

All three tests should pass:
- `should_complete_handshake_with_default_roots`
- `should_complete_handshake_with_custom_pem_roots`
- `should_trust_custom_root_store_when_crl_disabled`

### Why make this change?

Users who configure a custom root store path expect their custom CA certificates to be trusted regardless of whether CRL checking is enabled. The previous behavior silently discarded the custom root store when CRL was disabled, causing TLS handshake failures against servers using privately-signed certificates in environments where CRL checking is not needed.

### DIFF
--- a/odbc/src/api/data.rs
+++ b/odbc/src/api/data.rs
@@ -474,6 +474,7 @@ fn fetch_impl(statement_handle: sql::Handle, warnings: &mut Warnings) -> OdbcRes
 /// Advance `batch_idx` by `delta` rows within the current `RecordBatch`.
 /// Returns `OdbcError::InternalError` if advancing would cross the batch
 /// boundary or the statement is not in the `Fetching` state.
+#[allow(clippy::result_large_err)]
 fn bump_batch_idx(state: &mut crate::api::State<StatementState>, delta: usize) -> OdbcResult<()> {
     if delta == 0 {
         return Ok(());
--- a/sf_core/src/apis/database_driver_v1/error.rs
+++ b/sf_core/src/apis/database_driver_v1/error.rs
@@ -39,7 +39,8 @@ pub enum ApiError {
     Login {
         #[snafu(implicit)]
         location: Location,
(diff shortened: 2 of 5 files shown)


sfc-gh-ddas:
* **#1036** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1036?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1036?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1035** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1035?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sfc-gh-pbulawa (approved):

PR #1037, opened 2026-05-04
SNOW-3447930: Wire verify_hostname into TLS Client Construction

sfc-gh-ddas:
### TL;DR

Add optional hostname verification bypass to the TLS client, allowing connections to proceed even when the server's certificate hostname does not match the requested host.

### What changed?

A `verify_hostname` flag is now respected across all TLS client construction paths:

- **CRL disabled + custom root store**: Uses a new `NoHostnameVerifier` custom verifier that validates the certificate chain against the trust anchor but skips the hostname check.
- **CRL disabled + default system roots**: Uses reqwest's `danger_accept_invalid_hostnames(true)` when hostname verification is disabled.
- **CRL enabled/advisory**: Passes `verify_hostname` through to `CrlServerCertVerifier`, which switches between full `WebPkiServerVerifier` validation and a chain-only path (`verify_server_cert_signed_by_trust_anchor`) that omits the hostname check.

`NoHostnameVerifier` is a new `ServerCertVerifier` implementation that performs full chain and signature validation while ignoring the `ServerName`. A warning is emitted via `tracing::warn!` whenever hostname verification is disabled, regardless of path.

### How to test?

- The new unit test `hostname_bypass_allows_mismatched_server_name_in_advisory` verifies that `CrlServerCertVerifier` with `verify_hostname=false` accepts a mismatched server name, while `verify_hostname=true` rejects it.
- The new e2e test `should_skip_hostname_verification_when_disabled` spins up a TLS proxy using a self-signed certificate valid for `localhost`, then connects via `127.0.0.1` (a hostname mismatch). It asserts that `verify_hostname=false` succeeds and `verify_hostname=true` fails.

### Why make this change?

Some deployment environments require connecting to services by IP address or an alternative hostname that does not match the CN/SAN on the server certificate (e.g., internal infrastructure with self-signed or fixed certificates). This change allows hostname verification to be explicitly disabled while still enforcing certificate chain trust, giving operators a controlled escape hatch without disabling certificate validation entirely.

### DIFF
--- a/sf_core/src/tls/client.rs
+++ b/sf_core/src/tls/client.rs
@@ -1,7 +1,6 @@
 use crate::crl::config::{CertRevocationCheckMode, CrlConfig};
 use crate::tls::CrlServerCertVerifier;
 use crate::tls::config::TlsConfig;
-use crate::tls::crl_verifier::NoHostnameVerifier;
 use crate::tls::error::{
     ClientBuildSnafu, PemParseSnafu, RootStoreAddSnafu, TlsError, VerifierBuildSnafu,
 };
@@ -29,13 +28,12 @@ pub fn create_tls_client_with_config(tls_config: TlsConfig) -> Result<Client, Tl
     // Install aws-lc-rs provider (idempotent)
     let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
 
-    let custom_root_store = if let Some(pem_path) = tls_config.custom_root_store_path.as_ref() {
+    let custom_pem = if let Some(pem_path) = tls_config.custom_root_store_path.as_ref() {
         tracing::debug!(
             "Loading custom root certificate store from: {}",
             pem_path.display()
         );
-        let pem_data = std::fs::read(pem_path).context(PemParseSnafu)?;
-        Some(create_root_store_from_pem(&pem_data)?)
+        Some(std::fs::read(pem_path).context(PemParseSnafu)?)
     } else {
         None
     };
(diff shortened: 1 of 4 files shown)


sfc-gh-ddas:
* **#1037** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1037?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1037?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1039** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1039?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sf_core/src/tls/client.rs
```diff
@@ -44,43 +45,72 @@ pub fn create_tls_client_with_config(tls_config: TlsConfig) -> Result<Client, Tl
         CertRevocationCheckMode::Disabled => {
             if let Some(root_store) = custom_root_store {
                 tracing::debug!("CRL disabled, applying custom root store without CRL");
-                let tls_config = ClientConfig::builder()
-                    .with_root_certificates(Arc::new(root_store))
-                    .with_no_client_auth();
+                let root_store = Arc::new(root_store);
```
sfc-gh-jszczerbinski:
```
            let mut builder = configure_http_client(Client::builder());
            if let Some(ref pem_data) = pem_data {
                tracing::debug!("CRL disabled, applying custom root store without CRL");
                let certs = reqwest::Certificate::from_pem_bundle(pem_data)
                    .context(ClientBuildSnafu)?;
                builder = builder.tls_built_in_root_certs(false);
                for cert in certs {
                    builder = builder.add_root_certificate(cert);
                }
            } else {
                tracing::debug!("CRL disabled, using default system roots");
            }
            if !tls_config.verify_hostname {
                tracing::warn!("Hostname verification disabled");
                builder = builder.danger_accept_invalid_hostnames(true);
            }
```

sfc-gh-jszczerbinski: We can do sth like this - we do not need custome verifier

sfc-gh-jszczerbinski (changes requested):
Changes needed, please take a look at previous comment

sf_core/src/tls/client.rs
sfc-gh-ddas: Done.

sfc-gh-jszczerbinski (approved):

PR #1047, opened 2026-05-04
NO-SNOW: derive host from account and strip wrapper prefixes from query errors

sfc-gh-turbaszek:
## Summary

Fixes two regressions versus the legacy `snowflake-connector-python`:

1. **Connecting with only `account` now works again.** The old driver accepts just `account` (plus auth credentials) and derives `host = {account}.snowflakecomputing.com`. The new driver rejected the connection because the Rust core required `host` or `server_url`. Added `derive_host_from_account()` in the resolver to fill in the canonical host when neither is provided.

2. **Query-error messages no longer carry a doubled prefix.** Previously:
   ```
   002003 (42S02): Query execution failed: Query failed: SQL compilation error: ...
   ```
   Now matches the old driver exactly:
   ```
   002003 (42S02): SQL compilation error: ...
   ```
   Removed the SNAFU `display` prefixes on `ApiError::Query` and `RestError::QueryFailed` so the server's error surfaces verbatim in `DriverException.message`.

Also replaces `pkgs.rustup` in `shell.nix` with `rust-overlay`. The rustup-downloaded toolchain isn't patched against nix's glibc and breaks whenever the nix store changes — `rust-overlay` gives us a properly-linked toolchain while still honoring `rust-toolchain.toml`.

## Test plan

- [x] Rust unit tests for `derive_host_from_account` (6 cases covering simple, region-embedded, org-account, and skip-when-set scenarios)
- [x] Rust integration test that exercises the full `resolver::resolve_with_paths` → `ConnectionConfig::build` pipeline with only `account`
- [x] Rust unit tests asserting exact display of `ApiError::Query` and `DriverException.message` with no wrapper prefixes
- [x] Python unit test asserting `str(error) == "002003 (42S02): SQL compilation error: ..."` (exact match)
- [x] Python integration tests (`TestConnectionParameters`, `TestErrorMessageFormat`) runnable as reference tests against both the new and legacy drivers
- [ ] Verify integration tests pass in CI against a real Snowflake account

### DIFF
--- a/python/tests/integ/test_connection.py
+++ b/python/tests/integ/test_connection.py
@@ -36,18 +36,15 @@ class TestConnectionParameters:
     These tests exercise that behavior against both drivers.
     """
 
-    def test_connect_with_account_only_no_host_or_server_url(self, connection_factory):
+    def test_connect_with_account_only_no_host_or_server_url(self, connector_adapter, monkeypatch):
         """Connection succeeds when only ``account`` is given — host/server_url are derived.
 
         Skipped when the test environment targets a non-production host (preprod,
         localhost, or a dev deployment) — derivation only yields a valid URL for
         accounts whose canonical host is ``{account}.snowflakecomputing.com``.
-
-        Passing ``None`` for host/server_url/port/protocol causes the connection
-        factory to omit those parameters entirely, forcing the driver to derive
-        them from ``account``.
         """
         from tests.config import get_test_parameters
+        from tests.connector_factory import create_connection_with_adapter
 
         test_params = get_test_parameters()
         account = test_params.get("SNOWFLAKE_TEST_ACCOUNT")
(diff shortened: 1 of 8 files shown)


python/tests/integ/test_connection.py
```diff
@@ -28,6 +28,66 @@ def test_get_connection_info_returns_info_after_connect(self, connection):
         assert info is not None
 
 
+class TestConnectionParameters:
```
sfc-gh-pczajka, line 32: Could we make this test easier: let's use get parameters + pure connect() instead of monkeypatching get_connection_parameters

sfc-gh-turbaszek: Done — the test now uses `get_test_parameters()` + `connector_adapter.connect(...)` directly with a minimal params dict (`account` + JWT auth), no monkeypatching of `get_connection_parameters`.

sfc-gh-pczajka (approved):

PR #1048, opened 2026-05-04
NO-SNOW: python - bootstrap local reg test environment

sfc-gh-turbaszek:
## Summary
- Add `scripts/setup_local_reg.py`: generates an RSA keypair at `~/.snowflake-reg/`, registers the public key on the configured user via `ALTER USER`, creates the test database/schema/warehouse, and rewrites `parameters.json` with `SNOWFLAKE_TEST_PRIVATE_KEY_CONTENTS`. Idempotent and no-ops on non-`*.reg.local` hosts.
- Wire it into `python/tests/conftest.py` so any pytest session that targets a `*.reg.local` host auto-bootstraps JWT auth — no manual step.
- Pin the Rust toolchain in `shell.nix` via `rust-overlay` + `rust-toolchain.toml`, replacing `pkgs.rustup` which drifted away from the repo's declared toolchain.

## Test plan
- [ ] `parameters.json` pointing at `snowflake.reg.local:8082` (admin/test): integ suite bootstraps cleanly and most tests pass.
- [ ] `parameters.json` pointing at `s3testaccount.reg.local:8082` (snowman/test): put/get e2e suite passes (36/36).
- [ ] `parameters.json` pointing at a real Snowflake account: bootstrap is a no-op, test run is unchanged.


python/tests/conftest.py
```diff
@@ -21,10 +21,25 @@
...
-
 _MIN_XDIST_WORKERS = 8
 
 
+def pytest_configure(config):
+    # scripts/ is not a Python package (no __init__.py, not on sys.path), so
+    # load setup_local_reg.py by file path via importlib. bootstrap() is a
+    # no-op unless SNOWFLAKE_TEST_HOST points at a *.reg.local instance.
+    import importlib.util
+    import pathlib
```
sfc-gh-asolarski, line 32: let's maybe stick to top-level imports

sfc-gh-asolarski (approved):

PR #1059, opened 2026-05-06
NO-SNOW: Migrate ODBC version metadata from version.sh to Cargo.toml

sfc-gh-pbulawa:
### TL;DR

Consolidate ODBC versioning into `Cargo.toml` and replace `version.sh` with structured metadata under `[package.metadata.odbc]`.

### What changed?

- Removed `odbc/version.sh` and replaced it with two new keys in `[package.metadata.odbc]` inside `odbc/Cargo.toml`:
  - `odbc_preview_version` — the base release version (e.g. `0.0.1`) used for RPM/DMG/MSI artifact naming and the Windows DLL `VERSIONINFO` resource.
  - `odbc_api_version` — the ODBC API conformance level (e.g. `03.80`) reported by `SQLGetInfo(SQL_DRIVER_ODBC_VER)` and registered in the Windows ODBC registry and unixODBC.
- `build.rs` now reads both keys from `Cargo.toml` via a new `read_odbc_metadata` helper, exposes `SF_ODBC_API_VER` as a compile-time env var, and uses it in `connection.rs` instead of a hardcoded string.
- Packaging scripts (`package.sh` for macOS and Linux, `package.ps1` for Windows) now parse both keys directly from `Cargo.toml` instead of sourcing `version.sh`.
- Introduced shared ini templates under `odbc/installer/shared/templates/` (`odbcinst.ini.template` and `odbc.ini.template`) with placeholders (`__DRIVER_PATH__`, `__SF_ACCOUNT__`, `__ODBC_API_VERSION__`) that are resolved at packaging or install time.
- The macOS `postinstall` script was rewritten to use the shared templates and now also writes a default DSN to `odbc.ini` in addition to registering the driver in `odbcinst.ini`. It uses robust `write_ini_section` and `upsert_ini_kv` helpers to avoid duplicate entries on reinstall.
- The Linux `after_install.sh` script now renders the shared templates via `render_template` and pipes them to `odbcinst` instead of using inline heredocs.
- CI artifact names for MSI and DMG no longer embed the version string, simplifying artifact references.
- `DriverODBCVer` in both WiX installer sources (`snowflake_odbc_x64.wxs`, `snowflake_odbc_x86.wxs`) is now driven by the `OdbcApiVer` variable sourced from `Cargo.toml` rather than a hardcoded `03.52`.

### How to test?

- Build the ODBC driver and verify `SQLGetInfo(SQL_DRIVER_ODBC_VER)` returns `03.80`.
- Run the macOS packaging script and confirm the `.pkg` and `.dmg` are produced with the correct version, and that installing the `.pkg` correctly populates both `~/Library/ODBC/odbcinst.ini` and `~/Library/ODBC/odbc.ini`. Re-run the installer and verify no duplicate sections appear.
- Run the Linux packaging script and confirm the RPM is produced and that `after_install.sh` correctly registers the driver and DSN via `odbcinst`.
- Run the Windows packaging script and confirm the MSI is produced and that the registry entries contain `DriverODBCVer=03.80`.
- Verify CI artifact uploads succeed with the new simplified artifact names.

### Why make this change?

`version.sh` was a shell-only artifact that required separate parsing logic in every packaging script and CI step, and the ODBC API version was hardcoded in multiple places independently. Centralizing both values in `Cargo.toml` under `[package.metadata.odbc]` makes them the single source of truth, eliminates the shell-script dependency, and ensures the API version reported at runtime, embedded in the Windows DLL, and registered with driver managers is always consistent.

### DIFF
--- a/odbc/build.rs
+++ b/odbc/build.rs
@@ -1,6 +1,11 @@
 include!(concat!(env!("CARGO_MANIFEST_DIR"), "/../build_common.rs"));
 
 fn main() {
+    let manifest_dir = std::env::var("CARGO_MANIFEST_DIR").unwrap();
+    let odbc_api_version = read_odbc_metadata(&manifest_dir, "odbc_api_version");
+    println!("cargo:rustc-env=SF_ODBC_API_VER={odbc_api_version}");
+    println!("cargo:rerun-if-changed=Cargo.toml");
+
     #[cfg(not(target_os = "windows"))]
     {
         emit_loader_rpaths();
@@ -10,12 +15,11 @@ fn main() {
     // This avoids the PE/COFF 65535 export symbol limit.
     #[cfg(target_os = "windows")]
     {
-        let manifest_dir = std::env::var("CARGO_MANIFEST_DIR").unwrap();
         let manifest_path = std::path::Path::new(&manifest_dir);
         let def_path = manifest_path.join("exports.def");
         println!("cargo:rustc-cdylib-link-arg=/DEF:{}", def_path.display());
 
-        let (major, minor, patch) = parse_base_version(&manifest_dir);
+        let (major, minor, patch) = parse_odbc_preview_version(&manifest_dir);
         let commit_hash = resolve_commit_hash();
 
         let target_arch =
@@ -40,7 +44,6 @@ fn main() {
         )
         .expect("failed to write version_generated.h");
 
-        println!("cargo:rerun-if-changed=version.sh");
         println!("cargo:rerun-if-changed=exports.def");
         println!("cargo:rerun-if-changed=src/setup/resource.rc");
         println!("cargo:rerun-if-changed=src/setup/resource.h");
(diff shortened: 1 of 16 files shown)


odbc/installer/unix/after_install.sh
```diff
@@ -5,33 +5,30 @@
...
 if [[ -z "$SF_ACCOUNT" ]]; then
     echo "[WARN] SF_ACCOUNT is not set, please manually update the odbc.ini file after installation"
     SF_ACCOUNT=SF_ACCOUNT
 fi
 
-ODBC_DIR=/usr/lib64/snowflake/odbc
+render_template() {
+    sed \
+        -e "s|__DRIVER_PATH__|${DRIVER_PATH}|g" \
+        -e "s|__SF_ACCOUNT__|${SF_ACCOUNT}|g" \
```
sfc-gh-mhofman, line 26: what about replacing \__ODBC_API_VERSION__ in odbcinst.ini?

sfc-gh-pbulawa: It's done when the package is created. The DRIVER_PATH and SF_ACCOUNT are dynamic ones, replaced when package is installed.

odbc/build.rs
```diff
@@ -60,32 +63,66 @@ fn main() {
...
     }
-    panic!("BASE_VERSION not found in version.sh")
+    panic!(
+        "`{key}` not found in {SECTION} of {}",
+        cargo_toml_path.display()
+    );
+}
+
+#[cfg(target_os = "windows")]
+fn parse_odbc_preview_version(manifest_dir: &str) -> (u32, u32, u32) {
```
sfc-gh-mhofman, line 108: so this script now is fine tunded for preview version.  
I'm worried we might miss that when targeting PuPr or worse GA. Do you have any ideas behind that or comments that it's fine for now?  
It could be just renaming this script, having a constant equal to preview or sth. Lets have it visible and not applicable for non private preview.

sfc-gh-mhofman (approved):
check my comments

odbc/build.rs
sfc-gh-pbulawa: We won't miss it, as when the preview version is removed from the toml, the build will fail.

sfc-gh-pbulawa: Also, for the PuPr, we want to change the artifact names, config keys (e.g. `Snowflake ODBC UD` -> `Snowflake ODBC`).

sfc-gh-pbulawa:
* **#1059** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1059?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1059?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #987, opened 2026-04-27
SNOW-3441376: Add Core Windows x86_32 support

sfc-gh-pbulawa:
### TL;DR

Adds Windows x86 (32-bit) CI testing for `sf_core` and fixes compatibility issues that prevented it from building and running correctly on 32-bit targets.

### What changed?

- Added a new `test_windows_x86` CI job that builds and tests `sf_core` targeting `i686-pc-windows-msvc`, including a verification step that confirms the output DLL is 32-bit (machine type `14C`).
- Included `test_windows_x86` in the `rust-core-status` gate so failures block merges.
- Made `sonic-rs` an optional dependency behind a new `sonic-json` feature flag. When `sonic-json` is not enabled, JSON row parsing falls back to `serde_json`. This allows the crate to build on x86 Windows where `sonic-rs` (which requires SIMD intrinsics) is not available.
- Fixed a pointer width bug in `BinaryDataPtr` → `DataPtr` conversion: the raw pointer bytes are now read as `u64` before casting to `usize`, ensuring correct behavior on both 32-bit and 64-bit platforms.
- Updated the `sf_core` dev-dependency on itself to use `default-features = false` so tests don't inadvertently pull in features that break on constrained targets.

### How to test?

The new `test_windows_x86` CI job runs automatically on push or when `run_rust_core` changes are detected. It builds `sf_core` for `i686-pc-windows-msvc`, verifies the DLL is 32-bit via `dumpbin`, and then runs the full test suite against that target.

### Why make this change?

Some consumers of `sf_core` run on 32-bit Windows environments. Without explicit CI coverage, regressions on x86 targets go undetected. The `sonic-rs` dependency does not support x86 Windows, so it needed to be made optional with a `serde_json` fallback to unblock the build. The pointer cast fix was necessary to avoid undefined behavior when deserializing pointers on 32-bit platforms where `usize` is 4 bytes but the serialized pointer is stored as 8 bytes.

PR #1062, opened 2026-05-06
ci: extract setup-msvc composite action; fix Windows x86 test on windows-2025-vs2026

sfc-gh-makowalski:
## Why

`test (Windows x86)` (in `test-rust-core.yml`) has been failing on every PR and on `main` since **2026-05-05 ~12:30 UTC** with:

```
'dumpbin' is not recognized as an internal or external command,
operable program or batch file.
##[error]Process completed with exit code 255.
```

The job hard-coded the path to VS 2022's `vcvarsall.bat`. GitHub started redirecting `windows-latest` (`windows-2025`) to `windows-2025-vs2026` on 2026-05-05 (rollout completes 2026-05-12); that image only ships VS 2026, so the hard-coded path silently disappears. `cmd`'s `call` to a missing `.bat` does **not** fail, so the job continues with an empty MSVC environment. `cargo build` still succeeds (Rust auto-detects MSVC via the registry), then `dumpbin` blows up because the toolchain isn't on `PATH`.

The Windows x86 job and the `dumpbin` check were introduced in #987 (commit 2a741ec, 2026-05-04). The job was green for ~24 hours before GitHub's runner rollout broke it.

## What changed

A code review of the initial fix flagged that:

1. The same hard-coded `vcvarsall.bat` path appeared in **five** other places across the repo, all carrying the same time bomb (they only stay green because `windows-11-arm` hasn't been migrated yet).
2. The repo already has a more thorough MSVC setup pattern in `test-odbc.yml`, but it was a one-off pwsh blob — there was no shared building block.

This PR consolidates everything into a single composite action.

### New: `./.github/actions/setup-msvc`

Reusable composite action with input `arch` (e.g. `x86`, `x64`, `arm64`, `amd64_arm64`). It:

- Locates Visual Studio dynamically via `vswhere -latest -property installationPath` — works across VS major versions, no hard-coded paths.
- Sources `vcvarsall.bat`, captures its output to a variable, and inspects `$LASTEXITCODE` **before** the pipeline resets it (a piped `cmd /c ... | ForEach-Object` would mask a real failure).
- Propagates every env var `vcvarsall` set to **both** `$GITHUB_ENV` (visible to subsequent steps in the job) **and** the current PowerShell process (so the in-step post-condition checks see the freshly-updated `PATH`).
- Verifies `cl.exe`, `link.exe`, and `dumpbin.exe` are all on `PATH`, and that `VSCMD_ARG_TGT_ARCH` matches the requested target (handles cross-compile syntax like `amd64_arm64`).
- Fails fast with a clear, named error if any precondition isn't met. `vcvarsall.bat` can exit 0 with a broken environment (missing components, wrong host arch); without these checks the next build step fails several minutes later with a confusing missing-tool / linker error.

### Call sites migrated

All five hard-coded `Microsoft Visual Studio\2022\Enterprise\...\vcvarsall.bat` paths in the repo are now gone:

| File | Step | Before | After |
|---|---|---|---|
| `test-rust-core.yml` | `test_windows_x86` → Build and test | inline `call "...vcvarsall.bat" x86` | `uses: ./.github/actions/setup-msvc` (`arch: x86`) |
| `test-rust-core.yml` | `test_windows_x86` → Rerun tests | inline `call ...` | env from prior step (`$GITHUB_ENV` persists) |
| `test-rust-core.yml` | `test_windows_arm64_nonfips` → Build and test | inline `call "...vcvarsall.bat" arm64` | `uses: ./.github/actions/setup-msvc` (`arch: arm64`) |
| `test-rust-core.yml` | `test_windows_arm64_nonfips` → Rerun tests | inline `call ...` | env from prior step |
| `test-rust-core.yml` | `test_windows_arm64_nonfips` → Install ARM64 static OpenSSL | inline `call ...` | `uses: ./.github/actions/setup-msvc` (cache-miss-only) |
| `_build-python-wheels.yml` | `build_wheel` (Windows ARM64) → Build sf_core | inline `call ...` | `uses: ./.github/actions/setup-msvc` (`arch: arm64`) |
| `_build-python-wheels.yml` | `build_wheel` (Windows ARM64) → Build python wheel | inline `call ...` | env from prior step |
| `setup-windows-openssl/action.yml` | `Install OpenSSL via vcpkg (ARM64)` | inline `call ...` | `uses: ./.github/actions/setup-msvc` (cache-miss-only) |

The redundant per-step inline guards (`if not "%VSCMD_ARG_TGT_ARCH%"=="arm64"`, `where dumpbin >nul`) are dropped — the composite action enforces them once, centrally, for all consumers.

### Net effect on `test-rust-core.yml`

`66 lines removed, 33 lines added` for the workflow itself; `+93 / -0` for the new composite action. Each Windows job is now significantly smaller and reads as: setup MSVC → build → assert arch → test.

## Out of scope (intentional)

- **`test-odbc.yml`**'s `Setup MSVC environment (Windows)` step does the same vswhere dance plus ODBC-specific CMake-PATH re-prepending. Consolidating it onto the new action is a separate refactor and would expand the blast radius of this PR. Filed as a TODO follow-up.

## Why the cleanup matters even though only x86 is currently broken

The four non-x86 sites only stay green today because `windows-11-arm` hasn't yet had a similar VS-version refresh. Same trigger, same silent-`call`-no-op failure mode. Fixing them now removes the time bomb and cuts code duplication in the same change.

## Test plan

- [x] `test (Windows x86)` is green on this PR (the breakage being fixed).
- [ ] `test (Windows ARM64, non-FIPS)` remains green (now driven through the new action).
- [ ] Python ARM64 wheel build (`_build-python-wheels.yml::build_wheel` on `windows-11-arm`) remains green (now driven through the new action; only triggered when wheel-building paths run).
- [ ] `setup-windows-openssl` ARM64 path (cache-miss-only, exercised on cold runners) remains green.

### DIFF
--- a/.github/actions/setup-msvc/action.yml
+++ b/.github/actions/setup-msvc/action.yml
@@ -1,0 +1,93 @@
+name: Setup MSVC
+description: >
+  Locate Visual Studio via vswhere, source vcvarsall.bat for the requested
+  target architecture, and propagate the resulting environment to subsequent
+  steps. Fails fast if the toolchain isn't actually on PATH afterwards.
+
+  This exists because hard-coding e.g.
+  `C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsall.bat`
+  silently breaks whenever GitHub rolls out a new runner image with a different
+  VS major version (e.g. the windows-2025 -> windows-2025-vs2026 migration that
+  shipped only VS 2026). `cmd`'s `call` to a missing .bat does NOT fail, which
+  produces a cryptic `'dumpbin' is not recognized` error after a successful
+  Rust-auto-detected build instead of failing fast.
+
+inputs:
+  arch:
+    description: >
+      Architecture argument passed to vcvarsall.bat. Examples: x86, x64, amd64,
+      arm64, x86_arm64, amd64_arm64. The action verifies that
+      VSCMD_ARG_TGT_ARCH ends up matching the target portion of this value.
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Locate MSVC and source vcvarsall.bat
+      shell: pwsh
+      run: |
+        $ErrorActionPreference = "Stop"
+        $arch = "${{ inputs.arch }}"
+
+        $vswhere = "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe"
+        if (-not (Test-Path $vswhere)) {
+          throw "vswhere.exe not found at $vswhere"
+        }
+
+        $installPath = & $vswhere -latest -property installationPath
+        if (-not $installPath) {
+          throw "vswhere could not locate a Visual Studio installation"
+        }
+
+        $vcvars = Join-Path $installPath "VC\Auxiliary\Build\vcvarsall.bat"
+        if (-not (Test-Path $vcvars)) {
+          throw "vcvarsall.bat not found at $vcvars"
+        }
+
+        # Capture vcvarsall output to a variable BEFORE inspecting $LASTEXITCODE.
+        # Piping `cmd /c ... | ForEach-Object` would succeed as a pipeline even if
+        # vcvarsall.bat itself failed, leaving later steps running without cl.exe
+        # or link.exe on PATH. See test-odbc.yml's Setup MSVC step for the same
+        # rationale.
+        $vcvarsOutput = cmd /c "`"$vcvars`" $arch && set" 2>&1
+        if ($LASTEXITCODE -ne 0) {
+          Write-Host "vcvarsall.bat output:"
+          $vcvarsOutput | ForEach-Object { Write-Host $_ }
+          throw "vcvarsall.bat $arch failed with exit code $LASTEXITCODE"
+        }
+
+        # Propagate every env var vcvarsall set to BOTH:
+        #   - $env:GITHUB_ENV     -> visible to subsequent steps in the job
+        #   - the current process -> so the post-condition checks below see
+        #                            the freshly-updated PATH (writes to
+        #                            GITHUB_ENV are buffered and only applied
+        #                            at step boundary).
+        $vcvarsOutput | ForEach-Object {
+          if ($_ -match '^([^=]+)=(.*)$') {
+            $name = $Matches[1]
+            $value = $Matches[2]
+            "$name=$value" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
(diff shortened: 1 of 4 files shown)


.github/actions/setup-msvc/action.yml
```diff
@@ -0,0 +1,93 @@
...
+        $vcvars = Join-Path $installPath "VC\Auxiliary\Build\vcvarsall.bat"
+        if (-not (Test-Path $vcvars)) {
+          throw "vcvarsall.bat not found at $vcvars"
+        }
+
+        # Capture vcvarsall output to a variable BEFORE inspecting $LASTEXITCODE.
+        # Piping `cmd /c ... | ForEach-Object` would succeed as a pipeline even if
+        # vcvarsall.bat itself failed, leaving later steps running without cl.exe
+        # or link.exe on PATH. See test-odbc.yml's Setup MSVC step for the same
+        # rationale.
```
sfc-gh-pbulawa:
```suggestion
        # or link.exe on PATH.
```

.github/actions/setup-msvc/action.yml
```diff
@@ -0,0 +1,93 @@
...
+  steps. Fails fast if the toolchain isn't actually on PATH afterwards.
+
+  This exists because hard-coding e.g.
+  `C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsall.bat`
+  silently breaks whenever GitHub rolls out a new runner image with a different
+  VS major version (e.g. the windows-2025 -> windows-2025-vs2026 migration that
+  shipped only VS 2026). `cmd`'s `call` to a missing .bat does NOT fail, which
+  produces a cryptic `'dumpbin' is not recognized` error after a successful
+  Rust-auto-detected build instead of failing fast.
+
```
sfc-gh-pbulawa:
```suggestion
```

.github/actions/setup-msvc/action.yml
```diff
@@ -0,0 +1,93 @@
...
+          $vcvarsOutput | ForEach-Object { Write-Host $_ }
+          throw "vcvarsall.bat $arch failed with exit code $LASTEXITCODE"
+        }
+
+        # Propagate every env var vcvarsall set to BOTH:
+        #   - $env:GITHUB_ENV     -> visible to subsequent steps in the job
+        #   - the current process -> so the post-condition checks below see
+        #                            the freshly-updated PATH (writes to
+        #                            GITHUB_ENV are buffered and only applied
+        #                            at step boundary).
```
sfc-gh-pbulawa:
```suggestion
```

sfc-gh-pbulawa (approved):

PR #1063, opened 2026-05-06
Add WrapperPresets mechanism for per-wrapper behavioral configuration

sfc-gh-pczajka:
Each wrapper (Python, ODBC, JDBC) declares its identity as a string at startup. Rust builds an immutable WrapperPresets struct from defaults, overriding only what differs for that wrapper. First preset: put_get_resultset_flavor (Python vs Odbc column layouts).

Decision grounds:
 * all presets live in one place - `for_wrapper` constructor. Easy to find, develop and maintain
 * pushing flag differences to "init" in all languages caused the code to be unnecessarily complicated

This PR does not implement PUT/GET result logic, just passed WrapperPresets object reference as dummy.

### DIFF
--- a/jdbc_bridge/src/lib.rs
+++ b/jdbc_bridge/src/lib.rs
@@ -6,7 +6,7 @@ use jni::sys::{jint, jobject};
 use proto_utils::{ProtoError, Transport};
 use sf_core::logging::LogManager;
 use sf_core::protobuf::apis::RustTransport;
-use sf_core::protobuf::apis::database_driver_v1::DriverProviders;
+use sf_core::protobuf::apis::database_driver_v1::{DriverProviders, WrapperPresets};
 use sf_core::telemetry::snowflake_exporter::SessionRegistry;
 
 static JDBC_LOG_MANAGER: Mutex<Option<LogManager>> = Mutex::new(None);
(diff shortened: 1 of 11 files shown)


jdbc_bridge/src/lib.rs
```diff
@@ -23,6 +23,7 @@ impl JdbcBridge {
                 .lock()
                 .unwrap_or_else(|e| e.into_inner())
                 .take(),
+            wrapper_presets: WrapperPresets::for_wrapper("jdbc"),
```
sfc-gh-jszczerbinski, line 26: Use named constructors instead of string argument - `WrapperPresets::jdbc()`

sfc-gh-jszczerbinski (approved):
LGTM, but let's not use string as identifier

jdbc_bridge/src/lib.rs
sfc-gh-ddas: Each wrapper can also construct `WrapperPresets` at its own init site

sfc-gh-ddas (approved):

jdbc_bridge/src/lib.rs
sfc-gh-pczajka: That was the problem - both of these approaches required complicated construction to pass python-C-Rust bridge, requiring reconstructing whole WrapperPresets struct on Python side, defying point of having default at all.
This approaches are easy in languages that support Rust extensions, but ones using c_api.rs will be messy.

This is a golden mean - there is no much risk in using pure strings instead of enums as this needs to be passed once and won't change.

sfc-gh-pczajka: Fixed after discussion

PR #1068, opened 2026-05-07
feat(odbc): honor TIMESTAMP_TZ_OUTPUT_FORMAT for TZ -> CHAR/WCHAR fetch

sfc-gh-makowalski:
Closes the gap left by #1005: when the session has
`TIMESTAMP_TZ_OUTPUT_FORMAT` set to a value containing a `TZH:TZM`,
`TZHTZM`, or `TZH` token, the TIMESTAMP_TZ -> SQL_C_CHAR / SQL_C_WCHAR
fetch path now appends the offset suffix (`+/-HH:MM`, `+/-HHMM`, or
`+/-HH`) to the wall-clock literal, matching what the legacy 3.16.0
driver and the Snowflake server itself produce for the same format.

Behaviour is **opt-in** by design: an empty / unset / no-TZ-token format
keeps the legacy UTC-only fetch behaviour, so existing applications that
parse the bare `YYYY-MM-DD HH:MM:SS` shape see no change. The toggle is
also CHAR/WCHAR-only — `SQL_C_TYPE_TIMESTAMP` has no offset field and
the spec still requires UTC there.

Implementation
--------------
- `TzOffsetFormat` enum + `parse_tz_offset_format` parser in
  `odbc/src/conversion/timestamp.rs` recognise the longest matching
  Snowflake offset token (case-insensitive). `TZH` falls back to
  `+HH:MM` for sub-hour offsets so we never silently drop minutes.
- `NumericSettings` gains a `tz_offset_format: Option<TzOffsetFormat>`
  field. `update_numeric_settings` reads `TIMESTAMP_TZ_OUTPUT_FORMAT`
  on every execute via the existing `connection_get_parameter` RPC, so
  in-flight `ALTER SESSION SET ...` takes effect for the next query.
- `SnowflakeTimestampTz` carries the format into the converter; its
  `write_odbc_type` routes CHAR/WCHAR to a new
  `format_timestamp_tz_string_into` helper that emits the local
  wall-clock (`utc + offset_minutes`) plus the offset suffix.
- `column_size` widens by 7 chars (worst-case `+ +HH:MM`) when the
  format is set so apps that size buffers from the descriptor are safe.

Test coverage
-------------
- Unit: 12 new tests in `timestamp_tests.rs` cover the parser, all
  three `TzOffsetFormat` variants, negative / zero / sub-hour offsets,
  WCHAR encoding, buffer-too-small (`22003`), and that
  `SQL_C_TYPE_TIMESTAMP` keeps ignoring the format. Existing UTC-only
  tests are untouched (still pass with `tz_offset_format: None`).
- E2E: `timestamp_tz_output_format.cpp` does `ALTER SESSION SET
  TIMESTAMP_TZ_OUTPUT_FORMAT = '... TZH:TZM'` (and `TZHTZM`), confirms
  the offset comes back through SQLGetData for both `SQL_C_CHAR` and
  `SQL_C_WCHAR`. Skipped on the legacy reference driver since that's
  the behaviour we're mirroring (BD#000 placeholder, no semantic
  difference once both honour the format).

Co-authored-by: Cursor <cursoragent@cursor.com>


sfc-gh-makowalski:
* **#1068** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1068?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1068?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

odbc/src/conversion/timestamp.rs
```diff
@@ -31,6 +31,49 @@ use crate::conversion::{ReadArrowType, SnowflakeType, WriteODBCType};
...
+///
+/// The longest token wins (`TZH:TZM` and `TZHTZM` both contain `TZH`).
+/// Match is case-insensitive to mirror Snowflake's format grammar.
+pub(crate) fn parse_tz_offset_format(format: &str) -> Option<TzOffsetFormat> {
+    let upper = format.to_ascii_uppercase();
+    if upper.contains("TZH:TZM") {
+        Some(TzOffsetFormat::Colon)
+    } else if upper.contains("TZHTZM") {
+        Some(TzOffsetFormat::NoColon)
+    } else if upper.contains("TZH") {
```
sfc-gh-rkowalski: This is an unanchored substring scan, but the doc-comment above sells it as matching "Snowflake's format grammar". A real Snowflake format can contain double-quoted literals — e.g. `'"server-side TZH note: " YYYY-MM-DD HH24:MI:SS'` — and the literal-internal `TZH` would activate `HourOnly` here. Strings like `'TZHACK'`, `'TZHELP'`, `'YYYY-MM-DD HH24:MI:SS TZHIRD'` also false-match. Since the toggle changes the bytes the user sees on the wire, false-positive activation is a correctness bug, not just a warning concern.

Unit test coverage on adversarial inputs is also thin — only obvious negatives like `"any string with TZ but no H"` and bare `Z`.

Suggested: tokenize at non-alphanumeric boundaries and check whole-token equality (or a regex like `\b(TZH:TZM|TZHTZM|TZH)\b` with longest-match-wins), and strip Snowflake double-quoted literal runs (`"..."`) before scanning. Then add tests pinning at least:

- `'"comment with TZH" YYYY-MM-DD'` → `None`
- `'TZHACK'` → `None`
- `'literal_TZH_marker'` → `None`
- `'TZH:TZM TZHTZM'` → `Colon` (longest still wins)

While we're here, the parser also silently ignores legitimate Snowflake offset tokens like `TZHM` (no colon) and bare `TZM` — same tokenizer pass can recognize them, or at minimum log a warning when an offset-shaped token isn't one of the three we accept.

odbc/src/conversion/timestamp.rs
```diff
@@ -815,17 +947,74 @@ impl WriteODBCType for SnowflakeTimestampTz {
...
+    const MIN_CHAR_BYTES: sql::Len = 26;
+    const MIN_WCHAR_BYTES: sql::Len = MIN_CHAR_BYTES * 2;
+
+    let target = binding.target_type;
+    let min = if matches!(target, CDataType::WChar) {
+        MIN_WCHAR_BYTES
+    } else {
+        MIN_CHAR_BYTES
+    };
+    if binding.buffer_length > 0 && binding.buffer_length < min {
```
sfc-gh-rkowalski: This pre-emptive check returns SQLSTATE 22003 ("Numeric value out of range") for a too-small CHAR/WCHAR buffer. That's the wrong category — 22003 is reserved for *numeric* overflow. The spec for `SQLGetData`/`SQLFetch` on a too-small character buffer is **01004 "String data, right truncation"** with `SQL_SUCCESS_WITH_INFO`: write `buffer_length-1` bytes, NUL-terminate, set the indicator to the full untruncated length so the app can resize and reissue.

The right plumbing is already in this crate — `write_char_string` / `write_wchar_string` emit `Warning::StringDataTruncated` → `SqlState::StringDataRightTruncated`, which is what every other CHAR converter uses (e.g. `number.rs:392` for fractional truncation).

Worse, this also rejects buffers that *would have fit* a shorter rendering — `HourOnly` whole-hour is ~23 chars, but a 25-byte buffer is rejected up front.

Suggested: drop this whole pre-check. Always render into the 64-byte stack buffer and let `write_char_string` / `write_wchar_string` apply the standard 01004 truncation contract. The unit test (`write_tz_char_with_format_buffer_too_small_returns_22003`) needs a rename and should switch from `format!("{err:?}").contains("Buffer too small")` to `matches!(err, WriteOdbcError::StringDataTruncated { .. })` — string-matching the Debug repr breaks silently on any rename.

odbc_tests/tests/e2e/types/timestamp_tz_output_format.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include <catch2/catch_test_macros.hpp>
+
+#include "Connection.hpp"
+#include "compatibility.hpp"
+#include "conversion_checks.hpp"
+#include "get_diag_rec.hpp"
+
+TEST_CASE("TIMESTAMP_TZ to SQL_C_CHAR honors TIMESTAMP_TZ_OUTPUT_FORMAT with TZH:TZM",
+          "[timestamp_tz][conversion][c_char][output_format]") {
+  SKIP_OLD_DRIVER("BD#000",
```
sfc-gh-rkowalski: `BD#000` placeholder shipped to all three SKIP_OLD_DRIVER calls in this file (also lines 70 and 84). Every other `SKIP_OLD_DRIVER` in the suite references a real BD entry (BD#17, BD#18, BD#30, BD#33, BD#46, BD#49) — and the prior PR on this branch (`f10ab16c`) explicitly introduced BD#49 to fix exactly this discipline. Letting `BD#000` ship makes the eventual legacy-driver-deprecation cleanup a guessing game and any "which test gates which legacy gap?" audit impossible.

The commit message also implies the legacy 3.16.0 driver honors the format identically — if that's true, the SKIP itself is unjustified and these tests should run against both drivers asserting parity. Either way, please file a real BD entry (or reuse an existing one documenting "legacy 3.16.0 honors TIMESTAMP_TZ_OUTPUT_FORMAT identically; tests gated only because the sample DB lacks the parameter") and replace all three placeholders before merge.

odbc_tests/tests/e2e/types/timestamp_tz_output_format.cpp
```diff
@@ -0,0 +1,102 @@
...
+                  "New driver opts into legacy `TZH:TZM`-aware TZ rendering only when the session "
+                  "format is set; legacy driver's behaviour for the same format is what we mirror");
+  // Given Snowflake client is logged in
+  Connection conn;
+
+  // When the session opts into the verbose Snowflake offset token
+  conn.execute("ALTER SESSION SET TIMESTAMP_TZ_OUTPUT_FORMAT = 'YYYY-MM-DD HH24:MI:SS.FF TZH:TZM'");
```
sfc-gh-rkowalski, line 34: The load-bearing claim of this PR is that `update_numeric_settings` re-reads the parameter on **every execute**, so an `ALTER SESSION` mid-connection takes effect on the next query. That's the entire reason we're paying a per-execute sequential RPC. But every TEST_CASE in this file does exactly one ALTER followed by one fetch — a connection-time-only read would pass these tests identically. Same gap for `ALTER SESSION UNSET` reverting to bare UTC.

Would you mind adding one TEST_CASE that exercises the actual claim? Something like:

1. Initial fetch with no format set → assert bare UTC.
2. `ALTER SESSION SET TIMESTAMP_TZ_OUTPUT_FORMAT = '... TZH:TZM'`
3. Fetch → assert `+HH:MM` suffix.
4. `ALTER SESSION SET ... TZHTZM`
5. Fetch → assert `+HHMM`.
6. `ALTER SESSION UNSET TIMESTAMP_TZ_OUTPUT_FORMAT`
7. Fetch → assert bare UTC.

This is the only test shape that actually pins the per-execute reread.

odbc/src/api/statement.rs
```diff
@@ -189,6 +189,31 @@ fn update_numeric_settings(
...
+                conn_handle: Some(*conn_handle),
+                key: "TIMESTAMP_TZ_OUTPUT_FORMAT".to_string(),
+            })
+            .await
+        {
+            resp.value
+                .as_deref()
+                .and_then(crate::conversion::timestamp::parse_tz_offset_format)
+        } else {
+            None
```
sfc-gh-rkowalski: On a single transient RPC failure here, we silently overwrite `settings.tz_offset_format = None`, clobbering a previously cached `Some(Colon)` and flipping the next fetch from `+HH:MM` rendering back to bare-UTC — with no diagnostic the application can correlate to anything.

The other three settings in this function follow the same pattern, but they have meaningful server-side defaults so resetting to default is harmless. `tz_offset_format` doesn't — the customer set it deliberately via `ALTER SESSION`.

Suggested: on `Err(_)`, leave `settings.tz_offset_format` unchanged and `tracing::warn!("failed to refresh TIMESTAMP_TZ_OUTPUT_FORMAT; keeping cached value")`. Same on `resp.value.is_none()` / empty string. Only overwrite on a real successful response with non-empty value, and add a comment documenting the chosen semantic so the next reader doesn't re-introduce the silent reset.

Three-line change.

sfc-gh-rkowalski (changes requested):
Lots of this looks good — the local-time math is correct, sign handling is clean (`+00:00` not `-00:00`, `unsigned_abs()` for the negative branch), the `TZH:TZM` longest-token preference is right, the default path with `tz_offset_format: None` is byte-identical to before, and the bind path is correctly untouched. WCHAR encoding via the existing helper is fine.

Five things I'd like to see addressed before this merges. They're not refactors, they're surgical fixes — details inline. The biggest one in my mind is H4 (no e2e actually proves the per-execute reread that justifies the new RPC), and H5 (transient RPC failure silently flips a customer-set format back to `None`).

odbc/src/conversion/timestamp.rs
sfc-gh-makowalski: Done in ad632446b. Rewrote `parse_tz_offset_format` to: (a) strip Snowflake `"..."` literal runs first; (b) split the remainder on non-(alnum|underscore) boundaries so `_TZH_` is one identifier-shaped token, not three; (c) detect `TZH:TZM` via a colon-pair sequence with whole-token boundary anchoring (so `XTZH:TZMX` doesn't false-fire); (d) match `TZHTZM` and `TZH` via whole-token equality. `tracing::warn!` fires on unrecognised TZ-shaped tokens (`TZHM`, `TZM`, etc.) so a customer who configured a variant the driver doesn't render gets a log signal. Pinning tests added in `timestamp_tests.rs` cover all four adversarial cases you listed plus the boundary-anchored colon check, the longest-match tiebreaker, and the unrecognised-token return path. Resolving.

odbc/src/conversion/timestamp.rs
sfc-gh-makowalski: Done in ad632446b. Dropped the pre-emptive size check in `write_timestamp_tz_to_char` -- the standard 64-byte stack buffer + `write_char_string`/`write_wchar_string` path now applies the spec-mandated 01004 (`StringDataRightTruncated`) + `SQL_SUCCESS_WITH_INFO` + indicator-set-to-untruncated-length contract. Renamed the test to `write_tz_char_with_format_buffer_too_small_truncates_with_warning` and switched it from `format!(\"{err:?}\").contains(\"Buffer too small\")` to a `matches!` against `Warning::StringDataTruncated` plus indicator-length and buffer-contents asserts (rename-proof). Resolving.

odbc_tests/tests/e2e/types/timestamp_tz_output_format.cpp
sfc-gh-makowalski: Per offline decision, taking the "Microsoft spec compliance first, legacy parity second" path: dropped all three `SKIP_OLD_DRIVER` calls so the tests run against both drivers as parity assertions (ad632446b). If legacy 3.16.0 actually diverges in CI we'll file a real BD entry then; until then, asserting parity on both sides catches accidental regressions. No more `BD#000` placeholders in the tree. Resolving.

odbc_tests/tests/e2e/types/timestamp_tz_output_format.cpp
sfc-gh-makowalski: Done in ad632446b. New `TEST_CASE` toggles `TIMESTAMP_TZ_OUTPUT_FORMAT` four times within one connection (no-format -> `TZH:TZM` -> `TZHTZM` -> `UNSET`) and asserts each fetch reflects the latest setting. A connection-time-only cache would fail at phase 2 (still bare UTC), and a cache that never re-checked after a non-empty value would fail at phase 4 (still emitting the offset). Pins both the SET and UNSET reread paths. Resolving.

odbc/src/api/statement.rs
sfc-gh-makowalski: Done in ad632446b. Three-line semantic change as suggested: `Err(_)` now leaves `settings.tz_offset_format` untouched and emits `tracing::warn!` instead of silently resetting to `None`. `Ok(Some(\"\"))` / `Ok(None)` still clear the cache (explicit UNSET). Extracted the decision into a pure `apply_tz_offset_format_update` function with a six-case unit-test table covering: `Ok` recognised / unrecognised / empty / None and `Err` against both empty and populated caches. Resolving.

sfc-gh-rkowalski (approved):

PR #1069, opened 2026-05-07
SNOW-3480688  log_query_text and log_query_params parameters to control query for the INFO level (by default turned off)

sfc-gh-mhofman:


sfc-gh-mhofman:
* **#1074** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1074?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1069** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1069?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1069?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sf_core/src/rest/snowflake/mod.rs
```diff
@@ -974,12 +1011,10 @@ async fn execute_async_with_fallback<'a>(
                 ..
             },
         ) => {
+            let (sql, bindings) = query_log_fields(query_parameters, &query_input);
             tracing::error!(
```
sfc-gh-asolarski, line 1020: now in case of the error we won't even know which query/request caused it, comparing to `execute_sync_with_retry` where we log at least the `request_id` here we are completely blind

maybe `RestError::AsyncQuery` should have this request_id from the inside of `snowflake_query_async_style`, so we can somehow identify this log

wdyt 🤔

sfc-gh-asolarski (approved):
generally, the direction looks fine to me

I've just posted one doubt about what we currently log

sf_core/src/rest/snowflake/mod.rs
sfc-gh-mhofman: Good call out - I've extended AsyncQuerySnafu to transport request_id/query_id so at least this would be visible

PR #1070, opened 2026-05-07
SNOW-3355389 optimize perf PR check

sfc-gh-akolodziejczyk:
- Split ODBC and PYTHON build and test to separate nodes
- only run for valid changes

### DIFF
--- a/tests/performance/Jenkinsfile.pr-check
+++ b/tests/performance/Jenkinsfile.pr-check
@@ -16,8 +16,20 @@ import groovy.transform.Field
 @Field def REGRESSION_STATUS_CONTEXT = 'performance/regression'
 @Field def REGRESSION_EXIT_CODE = 77
 
-// Track whether the Python+ODBC lane detected a confirmed regression
+// Track whether any lane detected a confirmed regression
 @Field def regressionDetected = false
+@Field def pythonRegressionDetected = false
+@Field def odbcRegressionDetected = false
+
+// Change detection: which drivers need perf testing
+@Field def runCore = false
+@Field def runPython = false
+@Field def runOdbc = false
+@Field def skipAll = false
+
+def isMergeQueue() {
+    return env.BRANCH_NAME?.startsWith('gh-readonly-queue/')
+}
 
 def setupVenv() {
     dir('tests/performance') {
@@ -157,18 +169,24 @@ ${sep}
 PERFORMANCE REGRESSION CHECK SUMMARY
 ${sep}
   Driver   Test group   Result
-  ------   ----------   ------
-  Python   Fast         ${exitCodeLabel(python.exitFast)}
-  Python   Slow         ${exitCodeLabel(python.exitSlow)}
-  ODBC     Fast         ${exitCodeLabel(odbc.exitFast)}
-  ODBC     Slow         ${exitCodeLabel(odbc.exitSlow)}
-${sep}
+  ------   ----------   ------"""
+    if (python) {
+        echo """  Python   Fast         ${exitCodeLabel(python.exitFast)}
+  Python   Slow         ${exitCodeLabel(python.exitSlow)}"""
+    }
+    if (odbc) {
+        echo """  ODBC     Fast         ${exitCodeLabel(odbc.exitFast)}
+  ODBC     Slow         ${exitCodeLabel(odbc.exitSlow)}"""
+    }
+    echo """${sep}
   Fast tests : ${REGRESSION_FAST_TEST_FILTER}
   Slow tests : ${REGRESSION_SLOW_TEST_FILTER}
   Threshold  : ${REGRESSION_THRESHOLD}%  |  Rerun iters: ${REGRESSION_RERUN_ITERATIONS}
 ${sep}"""
 
     for (driver in ['python', 'odbc']) {
+        if (driver == 'python' && !python) continue
+        if (driver == 'odbc' && !odbc) continue
         def summaryFile = "results/regression_summary_${driver}.json"
         if (!fileExists(summaryFile)) {
             echo "  [${driver.toUpperCase()}] No regression summary file found (${summaryFile})"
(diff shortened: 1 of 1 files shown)


sfc-gh-akolodziejczyk:
* **#1070** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1070?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1070?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tests/performance/Jenkinsfile.pr-check
```diff
@@ -253,7 +352,38 @@ pipeline {
                                 }
                             }
                         }
-                        stage('Smoke Test - Core') {
+                    }
+                }
```
sfc-gh-makowalski, line 380: **The `Smoke Test - Core` stage was dropped in this refactor and isn't mentioned in the PR description.**

Before this PR, the `Core` lane had three stages: `Setup`, `Build`, and `Smoke Test - Core` — the last one ran:

```
python -m hatch run -- python -m pytest tests/test_select_1M.py::test_select_string_1M_arrow --driver=core --iterations=1 --warmup-iterations=0 -s -v
```

so the lane actually verified the core driver could execute. After this PR the `Core` lane only runs `Setup` and `Build` — it confirms `core/build.sh` and `wiremock/build.sh` succeed but never executes the driver. So a regression that breaks core at runtime (but not at link-time) won't be caught by this lane anymore.

If this was intentional (e.g., "Python+ODBC lanes already exercise core via FFI, the standalone smoke is redundant"), please call that out in the PR body. If it wasn't, please restore it — diff suggestion below.

```suggestion
                    agent { label 'drivers-perf-regular-memory-node-snowos' }
                    stages {
                        stage('Setup - Core') {
                            steps {
                                setupVenv()
                            }
                        }
                        stage('Build - Core + WireMock') {
                            steps {
                                withCredentials([
                                    usernamePassword(
                                        credentialsId: '063fc85b-62a6-4181-9d72-873b43488411',
                                        usernameVariable: 'AWS_ACCESS_KEY_ID',
                                        passwordVariable: 'AWS_SECRET_ACCESS_KEY'
                                    )
                                ]) {
                                    sh 'tests/performance/drivers/core/build.sh'
                                    sh 'tests/performance/wiremock/build.sh'
                                }
                            }
                        }
                        stage('Smoke Test - Core') {
                            steps {
                                withCredentials([
                                    string(credentialsId: 'sfctest0-parameters-secret', variable: 'PARAMETERS_SECRET')
                                ]) {
                                    sh './scripts/decode_secrets.sh'
                                }
                                dir('tests/performance') {
                                    withCredentials([
                                        file(credentialsId: '22401612-0c52-4682-8d00-fb9ee98927ac', variable: 'SF_PERF_CONFIG')
                                    ]) {
                                        sh '''
                                            source .venv/bin/activate
                                            export CLOUD=aws
                                            python -m hatch run -- python -m pytest tests/test_select_1M.py::test_select_string_1M_arrow --driver=core --iterations=1 --warmup-iterations=0 -s -v
                                        '''
                                    }
                                }
                            }
                        }
                    }
                }
```

tests/performance/Jenkinsfile.pr-check
```diff
@@ -264,30 +394,47 @@ pipeline {
...
+                                                def python = runRegressionTests('python')
+
+                                                printRegressionSummary(python, null)
+
+                                                if (python.combined == REGRESSION_EXIT_CODE) {
+                                                    pythonRegressionDetected = true
+                                                    regressionDetected = true
+                                                    echo "Performance regression detected in Python (exit code ${REGRESSION_EXIT_CODE})"
+                                                } else if (python.combined != 0) {
+                                                    error("Python tests failed with exit code ${python.combined}")
```
sfc-gh-makowalski, line 447: **Hard test failure (non-regression) no longer posts a `FAILURE` status to GitHub.**

Before this PR, the lane explicitly handled the "tests failed before regression check could complete" branch by posting a FAILURE status *and then* erroring:

```groovy
} else if (exitCode != 0) {
    postRegressionStatus('FAILURE', 'Tests failed before regression check could complete')
    error("Tests failed with exit code ${exitCode}")
}
```

In the new code that branch only calls `error(...)` — the status post is gone. With `failFast: true` on the parallel block and `Regression Gate` moved out of the lane into a top-level stage, the gate **does not run** when a lane errors out (Jenkins skips it because the parallel block failed). Net result: the `performance/regression` GitHub commit-status check stays at whatever it was before this run — typically `pending` on first run — and branch protection blocks the merge with "expected status check is missing" rather than a clear failure message.

This applies symmetrically to the ODBC lane below at line 477.

Two reasonable fixes:

1. **Cheapest:** restore the inline FAILURE post (mirrors the old behaviour, keeps the gate simple). Suggestion below for the Python lane; please apply the same to ODBC.
2. **Cleaner long-term:** wrap the gate logic in a pipeline-level `post { always { script { ... } } }` so it runs regardless of how the parallel block exited. Bigger change, worth a separate discussion.

```suggestion
                                                } else if (python.combined != 0) {
                                                    postRegressionStatus('FAILURE', 'Python tests failed before regression check could complete')
                                                    error("Python tests failed with exit code ${python.combined}")
                                                }
```

sfc-gh-makowalski (changes requested):
Two issues that I think need to be addressed before merge — both look like unintended consequences of the refactor rather than deliberate design choices. Otherwise the split-into-two-lanes idea and per-driver regression tracking are nice improvements. Inline below.

tests/performance/Jenkinsfile.pr-check
sfc-gh-makowalski: I see that it might have been done for performance gain, resolve if that's the case.

sfc-gh-makowalski (approved):

PR #1080, opened 2026-05-07
NO-SNOW: JSON->Arrow conversion for VECTOR type

sfc-gh-pczajka:
## Summary

Convert VECTOR cells arriving in JSON result format (JSON array strings like `"[1,2,3]"` or `"[1.5,2.5,3.5]"`) into Arrow `FixedSizeList(Int32|Float32, dimension)` arrays, matching the Arrow-format path.

Python VECTOR e2e tests previously skipped all JSON-format runs with `pytest.mark.skip_for_json_result_set` because `sf_core`'s JSON chunk parser only produced `Utf8` for VECTOR columns. After this change, only the FLOAT subnormal-boundary case is skipped (server-side JSON flushes subnormals to zero).

## Changes

- **`sf_core::query_types`** — `RowType::Vector` now carries `dimension: usize` and `element_type: VectorElementType { Int32, Float32 }`.
- **`sf_core::rest::snowflake::query_response`** — parse `vectorDimension` and `fields[0].type` from the server's `rowtype`; expose `fields`/`type_` previously marked unused. Extracted `parse_vector_row_type` helper. Test fixtures collapsed with `make_row_type` / `make_vector_row_type` helpers.
- **`sf_core::arrow_utils`** — emit `DataType::FixedSizeList(Int32|Float32, dimension)` for VECTOR, preserving the `logicalType=VECTOR` metadata so Python's `FixedSizeListConverter` handles it.
- **`sf_core::chunks::json_parser`** — new `ColumnBuilder::Vector` + `VectorValuesBuilder` parses each JSON-array cell into child elements and builds a `FixedSizeListArray`. Null outer rows append `dimension` placeholder child elements masked by the outer null buffer to keep the `child_len == outer_len * dimension` invariant.
- **`python/tests/e2e/types/test_vector.py`** — removed the module-level `pytestmark` skip; split the FLOAT boundary case so the non-subnormal values run in both formats, and the `FLOAT32_SMALLEST_NORMAL` case is a dedicated test skipped only in JSON format (server flushes subnormals to zero).

## Test plan

- [x] `cargo test -p sf_core --lib` — 716 passed (+4 new: VECTOR int values + nulls, VECTOR float values + nulls, dimension-mismatch, non-integer element, malformed JSON cell; and query_response: case-insensitive element type)
- [x] Python VECTOR e2e in Arrow format — 12 passed
- [x] Python VECTOR e2e in JSON format (`QUERY_RESULT_FORMAT=JSON`) — 11 passed, 1 skipped (FLOAT subnormals)
- [x] `cargo build --release` — clean

### DIFF
--- a/sf_core/src/chunks/json_parser.rs
+++ b/sf_core/src/chunks/json_parser.rs
@@ -259,7 +259,7 @@ pub(super) enum ColumnBuilder {
         builder: arrow::array::StringBuilder,
     },
     Vector {
-        dimension: i32,
+        dimension: usize,
         values: VectorValuesBuilder,
         nulls: arrow::array::BooleanBufferBuilder,
     },
@@ -338,8 +338,10 @@ impl ColumnBuilder {
             RowType::Text { .. }
             | RowType::Variant { .. }
             | RowType::Object { .. }
-            | RowType::Array { .. }
-            | RowType::Geography { representation, .. }
+            | RowType::Array { .. } => ColumnBuilder::Text {
+                builder: arrow::array::StringBuilder::with_capacity(capacity, capacity * 8),
+            },
+            RowType::Geography { representation, .. }
             | RowType::Geometry { representation, .. } => match representation {
                 crate::query_types::GeoRepresentation::Text => ColumnBuilder::Text {
                     builder: arrow::array::StringBuilder::with_capacity(capacity, capacity * 8),
@@ -353,9 +355,9 @@ impl ColumnBuilder {
                 element_type,
                 ..
             } => {
-                // Dimension is bounded to VECTOR_MAX_DIMENSION at parse time, so
-                // capacity * dimension fits in usize on all supported platforms.
-                let elem_capacity = capacity.saturating_mul(*dimension as usize);
+                // Snowflake bounds VECTOR dimension at 4096, so the multiplication
+                // fits in usize on all supported platforms.
+                let elem_capacity = capacity.saturating_mul(*dimension);
                 let values = match element_type {
                     VectorElementType::Int32 => VectorValuesBuilder::Int32(
                         arrow::array::PrimitiveBuilder::with_capacity(elem_capacity),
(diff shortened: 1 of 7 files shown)


sf_core/src/query_types.rs
```diff
@@ -89,9 +89,25 @@ pub enum RowType {
...
+/// Stored as an enum (rather than `arrow::DataType`) to keep this module independent
+/// of the Arrow crate.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum VectorElementType {
+    Int32,
+    Float32,
+}
+
+/// Maximum Snowflake VECTOR dimension (per SQL reference).
+pub const VECTOR_MAX_DIMENSION: i32 = 4096;
```
sfc-gh-asolarski: what's the purpose of "validating backend" in this case?  
it looks like we have this logic here but no real gain from it?

sfc-gh-asolarski: maybe even from the query response it should be already a `usize` not `int32` :thinking_face:

sf_core/src/chunks/json_parser.rs
```diff
@@ -875,6 +951,70 @@ fn empty_batch_from_row_types(row_types: &[RowType]) -> Result<Vec<RecordBatch>,
...
+        VectorValuesBuilder::Float32(builder) => {
+            for elem in elems {
+                let v = elem.as_f64().ok_or_else(|| {
+                    ArrowError::InvalidArgumentError(
+                        "VECTOR(FLOAT) element must be a JSON number".to_string(),
+                    )
+                })?;
+                // Snowflake serializes VECTOR(FLOAT) elements already in f32 precision,
+                // so this narrowing cast is lossless for conformant server output.
+                builder.append_value(v as f32);
```
sfc-gh-asolarski: we have two different approaches here - for integers we try to downcast from i64 to i32 and throw is it fails  
but for floats we are saying it's pointless as `narrowing cast is lossless for conformant server output` so what's the point of this logic for ints?  
  
let's unify the approaches

sfc-gh-asolarski (changes requested):
I have some questions about the chosen approach of "validating backed" here

sf_core/src/query_types.rs
sfc-gh-pczajka: done

sf_core/src/chunks/json_parser.rs
sfc-gh-pczajka: done

sfc-gh-asolarski (approved):

PR #1082, opened 2026-05-08
SNOW-3098562: Add flag for enabling/disabling traces in error messages

sfc-gh-rkowalski:
## Summary

- Adds an `ErrorTraceEnabled` INI / `error_trace_enabled` TOML config key that controls whether `OdbcError::message_text()` appends the full error trace to user-facing diagnostics.    Defaults to `true`, preserving existing behavior; set to `false` to get a terse base message.
- `sf_core::logging::LoggingConfig` parses the value from INI/TOML and exposes it via `LogManager::error_trace_enabled()`.
- The odbc crate owns the runtime toggle (`odbc::api::error_trace_flag`), seeded at `SQLAllocHandle(ENV)` time from the `LogManager`. Other consumer crates (`jdbc_bridge`, Python    `sf_core_init`) are untouched — no cross-crate coupling through a sf_core-level static.
- Previously-unconditional `"\nTrace:\n<...>"` suffix is now conditional on the flag; label also updated to `"\nerror trace:\n<...>"`.

### DIFF
--- a/odbc/src/api/error_trace_flag.rs
+++ b/odbc/src/api/error_trace_flag.rs
@@ -3,8 +3,9 @@ use std::sync::atomic::{AtomicBool, Ordering};
 /// User-facing error-rendering policy toggle, seeded at env allocation time
 /// from `LogManager::error_trace_enabled()`. Defaults to `true` so errors
 /// produced before env init (lock poisoning, invalid handles) still print a
-/// full trace. `Relaxed` is sufficient: the flag is a standalone boolean that
-/// gates a formatting branch and publishes no other memory.
+/// full trace. `Relaxed` is sufficient: the flag is a standalone boolean that is
+/// initialized as a part of env allocation during sqlallochandle init.
+/// Nothing happens before that concludes.
 static ERROR_TRACE_ENABLED: AtomicBool = AtomicBool::new(true);
 
 /// Whether `OdbcError::message_text()` should append the full error trace.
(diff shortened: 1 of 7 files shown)


odbc/src/api/error_trace_flag.rs
```diff
@@ -0,0 +1,75 @@
...
+    ERROR_TRACE_ENABLED.store(value, Ordering::Relaxed);
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::api::OdbcError;
+    use std::sync::Mutex;
+
+    /// Serialises any test that reads or writes the process-wide
```
sfc-gh-mcepiga:
```suggestion
    /// Serializes any test that reads or writes the process-wide
```

odbc/src/api/error_trace_flag.rs
```diff
@@ -0,0 +1,75 @@
+use std::sync::atomic::{AtomicBool, Ordering};
+
+/// User-facing error-rendering policy toggle, seeded at env allocation time
+/// from `LogManager::error_trace_enabled()`. Defaults to `true` so errors
+/// produced before env init (lock poisoning, invalid handles) still print a
+/// full trace. `Relaxed` is sufficient: the flag is a standalone boolean that
```
sfc-gh-mcepiga: While I agree with the conclusion, I don't understand this rationale - "flag is a standalone boolean that gates a formatting branch and publishes no other memory.".

Maybe I'm missing something, but this is more confusing than explaining. What about sth like
> `Relaxed` is sufficient: the flag is a standalone boolean that is initialized as a part of env allocation during sqlallochandle init. Nothing happens before that concludes.

sfc-gh-mcepiga (approved):
nitpicks

odbc/src/api/error_trace_flag.rs
sfc-gh-rkowalski: Fixed

odbc/src/api/error_trace_flag.rs
sfc-gh-rkowalski: Fixed

PR #1090, opened 2026-05-08
NO-SNOW: Add backward-compatible converter_class and converter stubs

sfc-gh-pczajka:
## Summary
Add backward-compatible converter_class and converter properties to Connection, matching the old connector's public API surface
Create stub modules converter.py (SnowflakeConverter) and converter_null.py (SnowflakeNoConverterToPython) preserving the documented import paths (from snowflake.connector.converter import ..., from snowflake.connector.converter_null import ...)
Both properties are deprecated no-ops — the universal driver performs conversion in its C++ Arrow layer, not through a Python converter class — and emit DeprecationWarning on access via the @backward_compatibility decorator
## Context
External code (snowflake-sqlalchemy, user scripts following official Snowflake docs) reads connection.converter_class, connection.converter, and imports SnowflakeConverter / SnowflakeNoConverterToPython. These were missing in the universal driver, causing AttributeError or ImportError. This change adds the minimum shims to unblock that code, marked as deprecated for future removal.


python/src/snowflake/connector/connection.py
```diff
@@ -198,6 +199,10 @@ def __init__(
         # Extract Python-only params before processing kwargs for Rust core
         self._numpy: bool = self._resolve_numpy_option(kwargs)
         self._arrow_number_to_decimal: bool = bool(kwargs.pop("arrow_number_to_decimal", False))
+        self._converter_class: type[SnowflakeConverter] = cast(
+            "type[SnowflakeConverter]", kwargs.pop("converter_class", SnowflakeConverter)
+        )
+        self._converter: SnowflakeConverter = self._converter_class(use_numpy=self._numpy)
```
sfc-gh-asolarski: do we have to keep `self._converter_class` as a field?
we could just have `self._converter` and for property `def converter_class(self)` just return `type(self._converter)`

python/src/snowflake/connector/converter_null.py
```diff
@@ -0,0 +1,24 @@
...
+class SnowflakeNoConverterToPython(SnowflakeConverter):
+    """No-op converter stub that mirrors the old connector's null converter.
+
+    .. deprecated::
+        This class has no effect in the universal driver and will be removed
+        in a future version.
+    """
+
+    def __init__(self, **kwargs: object) -> None:
+        super().__init__(**kwargs)
```
sfc-gh-asolarski, line 27: maybe let's add `install_backward_compatibility_getattr(__name__)` to give warning on import as well?

python/src/snowflake/connector/converter.py
```diff
@@ -0,0 +1,30 @@
...
+class SnowflakeConverter:
+    """No-op converter stub retained for backward compatibility.
+
+    .. deprecated::
+        This class has no effect in the universal driver and will be removed
+        in a future version.
+    """
+
+    def __init__(self, **kwargs: object) -> None:
+        pass
```
sfc-gh-asolarski, line 34: `install_backward_compatibility_getattr` and warning on import?

python/src/snowflake/connector/connection.py
sfc-gh-asolarski, line 1: are we sure we want to deprecate this whole "converter" concept? I guess we should discuss the use-cases

sfc-gh-asolarski (changes requested):

sfc-gh-asolarski:
I think it's worth rebasing on the top of https://github.com/snowflakedb/universal-driver/pull/1127 and these PR introduce a change in how `ConnectionConfig` is built

PR #1093, opened 2026-05-11
SNOW-2872430: Add memory budget for chunk prefetch pipeline

sfc-gh-jszczerbinski:
## Summary

- Adds `MemoryBudget` to the chunk prefetch pipeline — tracks committed bytes across concurrent download+parse tasks and applies backpressure (async wait) when the budget is exceeded
- New `CLIENT_MEMORY_LIMIT` session parameter (default 1536 MB, 0 = unlimited) controls the budget
- First-chunk deadlock escape: the first remote chunk always proceeds regardless of budget to guarantee forward progress
- Memory is released eagerly on last batch receive per chunk, signaling waiting producers via `Notify`

## Design

The memory budget uses atomic tracking (`AtomicU64`) with a `Notify`-based wakeup:
- Before spawning a download task, the producer calls `try_commit(estimate)` — if over budget, it awaits `wait_for_capacity()`
- The consumer releases bytes when it receives the last batch of a chunk (via a `release_bytes` marker in the channel message)
- `estimated_memory_bytes()` uses `uncompressed_size * 2` as a conservative heuristic

Dual backpressure: slot-based (mpsc channel capacity = `prefetch_threads`) + byte-based (memory budget).

## Test plan

- [x] Unit tests for `MemoryBudget` (unlimited, first-chunk escape, within-limit, release-frees, async wake)
- [x] `cargo check -p sf_core` passes
- [ ] Existing prefetch integration tests pass in CI
- [ ] Benchmark with large result sets to verify backpressure activates correctly

### DIFF
--- a/sf_core/src/chunks/mod.rs
+++ b/sf_core/src/chunks/mod.rs
@@ -27,7 +27,7 @@ use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
 use snafu::{OptionExt, ResultExt};
 
 pub const DEFAULT_PREFETCH_THREADS: usize = 4;
-pub const DEFAULT_MEMORY_LIMIT_MB: u64 = 1536;
+pub const DEFAULT_MEMORY_LIMIT_MB: u32 = 1536;
 
 /// Configuration for the chunk prefetch pipeline.
 #[derive(Debug, Clone)]
@@ -35,7 +35,7 @@ pub struct PrefetchConfig {
     /// Number of concurrent chunk download+parse tasks.
     pub prefetch_threads: usize,
     /// Memory budget in MB for buffered chunks. 0 means unlimited.
-    pub memory_limit_mb: u64,
+    pub memory_limit_mb: u32,
 }
 
 impl Default for PrefetchConfig {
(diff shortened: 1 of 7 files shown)


sf_core/src/chunks/mod.rs
```diff
@@ -26,18 +27,22 @@ use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
 use snafu::{OptionExt, ResultExt};
 
 pub const DEFAULT_PREFETCH_THREADS: usize = 4;
+pub const DEFAULT_MEMORY_LIMIT_MB: u64 = 1536;
```
sfc-gh-asolarski: why to define one as usize and another as unsigned int64?
are we expecting more than 4PB memory limit to be set on 32bit target? 😉  (do we even support 32bit arch?) 
unless we just don't like the concept of usize, then I would get rid of it for DEFAULT_PREFETCH_THREADS as well 😅

sf_core/src/chunks/mod.rs
```diff
@@ -139,6 +144,15 @@ impl ChunkDownloadData {
             headers: chunk_headers.clone(),
         }
     }
+
+    pub fn estimated_memory_mb(&self) -> u64 {
```
sfc-gh-asolarski: let's at least document this magic estimation

sf_core/src/chunks/mod.rs
```diff
@@ -139,6 +144,15 @@ impl ChunkDownloadData {
...
+        const MEMORY_ESTIMATE_NUMERATOR: u64 = 3;
+        const MEMORY_ESTIMATE_DENOMINATOR: u64 = 2;
+        const BYTES_PER_MB: u64 = 1024 * 1024;
+        let bytes = (self.uncompressed_size.max(0) as u64) * MEMORY_ESTIMATE_NUMERATOR
+            / MEMORY_ESTIMATE_DENOMINATOR;
+        (bytes / BYTES_PER_MB).max(1)
```
sfc-gh-asolarski: let's maybe simplify?
```suggestion
        const BYTES_PER_MB: u64 = 1024 * 1024;
        let bytes = (self.uncompressed_size.max(0) as u64) * 3 / 2; // 1.5x Arrow overhead heuristic
        (bytes / BYTES_PER_MB).max(1)
```

sf_core/src/chunks/mod.rs
sfc-gh-asolarski: and anyway limit in memory budget is u32, so we have to do nasty downcasting there
`        let limit_mb = u32::try_from(limit_mb).unwrap_or(u32::MAX);`

sfc-gh-asolarski (approved):
nitpicks

sfc-gh-turbaszek (dismissed):
Please rebase and rerun precommit locally. This will update Python configuration options

PR #1108, opened 2026-05-11
perf(odbc): cache TIMESTAMP_TZ_OUTPUT_FORMAT, refresh only on ALTER SESSION

sfc-gh-makowalski:
Follow-up to #1068. Mitigates the +10.4% regression on `select_timestamp_tz_1M_arrow_recorded_http` that #1068 introduced for the new per-execute `TIMESTAMP_TZ_OUTPUT_FORMAT` lookup.

## Summary
- Snowflake session parameters can only be mutated via SQL on the same connection, so we don't have to re-read on every execute.
- Track whether the cache has been initialised (`tz_format_loaded` on `NumericSettings`).
- On `SQLExecDirect`, parse the SQL prefix; refresh only if it starts with `ALTER SESSION` (case-insensitive, after stripping leading whitespace and `--` / `/* ... */` comments).
- On prepared `SQLExecute`, the SQL text isn't kept on the client &mdash; pass `None` and skip the refresh after the initial load. A prepared `ALTER SESSION` (very rare in practice) is documented in-line as the edge case where the new format won't take effect until the next `SQLExecDirect`.
- The initial fetch still happens once per connection on the first execute, so correctness is preserved from the first TZ &rarr; CHAR fetch onward.

Net effect: **zero per-execute RPCs in steady state**, full correctness preserved for `ALTER SESSION SET/UNSET TIMESTAMP_TZ_OUTPUT_FORMAT`. Should restore the pre-#1068 perf baseline on `select_timestamp_tz_1M_arrow_recorded_http`.

## Tests
- 6 new unit tests in `tz_format_needs_refresh_tests` (positive `ALTER SESSION` matches, comment/whitespace stripping, substring guard so e.g. `SELECT 'ALTER SESSION ...'` doesn't trigger, prepared-execute (`None`) skip path, first-load always-refresh).
- All 1221 existing `odbc` lib tests still pass.
- `cargo clippy -p odbc --tests --all-features -- -D warnings`: clean.

## BD impact
None. The user-visible behaviour described in **BD#52** ("`TIMESTAMP_TZ` &rarr; CHAR/WCHAR fetch honors session `TIMESTAMP_TZ_OUTPUT_FORMAT`") is unchanged &mdash; this PR only changes how often we ask the server for the current value.

## Test plan
- [ ] CI green on `perf/cache-tz-output-format`.
- [ ] Performance check confirms the `select_timestamp_tz_1M_arrow_recorded_http` regression is recovered (compared to #1068 head). If still regressed, revisit (e.g. batch the remaining three numeric-settings RPCs into `ConnectionGetAllParameters`).

### DIFF
--- a/odbc/src/api/statement.rs
+++ b/odbc/src/api/statement.rs
@@ -139,16 +139,30 @@ fn exec_direct_impl(statement_handle: sql::Handle, statement_text: &str) -> Odbc
     tracing::info!("exec_direct: response={:?}", response);
     let response = response?;
 
-    update_numeric_settings(&conn_handle, &mut conn.numeric_settings)?;
+    update_numeric_settings(
+        &conn_handle,
+        &mut conn.numeric_settings,
+        Some(statement_text),
+    )?;
     apply_execute_response(&mut inner, stmt_handle, response, ExecutionOrigin::Direct)?;
     Ok(())
 }
 
 use crate::conversion::NumericSettings;
 
+/// Refresh the cached connection-level numeric settings after an execute.
+///
+/// `last_sql` is the SQL text that was just executed, when available
+/// (it always is for `SQLExecDirect`; it's `None` for prepared
+/// `SQLExecute` since the prepared statement's text isn't kept on the
+/// client). It is consulted only by [`tz_format_needs_refresh`] to
+/// decide whether to issue the `TIMESTAMP_TZ_OUTPUT_FORMAT` RPC for
+/// this call -- see that function and PR #1068 follow-up for the
+/// rationale.
 fn update_numeric_settings(
     conn_handle: &ConnectionHandle,
     settings: &mut NumericSettings,
+    last_sql: Option<&str>,
 ) -> OdbcResult<()> {
     let g = global().context(OdbcRuntimeSnafu)?;
     g.block_on(async |c| {
@@ -191,16 +205,24 @@ fn update_numeric_settings(
             tracing::info!("Server parameter VARCHAR_AND_BINARY_MAX_SIZE_IN_RESULT = {size}");
         }
 
-        // TIMESTAMP_TZ_OUTPUT_FORMAT: read on every execute so an in-flight
-        // `ALTER SESSION SET TIMESTAMP_TZ_OUTPUT_FORMAT = ...` takes effect
-        // for the next statement. Empty / unset / no-TZ-token formats keep
-        // the legacy UTC-only fetch behaviour (see
-        // `crate::conversion::timestamp::parse_tz_offset_format`).
+        // TIMESTAMP_TZ_OUTPUT_FORMAT: lazy + invalidate-on-DDL.
         //
-        // Update semantics differ from the other settings in this
-        // function: those have meaningful server-side defaults so
-        // resetting to default on a transient RPC failure is harmless.
-        // `tz_offset_format` does NOT -- the customer set it
+        // Snowflake session parameters can only be mutated by SQL run on
+        // the same connection (`ALTER SESSION SET/UNSET ...`), so we
+        // don't have to re-read the server cache on every execute. We
+        // refresh only when:
+        //   - the cache has never been populated (first execute), OR
+        //   - the SQL we just ran could have mutated it (`ALTER SESSION`
+        //     prefix, see `tz_format_needs_refresh`).
+        //
+        // Skipping this RPC removes the per-execute regression measured
+        // in `select_timestamp_tz_1M_arrow_recorded_http` (#1068) for
+        // every workload that doesn't issue ALTER SESSION mid-stream.
+        //
+        // Update semantics for an actual refresh differ from the other
+        // settings in this function: those have meaningful server-side
+        // defaults so resetting to default on a transient RPC failure is
+        // harmless. `tz_offset_format` does NOT -- the customer set it
         // deliberately via `ALTER SESSION` and a transient blip silently
         // flipping the next fetch from `+HH:MM` rendering back to bare
         // UTC is a wire-format regression with no diagnostic the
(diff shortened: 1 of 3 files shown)


sfc-gh-makowalski:
* **#1108** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1108?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1108?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

odbc/src/api/statement.rs
```diff
@@ -212,19 +234,110 @@ fn update_numeric_settings(
...
+            let rpc_result = c
+                .connection_get_parameter(ConnectionGetParameterRequest {
+                    conn_handle: Some(*conn_handle),
+                    key: "TIMESTAMP_TZ_OUTPUT_FORMAT".to_string(),
+                })
+                .await
+                .map(|resp| resp.value)
+                .map_err(|e| format!("{e:?}"));
+            apply_tz_offset_format_update(&mut settings.tz_offset_format, rpc_result);
+            settings.tz_format_loaded = true;
```
sfc-gh-rkowalski, line 247: **Correctness regression on transient RPC failure.**

`apply_tz_offset_format_update` deliberately leaves `tz_offset_format` untouched on `Err(_)` (per its own doc and the comment immediately above), but `tz_format_loaded` is set unconditionally here. A transient failure on the *first* execute leaves the connection in `loaded=true, tz_offset_format=None` until an `ALTER SESSION` happens to come along — which is exactly the wire-format regression the surrounding comment ("a transient blip silently flipping the next fetch from `+HH:MM` rendering back to bare UTC") was written to prevent.

Fix options:
- Have `apply_tz_offset_format_update` return whether it actually applied, and gate `tz_format_loaded = true` on that.
- Or inline the `match rpc_result` here and only flip `loaded` in the `Ok(_)` arms.

Please also add a regression test: first-execute returning `Err(_)` must leave `loaded == false` so the next non-ALTER execute retries. The existing `apply_tz_offset_format_update_tests` cover the cache itself but not this interaction.

odbc/src/api/statement.rs
```diff
@@ -212,19 +234,110 @@ fn update_numeric_settings(
...
+/// Pure function so the truth table can be unit-tested without an RPC
+/// mock. Returns `true` when:
+/// - the cache has never been populated (`!loaded`), OR
+/// - the SQL we just ran starts with `ALTER SESSION` (case-insensitive,
+///   leading whitespace and SQL comments stripped).
+///
+/// `last_sql == None` is the prepared `SQLExecute` path -- a prepared
+/// statement that holds an `ALTER SESSION` is pathological enough that
+/// we trade exactness for the perf win.
+pub(crate) fn tz_format_needs_refresh(loaded: bool, last_sql: Option<&str>) -> bool {
```
sfc-gh-rkowalski, line 265: Two scope gaps worth either fixing or documenting alongside the existing prepared-`ALTER SESSION` caveat:

1. **Multi-statement.** `BEGIN; ALTER SESSION SET TIMESTAMP_TZ_OUTPUT_FORMAT = '...'; SELECT ...` (Snowflake supports `MULTI_STATEMENT_COUNT > 1`) won't match the leading-prefix check.
2. **Stored procedures.** `CALL my_proc()` where `my_proc` runs `ALTER SESSION` server-side won't invalidate the cache — the client-side SQL prefix doesn't see it.

The PR description says the parser is "conservative on purpose: we'd rather over-refresh than miss" — but both of the above are genuine *misses*, not over-refresh. At minimum, mention them in this doc comment as known edge cases.

odbc/src/api/statement.rs
```diff
@@ -212,19 +234,110 @@ fn update_numeric_settings(
...
+    let mut head = trimmed.chars();
+    let alter = "ALTER";
+    let session = "SESSION";
+    if !consume_keyword(&mut head, alter) {
+        return false;
+    }
+    if !consume_at_least_one_space(&mut head) {
+        return false;
+    }
+    consume_keyword(&mut head, session) && head.next().is_none_or(|c| c.is_whitespace() || c == '(')
```
sfc-gh-rkowalski, line 293: **Comment-between-keywords is a miss.**

`strip_leading_whitespace_and_comments` handles only *leading* whitespace/comments. Once we're past `ALTER`, this predicate rejects anything that isn't whitespace or `(`, so:

- `ALTER /* x */ SESSION SET ...` — fails between `ALTER` and `SESSION` (inter-keyword space-only check).
- `ALTER SESSION/* x */ SET ...` — fails here on the post-`SESSION` byte being `/`.

Probably rare in practice, but unlike the prepared-execute case these aren't documented.

**Minor on the same line:** why `c == '('`? `ALTER SESSION` isn't followed by a paren in the grammar. If defensive, a one-line comment would help; if dead, drop it.

odbc/src/api/statement.rs
```diff
@@ -212,19 +234,110 @@ fn update_numeric_settings(
...
+    for kc in keyword.chars() {
+        match iter.next() {
+            Some(c) if c.eq_ignore_ascii_case(&kc) => continue,
+            _ => return false,
+        }
+    }
+    true
+}
+
+fn consume_at_least_one_space(iter: &mut std::str::Chars<'_>) -> bool {
```
sfc-gh-rkowalski, line 327: Minor / style: cloning the iterator and advancing the original in lockstep works but reads as slightly clever. A more idiomatic equivalent:

```rust
fn consume_at_least_one_space(iter: &mut std::str::Chars<'_>) -> bool {
    let mut consumed = false;
    while iter.clone().next().is_some_and(|c| c.is_whitespace()) {
        iter.next();
        consumed = true;
    }
    consumed
}
```

Not a blocker.

odbc/src/api/statement.rs
```diff
@@ -329,6 +442,98 @@ mod apply_tz_offset_format_update_tests {
...
+            "ALTER SESSION SET TIMESTAMP_TZ_OUTPUT_FORMAT = 'YYYY-MM-DD HH24:MI:SS TZH:TZM'",
+            "alter session set timestamp_tz_output_format = 'YYYY-MM-DD'",
+            "ALTER SESSION UNSET TIMESTAMP_TZ_OUTPUT_FORMAT",
+            "AlTeR  SeSsIoN  SET FOO = 1",
+            "  \t\nALTER SESSION SET FOO = 1",
+            "/* warm-up */ ALTER SESSION SET FOO = 1",
+            "-- toggle\nALTER SESSION SET FOO = 1",
+            "/* a */ -- b\n /* c */ ALTER SESSION SET FOO = 1",
+            "ALTER\tSESSION\tSET FOO = 1",
+        ] {
```
sfc-gh-rkowalski, line 498: Consider adding `"ALTER SESSION SET FOO = 1;"` (trailing semicolon, no whitespace before it) to this array. It passes today because the post-`SESSION` byte is whitespace, but the post-keyword predicate is non-obvious enough that it's worth pinning.

sfc-gh-rkowalski (changes requested):

sfc-gh-makowalski:
Migrating to the new repo. Reopened as snowflake-eng/universal-driver#74 (rebased onto the new `main`, same change + unit tests, with the outstanding review feedback from @sfc-gh-rkowalski carried over as a checklist). Closing here.

PR #1113, opened 2026-05-12
test(python): reduce per-test overhead in integ/e2e suites

sfc-gh-turbaszek:
## Summary

Reduces wall-clock time of the Python `hatch run dev.py3.13:integ` / `:e2e` suites by trimming per-test fixed costs. No production code changes; tests only.

- **Module-scoped `connection` fixture.** Promote the connection fixture from function scope to module scope. Most test modules only query through the connection and never mutate it; sharing the session across a module avoids repeated JWT login/logout round-trips. A new function-scoped `function_connection` fixture is provided for tests that mutate state (`close`, `autocommit`, `commit/rollback`, `set_autocommit`, `ALTER SESSION`). Three modules override `connection` to `function_connection` where needed: `test_connection.py`, `test_session_parameters.py`, `test_cursor_format_properties.py`. `with_paramstyle` continues to parametrize correctly — pytest creates a module-scoped instance per paramstyle.
- **Session-scoped Wiremock JVM.** Add a `wiremock` fixture that lives for the full session and is reset between tests via a new `WiremockClient.reset()` that hits the `/__admin/reset` admin endpoint. Replaces ~28 per-test `WiremockClient().start()` calls across `e2e/session/test_close.py`, `e2e/session/test_logout.py`, `integ/session/test_logout.py`, `integ/telemetry/test_connection_telemetry.py`, `integ/telemetry/test_client_identity.py`, and `integ/put_get/test_put_get_source_compression.py`. Saves ~1–3 s of JVM startup per invocation.
- **Disable `log_cli`.** `log_cli=true` streams INFO to stdout under `-n auto` — causes worker stdout lock contention and bloats CI logs with no upside. Flip to `false`; people who want live logs can opt in with `--log-cli-level=DEBUG`.
- **xdist worker count.** Drop the 8-worker floor in `pytest_xdist_auto_num_workers`; use `os.cpu_count() or 1`. Matches actual CPU count on small machines and avoids interpreter over-subscription.

## Test plan

- [x] `hatch run dev.py3.13:integ` against local reg: 433 passed / 1 failed / 7 skipped in 23.2s. The 1 failure (`TestCursorDescription::test_description_empty_subquery_type_code[TIMESTAMP_NTZ]`) is a pre-existing server-side type-code mismatch unrelated to fixture scoping.
- [x] `hatch run dev.py3.13:e2e` against local reg: 743 passed / 4 failed / 10 skipped / 44 errors in 76.2s. The 44 errors are the optional pandas/numpy deps not installed in the default `dev` env. The 4 failures are pre-existing server behaviors on local reg (TZ mapping, missing sample data, vector overflow) — none traceable to fixture changes.
- [ ] CI matrix (full OS × Python × cloud-provider sweep) to confirm the module-scoped fixture works against remote Snowflake too, and that the Wiremock reset semantics hold on every platform.

Not included (will follow in later PRs): de-parametrizing large-result-set tests per numeric-type synonym, collapsing serial-INSERT binding loops, reducing `SYSTEM$WAIT(30)` to 3s.

### DIFF
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -16,12 +16,11 @@ from snowflake.connector.cursor import DictCursor
 from .compatibility import IS_UNIVERSAL_DRIVER
 from .connector_factory import ConnectorFactory, create_connection_with_adapter
 from .private_key_helper import get_test_private_key_path
+from .wiremock_client import WiremockClient
 
 
 # Type alias for a single row returned from cursor
 Row = tuple[Any, ...]
-
-_MIN_XDIST_WORKERS = 8
 
 
 def pytest_configure(config):
@@ -41,8 +40,7 @@ def pytest_configure(config):
 
 
 def pytest_xdist_auto_num_workers(config):
-    num_cpus = os.cpu_count() or _MIN_XDIST_WORKERS
-    return max(num_cpus, _MIN_XDIST_WORKERS)
+    return os.cpu_count() or 1
 
 
 @pytest.mark.optionalhook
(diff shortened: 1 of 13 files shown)


python/tests/e2e/pandas/conftest.py
```diff
@@ -0,0 +1,27 @@
...
+    failing ``is_float_dtype`` assertions. Reset to the default (False) after
+    each test so the shared connection stays untainted.
+    """
+    yield
+    try:
+        connection.arrow_number_to_decimal = False
+    except Exception:
+        pass
+    try:
+        connection.arrow_number_to_decimal_setter = False
```
sfc-gh-pczajka, line 25: Do we want to override the setter too? Is there something I don't know :D?

sfc-gh-turbaszek: We only reset the value to default. We don't override implementation

python/tests/conftest.py
```diff
@@ -41,8 +40,7 @@ def pytest_configure(config):
 
 
 def pytest_xdist_auto_num_workers(config):
-    num_cpus = os.cpu_count() or _MIN_XDIST_WORKERS
```
sfc-gh-pczajka, line 44: Let's remove the constant as well

sfc-gh-pczajka (changes requested):

sfc-gh-pczajka (approved):

sfc-gh-pbulawa (approved):

PR #1115, opened 2026-05-12
NO-SNOW: remove deprecated "region" property

sfc-gh-pczajka:
## Summary

- Remove deprecatd `region` property
- Add BCR entry
- Remove unused `self.kwargs` from `Connection.__init__`

## 

### DIFF
--- a/python/src/snowflake/connector/connection.py
+++ b/python/src/snowflake/connector/connection.py
@@ -211,9 +211,6 @@ class Connection(ErrorHandlerMixin):
         self._session_parameters = SessionParametersProxy(self.db_api, self.conn_handle)
         self._connection_info = ConnectionInfoProxy(self.db_api, self.conn_handle)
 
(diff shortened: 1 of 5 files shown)


python/src/snowflake/connector/connection.py
```diff
@@ -647,9 +644,10 @@ def port(self) -> int | None:
         return cast("int | None", self._connection_info["port"])
 
     @property
+    @backward_compatibility
     def region(self) -> str | None:
         """Deprecated. The region for the Snowflake account."""
-        raise NotImplementedError("region is not implemented")
+        return self.config.region
```
sfc-gh-turbaszek: Is this used by anyone? It was deprecated already so next major is a good opportunity to remove it altogether

sfc-gh-turbaszek (changes requested):
What is the reasoning behind keeping already deprecated param?

sfc-gh-turbaszek (approved):

PR #1114, opened 2026-05-12
NO-SNOW: Clippy fix using multiple_of

sfc-gh-ddas:
### TL;DR

Replaced manual modulo check with `is_multiple_of` for even-length validation in hex decoding.

### What changed?

The even-length check in `hex_decode_ascii` was updated to use the `is_multiple_of(2)` method instead of the explicit `% 2 != 0` modulo expression.

### How to test?

Pass odd-length and even-length hex strings to the binary conversion logic and verify that odd-length inputs still return a `SQLSTATE 22018` error, while valid even-length hex strings decode correctly.

### Why make this change?

`is_multiple_of` is more expressive and idiomatic, making the intent of the check immediately clear without requiring the reader to mentally parse the modulo arithmetic.

PR #1112, opened 2026-05-12
NO-SNOW: Return pre-decryption size in ODBC resultset

sfc-gh-ddas:
### TL;DR

Fixes BD#4: GET rowset `size` column now reports the on-cloud (pre-decryption) byte count for ODBC, matching legacy `libsnowflakeclient` behavior.

### What changed?

Each cloud download function (S3, GCS, Azure) now returns a `DownloadResponse` struct instead of a bare tuple. This struct carries the raw bytes, digest, encryption metadata, and a new `cloud_byte_count` field representing the wire byte count of the blob before decryption.

A `PutGetResultsetFlavor` is threaded from `to_file_download_data` through `DownloadData` and `SingleDownloadData` into `download_single_file`. A new `download_result_size` function uses this flavor to select which byte count to report in the GET result's `size` column: ODBC reports `cloud_byte_count` (pre-decryption, matching legacy `srcFileSize`), while Python continues to report the post-decryption buffer length.

The ODBC end-to-end test for GET rowset metadata is updated to assert a single expected size of `32` (the on-cloud byte count), removing the old `OLD_DRIVER_ONLY`/`NEW_DRIVER_ONLY` split. BD#4 in `BehaviorDifferences.yaml` is marked `fixed`.

### How to test?

- Run the ODBC end-to-end test `should return correct rowset for GET` and confirm it passes with the size asserting `32`.
- Run the new unit tests in `file_manager/mod.rs`:
  - `download_result_size_odbc_uses_cloud_byte_count`
  - `download_result_size_python_uses_output_length`
  - `download_result_size_sse_branches_collapse_to_same_value`
- Run the new `query_response` tests `download_data_forwards_flavor_python` and `download_data_forwards_flavor_odbc` to confirm flavor propagation.
- Run the Azure integration test `azure_download_success_returns_data_and_metadata` to confirm `cloud_byte_count` equals the body length.

### Why make this change?

The legacy ODBC driver (`libsnowflakeclient`) reported the pre-decryption (on-cloud) file size in the GET result's `size` column. The new driver was incorrectly reporting the post-decryption size, causing a mismatch. This fix restores parity with the legacy behavior for ODBC while leaving the Python driver's behavior unchanged.

PR #1097, opened 2026-05-11
NO-SNOW: Skip message in ODBC when file not put

sfc-gh-ddas:
### TL;DR

Populate the `message` column in PUT result sets with the legacy ODBC skipped-file literal when `overwrite=false` and the target file already exists.

### What changed?

A new `PutGetResultsetFlavor` type is now threaded through the file upload pipeline — from `WrapperPresets` through `to_file_upload_data`, into `UploadData` and `SingleUploadData`, and finally consumed in `upload_single_file`. A new helper `upload_result_message` uses the flavor to decide what to write into the `message` column of the upload result:

- **ODBC flavor + Skipped status** → `"File with same name already exists. SKIPPED"` (matching the legacy `libsnowflakeclient` `MESSAGE_SKIPPED` macro)
- **All other (flavor, status) combinations** → empty string

The constant `ODBC_PUT_MESSAGE_SKIPPED` is defined in both the Rust core and the ODBC test headers so both sides of the contract are pinned to the same literal.

E2E tests now assert that a successful upload produces an empty `message` column, and that a skipped upload (OVERWRITE=FALSE on an existing file) produces the expected ODBC message string.

### How to test?

- Run the existing unit tests in `sf_core::file_manager` — the new `upload_result_message` tests cover all (flavor, status) combinations and pin the exact string value.
- Run the ODBC E2E tests under `put_get`:
  - `should return correct rowset for PUT` — verifies the `message` column is empty on a successful upload.
  - `should not overwrite file when OVERWRITE is set to false` — verifies the `message` column equals `PUT_ROW_MESSAGE_SKIPPED` when a file is skipped.

### Why make this change?

Legacy ODBC applications that parse the PUT result set's `message` column rely on the exact string `"File with same name already exists. SKIPPED"` to detect skipped uploads. The universal driver was previously returning an empty string for all outcomes, breaking that contract. This change restores the expected ODBC behavior while keeping the Python flavor's message column empty, matching its historical behavior.

PR #1092, opened 2026-05-10
NO-SNOW: Encryption text column in ODBC PUT/GET result

sfc-gh-ddas:
### TL;DR

Adds an `encryption` column to PUT/GET result sets for the ODBC flavor, restoring parity with the legacy `libsnowflakeclient` behavior.

### What changed?

- The ODBC PUT result set now includes an `encryption` column (index 8) between `status` and `message`, always populated with `"ENCRYPTED"`.
- The ODBC GET result set now includes an `encryption` column (index 4) between `status` and `message`, always populated with `"DECRYPTED"`.
- `upload_row_types` and `download_row_types` conditionally insert the `encryption` column when `put_get_resultset_flavor == PutGetResultsetFlavor::Odbc`.
- Two constants, `ODBC_PUT_ENCRYPTION_LITERAL` and `ODBC_GET_ENCRYPTION_LITERAL`, are introduced as the single source of truth for these string values.
- ODBC tests now assert the total column count (`PUT_ROW_NUM_COLS = 9`, `GET_ROW_NUM_COLS = 5`) and read the encryption value directly from the new dedicated column index rather than branching on old/new driver behavior.
- The behavior difference entry `BD#3` ("Encryption field is no longer included in the result") is marked as `fixed`.
- Unit tests are split into Python and ODBC variants to cover both column layouts.

### How to test?

Run the existing ODBC end-to-end PUT/GET tests:

```
# PUT test
TEST_CASE("should return correct rowset for PUT", "[put_get]")

# GET test
TEST_CASE("should return correct rowset for GET", "[put_get]")
```

Run the Rust unit tests:

```
cargo test upload_column_metadata_has_correct_structure_odbc
cargo test download_column_metadata_has_correct_structure_odbc
cargo test upload_column_metadata_has_correct_structure_python
cargo test download_column_metadata_has_correct_structure_python
```

### Why make this change?

The legacy `libsnowflakeclient` ODBC driver included an `encryption` column in PUT/GET result sets, reporting `"ENCRYPTED"` and `"DECRYPTED"` respectively. This column was missing from the new driver, causing a behavioral regression for ODBC users who depend on it. This change restores that column exclusively for the ODBC result set flavor, keeping non-ODBC wrappers (Python, etc.) unaffected.

PR #1120, opened 2026-05-12
NO-SNOW: Silently ignore unknown compression type in ODBC

sfc-gh-ddas:
### TL;DR

Restores legacy libsnowflakeclient behavior for the ODBC flavor: files with unsupported compression types are silently treated as uncompressed during `AUTO_DETECT` instead of returning an error (BD#6).

### What changed?

The `get_source_compression` function now accepts the active `PutGetResultsetFlavor`. A new `auto_detect_source_compression` function gates the fallback behavior: when the `Odbc` flavor is active and `AUTO_DETECT` encounters an `UnsupportedCompressionType` error (whether triggered by filename extension or magic-byte detection), it falls back to `CompressionType::None` and continues the upload. The `Python` flavor retains the existing behavior of surfacing the error. The fallback applies uniformly to all unsupported formats (`.xz`, `.lzma`, `.lz`, `.lzo`, `.Z`, `.parquet`, `.orc`, etc.).

The ODBC end-to-end test was updated to reflect the corrected behavior: it now verifies that the upload succeeds, the file is stored as gzip-compressed (since `AUTO_COMPRESS=TRUE` then runs over the "uncompressed" source), and the downloaded payload decompresses byte-for-byte to the original file content. The shared feature file splits the scenario into a `@core_int @python_int` scenario (error is thrown) and a separate `@odbc_e2e` scenario (silent fallback with roundtrip verification). BD#6 in `BehaviorDifferences.yaml` is marked `fixed`.

### How to test?

- Run the ODBC end-to-end test `should silently upload file with unsupported compression type as uncompressed` — it uploads an LZMA-compressed file with `SOURCE_COMPRESSION=AUTO_DETECT`, verifies the PUT result shows `NONE` source compression and `GZIP` target compression, then GETs the file and confirms the decompressed bytes match the original.
- Run the Python/core integration test `should return error for unsupported compression type` to confirm the error-surfacing path is unchanged.
- Run the new unit tests in `sf_core`: `auto_detect_source_compression_odbc_falls_back_to_none_for_unsupported`, `auto_detect_source_compression_python_surfaces_unsupported_error`, and the buffer-detection and explicit-param variants.

### Why make this change?

Legacy `libsnowflakeclient` silently ignored unsupported compression types during `AUTO_DETECT` and proceeded with the upload treating the file as uncompressed. The new driver was returning `SQL_ERROR` instead, breaking existing ODBC workflows that relied on this silent fallback. This change restores parity with the legacy behavior specifically for the ODBC flavor while keeping the stricter error-surfacing behavior intact for the Python connector.

### DIFF
--- a/odbc_tests/BehaviorDifferences.yaml
+++ b/odbc_tests/BehaviorDifferences.yaml
@@ -31,7 +31,7 @@ behavior_differences:
 
   6:
     name: "Unsupported compression type returns error instead of being silently ignored"
-    status: unknown
+    status: fixed
     type: bug
     reviewed: false
 
--- a/odbc_tests/tests/e2e/put_get/put_get_source_compression.cpp
+++ b/odbc_tests/tests/e2e/put_get/put_get_source_compression.cpp
@@ -173,21 +173,33 @@ TEST_CASE("should compress uncompressed file when SOURCE_COMPRESSION set to NONE
   CHECK(get_data<SQL_C_CHAR>(stmt, PUT_ROW_STATUS_IDX) == std::string("UPLOADED"));
 }
 
-TEST_CASE("should return error for unsupported compression type", "[put_get]") {
-  // Given Snowflake client is logged in
+TEST_CASE("should silently upload file with unsupported compression type as uncompressed", "[put_get]") {
   Connection conn;
+  // Given Snowflake client is logged in
   const std::string stage = create_stage(conn, unique_stage_name("ODBCTST_SC_UNSUPPORTED"));
 
   // And File compressed with unsupported format
   auto [filename, file] = test_file("LZMA");
 
   // When File is uploaded with SOURCE_COMPRESSION set to AUTO_DETECT
-  std::string put_sql = "PUT 'file://" + as_file_uri(file) + "' @" + stage + " SOURCE_COMPRESSION=AUTO_DETECT";
+  auto put_stmt =
+      conn.execute_fetch("PUT 'file://" + as_file_uri(file) + "' @" + stage + " SOURCE_COMPRESSION=AUTO_DETECT");
 
-  auto stmt = conn.createStatement();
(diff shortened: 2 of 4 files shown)


sfc-gh-ddas:
* **#1201** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1201?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1139** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1139?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1120** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1120?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1120?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1112** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1112?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1097** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1097?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1092** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1092?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1114** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1114?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sfc-gh-turbaszek (approved):
The change looks ok. Any thought about deprecating tis behavior instead of silently failing to auto detection?  Or should this be default behavior for all drivers?

sfc-gh-ddas:
This behaviour is not ideal but fixing it would be a big change for ODBC users, we should stick to the legacy drivers' behaviour for the inital realeases. The driver presets allow us to preserve the descrepency between drivers. Long term, we should throw an error like Python does.

PR #1123, opened 2026-05-12
SNOW-2872349: ODBC logout best-effort strategy and WireMock e2e tests

sfc-gh-jszczerbinski:
## Summary

- Set `logout_error_strategy=best_effort` for ODBC wrapper to match legacy driver behavior (destructor `catch(...)` silently swallows all logout errors)
- Add `SSL` parameter support in sf_core (maps `ssl=on/off` to protocol selection)
- Add WireMock-based ODBC e2e tests verifying disconnect idempotency, thread safety, and best-effort error handling
- Tag shared logout Gherkin scenarios with `@odbc_e2e`
- Extract platform abstraction layer (`Subprocess`, `platform.hpp`) from WiremockClient to eliminate all `#ifdef _WIN32` blocks

## Context

Legacy ODBC calls `disconnect()` from `~Connection()` inside a `catch(...)` block, so logout failures never reach the application. The UD default is `Strict` (surfaces errors). This PR aligns UD ODBC with legacy by setting `best_effort` — errors are logged as WARN but `SQLDisconnect` always returns `SQL_SUCCESS`.

## Changes

- **`odbc/src/api/connection.rs`**: Inject `logout_error_strategy=best_effort` into connection options
- **`sf_core/src/config/`**: Add `SSL` param (bool, maps to protocol), add `on`/`off` to `get_bool`, validate ssl+protocol conflict
- **`odbc_tests/common/include/WiremockClient.hpp`**: Rewritten to use platform layer — zero ifdefs, clean linear logic
- **`odbc_tests/common/include/Subprocess.hpp`** + platform `.cpp` files: RAII process wrapper (pimpl pattern)
- **`odbc_tests/common/include/platform.hpp`**: Portable `find_free_port()`, `exec_command()`, `null_device()`
- **`odbc_tests/tests/e2e/session/logout.cpp`**: Three WireMock tests (idempotent, concurrent, server-error-during-logout)

## Test plan

- [x] WireMock e2e tests pass on Linux CI (idempotency + concurrent disconnect + best-effort 500)
- [x] Windows tests skip gracefully (`SKIP("WireMock tests not yet validated on Windows")`)
- [x] Existing e2e/integ tests still pass (no behavioral change for successful logout)
- [x] sf_core unit tests pass (config, validation)
- [ ] Windows CI build passes with new Subprocess/platform abstraction

### DIFF
--- a/odbc/src/api/connection.rs
+++ b/odbc/src/api/connection.rs
@@ -49,11 +49,6 @@ fn normalize_crl_enabled_value(value: &str) -> String {
     }
 }
 
-fn is_truthy(value: &str) -> bool {
-    let v = value.trim();
-    v.eq_ignore_ascii_case("true") || v.eq_ignore_ascii_case("on") || v == "1"
-}
-
 fn normalize_connection_string_options(
     connection_string_map: HashMap<String, String>,
 ) -> HashMap<String, ConfigSetting> {
@@ -89,11 +84,7 @@ fn normalize_connection_string_option(
         "PRIV_KEY_FILE_PWD" | "PRIV_KEY_PWD" => {
             Some(("private_key_password".to_owned(), value.into()))
         }
-        "SSL" => {
-            let protocol = if is_truthy(&value) { "https" } else { "http" };
-            Some(("protocol".to_owned(), protocol.to_owned().into()))
-        }
-        // Forward other keys (e.g. SERVER, UID) for `sf_core` alias resolution; do not
+        // Forward other keys (e.g. SERVER, UID, SSL) for `sf_core` alias resolution; do not
         // pre-canonicalize here to avoid duplicate seed keys.
         _ => Some((upper, value.into())),
     }
--- a/odbc/src/api/connection.rs
+++ b/odbc/src/api/connection.rs
@@ -240,7 +240,7 @@ fn connect_with_params(
 
     // Legacy ODBC silently swallows all logout errors (destructor catch-all).
     options
-        .entry("logout_error_strategy".to_owned())
+        .entry("LOGOUT_ERROR_STRATEGY".to_owned())
         .or_insert_with(|| "best_effort".to_owned().into());
 
     let dbc = conn_from_handle(connection_handle)?;
--- a/odbc/src/api/connection.rs
+++ b/odbc/src/api/connection.rs
@@ -49,6 +49,11 @@ fn normalize_crl_enabled_value(value: &str) -> String {
     }
 }
 
+fn is_truthy(value: &str) -> bool {
+    let v = value.trim();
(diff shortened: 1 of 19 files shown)


odbc/src/api/connection.rs
```diff
@@ -238,6 +238,11 @@ fn connect_with_params(
         options.insert("port".to_owned(), port_int.into());
     }
 
+    // Legacy ODBC silently swallows all logout errors (destructor catch-all).
+    options
+        .entry("logout_error_strategy".to_owned())
```
sfc-gh-pbulawa: `parse_connection_string` upper-cases every key and `normalize_connection_string_option` keeps unknown keys upper-case, so a user-supplied `LOGOUT_ERROR_STRATEGY=strict` lands in `options` under the key `"LOGOUT_ERROR_STRATEGY"`. The new `entry("logout_error_strategy")` (lower-case) does a string-equal lookup, finds nothing, and inserts `"best_effort"`. Both keys then coexist in the map.

Downstream, `sf_core::apis::database_driver_v1::validation::resolve_options` resolves both keys case-insensitively to the same canonical `logout_error_strategy` and emits a `ConflictingParameters` `Error`

odbc_tests/tests/e2e/session/logout.cpp
```diff
@@ -0,0 +1,127 @@
...
+  int expected_error_count = 0;
+  for (auto r : results) {
+    if (r == SQL_SUCCESS) {
+      success_count++;
+    } else if (r == SQL_ERROR) {
+      expected_error_count++;
+    }
+  }
+  CHECK(success_count + expected_error_count == num_threads);
+  CHECK(success_count == 1);
```
sfc-gh-pbulawa, line 102: Let's assert that there are 4 diag records with state `08003`

sf_core/src/apis/database_driver_v1/validation.rs
```diff
@@ -58,8 +58,8 @@ fn try_coerce_to_value_type(setting: &Setting, expected: ValueType) -> Option<Se
         ValueType::Int => s.parse::<i64>().ok().map(Setting::Int),
         ValueType::Double => s.parse::<f64>().ok().map(Setting::Double),
         ValueType::Bool => match s.to_lowercase().as_str() {
-            "true" => Some(Setting::Bool(true)),
-            "false" => Some(Setting::Bool(false)),
+            "true" | "1" | "on" => Some(Setting::Bool(true)),
```
sfc-gh-pbulawa, line 61: Should we document it somewhere that we accept the "on"/"off" now as a boolean value?

sfc-gh-pbulawa (changes requested):

sfc-gh-pbulawa (approved):

PR #1129, opened 2026-05-12
SNOW-2881745: external browser auth

sfc-gh-asolarski:


sfc-gh-asolarski:
* **#1132** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1132?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1133** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1133?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1129** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1129?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1129?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sf_core/src/config/connection_config.rs
```diff
@@ -374,14 +385,9 @@ fn build_auth_config(settings: &ParamStore) -> Result<AuthConfig, ConfigError> {
                     value: authenticator.clone(),
                     explanation: "The authenticator URL is not a valid URL".to_string(),
                 }
-                .build()
+                    .build()
```
sfc-gh-turbaszek: unnecessary changes?

sf_core/src/config/connection_config.rs
```diff
@@ -396,13 +402,19 @@ fn build_auth_config(settings: &ParamStore) -> Result<AuthConfig, ConfigError> {
                 disable_saml_url_check: settings
                     .get_bool(DISABLE_SAML_URL_CHECK)
                     .unwrap_or(false),
-                authentication_timeout_secs,
+                authentication_timeout_secs: parse_authentication_timeout(settings),
             }))
         }
+        "EXTERNALBROWSER" => Ok(AuthConfig::ExternalBrowser {
```
sfc-gh-turbaszek, line 407: Is the pattern of string literals best? We have `AuthConfig` enum, maybe it's worth adding the names there? Especially that that would help us build later message automatically + we could expose auth methods to wrappers to build nativ enums. WDYT? Potentially a followup change

sf_core/src/config/rest_parameters.rs
```diff
@@ -451,10 +455,21 @@ impl LoginMethod {
...
+                    username: Self::non_empty_string(settings, "user")
+                        .context(MissingParameterSnafu { parameter: "user" })?,
+                    authentication_timeout_secs,
+                })
+            }
             _ => InvalidParameterValueSnafu {
                 parameter: "authenticator",
                 value: authenticator,
-                explanation: "Allowed values are snowflake, snowflake_jwt, snowflake_password, programmatic_access_token, username_password_mfa, or an https:// URL for native Okta SSO (case-insensitive)",
+                explanation: "Allowed values are snowflake, snowflake_jwt, snowflake_password, programmatic_access_token, username_password_mfa, externalbrowser, or an https:// URL for native Okta SSO (case-insensitive)",
```
sfc-gh-turbaszek: Repeated message

sf_core/src/rest/snowflake/external_browser.rs
```diff
@@ -0,0 +1,774 @@
...
+///
+/// 1. Bind a local TCP listener on an OS-assigned port.
+/// 2. Ask Snowflake for the SSO URL and proof key.
+/// 3. Open the SSO URL in the user's default browser.
+/// 4. Wait for the IdP to redirect back with a token (or time out).
+#[tracing::instrument(
+    skip(client, login_parameters, browser_opener),
+    fields(authentication_timeout_secs)
+)]
+pub(crate) async fn external_browser_authenticate(
```
sfc-gh-turbaszek, line 135: What end users of python will see? Today there is a prompt printed that includes link in case the process can't open browser or you incidentally closed the window.

sfc-gh-turbaszek (approved):
The changes looks reasonable. I think testing e2e with a wrapper is necessary to make sure we have same behavior as in the past. But I assume this can be done in followup PRs

sf_core/src/config/connection_config.rs
sfc-gh-asolarski: This is rustfmt reformatting the surrounding match arm when the new code was added. The indentation is consistent with `cargo fmt` — reverting would cause the formatter to re-apply it.

sf_core/src/config/connection_config.rs
sfc-gh-asolarski: Good call — agreed. The string literal pattern is fragile and having canonical names on the enum (or a shared constant) would help with auto-generated messages and wrapper enum exposure. Will do in a follow-up PR.

sf_core/src/config/rest_parameters.rs
sfc-gh-asolarski: Good catch — extracted into a shared `AUTHENTICATOR_ALLOWED_VALUES` constant in `config/mod.rs` so both `connection_config.rs` and `rest_parameters.rs` reference the single source of truth.

sf_core/src/rest/snowflake/external_browser.rs
sfc-gh-asolarski: Great catch — thank you! Added a `tracing::warn!` that prints the SSO URL so the user can manually open it if the browser doesn't launch. Uses the existing `CallbackLayer` so wrappers surface it through their logging callbacks. Two cases: success (friendly message with URL as fallback) and failure (direct "open this URL manually" with the error).

sfc-gh-asolarski: Updated: using `eprintln!` to unconditionally print the SSO URL to stderr, matching the old Python connector's `print()` behaviour. This ensures the URL is always visible to the user regardless of logging configuration — important because `tracing::warn!` only reaches users who have explicitly set up the logging callback. The `eprintln!` is documented with a comment explaining the backward-compatibility rationale.

python/src/snowflake/connector/connection_config.py
```diff
@@ -92,6 +92,12 @@ class ConnectionConfig:
     log_max_query_length: int | None = 80
     """Maximum number of characters of a query string to include in log messages. Default: 80"""
 
+    log_query_parameters: bool | str | None = False
```
sfc-gh-mhofman, line 95: isn't that already in mine PR added?

sf_core/src/config/connection_config.rs
```diff
@@ -65,6 +65,10 @@ pub enum AuthConfig {
         token: SensitiveString,
     },
     NativeOkta(NativeOktaConfig),
+    ExternalBrowser {
+        user: String,
```
sfc-gh-mhofman, line 69: let's check if it's not optional?

sf_core/src/config/connection_config.rs
```diff
@@ -393,16 +400,20 @@ fn build_auth_config(settings: &ParamStore) -> Result<AuthConfig, ConfigError> {
...
-                disable_saml_url_check: settings
-                    .get_bool(DISABLE_SAML_URL_CHECK)
-                    .unwrap_or(false),
-                authentication_timeout_secs,
+                disable_saml_url_check: settings.get_bool(DISABLE_SAML_URL_CHECK).unwrap_or(false),
+                authentication_timeout_secs: parse_authentication_timeout(settings),
             }))
         }
+        "EXTERNALBROWSER" => Ok(AuthConfig::ExternalBrowser {
+            user: non_empty_string(settings, USER).context(MissingParameterSnafu {
```
sfc-gh-mhofman, line 408: is it mandatory? I'm not sure but some drivers might allow to enter user on the browser

sf_core/src/rest/snowflake/external_browser.rs
```diff
@@ -0,0 +1,792 @@
...
+    /// Whether the IdP consented to caching the ID token.
+    /// `None` when the callback was a plain GET redirect (no consent info).
+    /// TODO: Read when ID token caching is implemented (follow-up PR).
+    #[allow(dead_code)]
+    pub consent_cache_id_token: Option<bool>,
+}
+
+/// Allows injecting a no-op browser opener in tests.
+pub(crate) trait BrowserOpener: Send + Sync {
+    fn open(&self, url: &str) -> Result<(), String>;
```
sfc-gh-mhofman, line 40: could be passed as Url?

sfc-gh-mhofman (approved):

sf_core/Cargo.toml
```diff
@@ -79,6 +79,7 @@ lru = "0.12"
 chrono = { version = "0.4", features = ["serde"] }
 url = "2.5"
 html-escape = "0.2"
+open = "5"
```
sfc-gh-mhofman: let's use webbrowser

sfc-gh-mhofman (approved):
some suggestions

sf_core/src/config/connection_config.rs
sfc-gh-asolarski: will be checked and addressed as a followup

sf_core/src/rest/snowflake/external_browser.rs
sfc-gh-asolarski: anyway will be validated right after that with open()

sf_core/Cargo.toml
sfc-gh-asolarski: done

PR #1131, opened 2026-05-12
SNOW-3406388: refresh AWS STS creds on S3 ExpiredToken

sfc-gh-pczajka:
## Summary

- Adds an STS-token refresh path for S3 PUT/GET file transfers. When S3 returns `ExpiredToken` mid-transfer, the storage client re-fetches fresh stage credentials via a new `CredentialRefresher` trait and rebuilds the SDK client before retrying (bounded by `MAX_CREDENTIAL_REFRESH_ATTEMPTS = 2`). Mirrors `StorageCredential.update()` in the Python connector (`file_transfer_agent.py`, `storage_client.py`, `s3_storage_client.py`).
- HEAD failures carry no response body, so the SDK cannot populate the normal error code. S3 mirrors the code into the `x-amz-error-code` response header — we detect `ExpiredToken` there via a new `UploadFileError::HeadExpiredToken` variant.
- Production refresher (`PutGetCredentialRefresher`) re-executes the original PUT/GET SQL under `RefreshContext` and parses fresh `stageInfo.creds` from the response.

## Test plan

- [x] Added 6 wiremock-backed unit tests in `sf_core/src/file_manager/s3_transfer.rs::tests` covering upload-retry, download-retry, HEAD-refresh, no-refresher fail-fast, max-attempts give-up, and non-expired-token no-refresh.
- [x] Full `sf_core` unit test suite passes (721 tests).
- [x] `cargo clippy --workspace --lib` clean.
- [ ] E2E put/get against a real account to confirm no regression on the happy path.

## Notes

Spawned 4 review agents (security, consistency, readability, minimalism) before opening this PR and applied all MAJOR/MEDIUM findings. Two security MEDIUMs were deferred as out-of-scope for this PR:

1. **Refresh amplification across N files**: `stage_info` is cloned per file in `upload_files`/`download_files`, so a file batch after a refresh still starts from the original stale creds; a misbehaving server can induce `2 * N` PUT/GET re-executions. Fix would require caching the last-known-good creds on the refresher. Low-risk against a well-behaved Snowflake control plane.
2. **Ciphertext not zeroized**: `PreparedUpload.data: Vec<u8>` is cloned per attempt and isn't `Zeroize`. Pre-existing across all upload paths (GCS/Azure too), not introduced by this PR.

### DIFF
--- a/sf_core/src/apis/database_driver_v1/connection.rs
+++ b/sf_core/src/apis/database_driver_v1/connection.rs
@@ -2423,11 +2423,14 @@ mod tests {
         use wiremock::matchers::{method, path};
         use wiremock::{Mock, MockServer, ResponseTemplate};
 
+        // Non-session codes stay on the "explicit failure → false" path.
+        // 390112 is intercepted by the RefreshContext and covered by the
+        // heartbeat_refreshes_session_on_390112 test below.
         let server = MockServer::start().await;
         Mock::given(method("POST"))
             .and(path("/session/heartbeat"))
             .respond_with(ResponseTemplate::new(200).set_body_json(
-                serde_json::json!({"success": false, "message": "Session gone", "code": "390112"}),
+                serde_json::json!({"success": false, "message": "Heartbeat failed", "code": "390100"}),
             ))
             .mount(&server)
             .await;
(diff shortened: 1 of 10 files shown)


sfc-gh-pczajka:
reimplemented as #1166

PR #1136, opened 2026-05-13
SNOW-3406388: share refreshed creds across files in a PUT/GET batch

sfc-gh-pczajka:
## Summary

Follow-up to #1131. Closes the batch-amplification gap called out in that PR's security review.

**Problem.** #1131 refreshes STS creds per-file: each file's `upload_to_s3_or_skip` starts from the original (stale) `stage_info.creds`. In a batch of N files with persistent server-side expiry, the client wastes up to N `ExpiredToken` round-trips before each file's refresh loop kicks in — and if N files see expiry simultaneously, each independently re-executes the PUT/GET SQL.

**Fix.** `PutGetCredentialRefresher` now caches its last successful refresh and exposes it via three new trait methods, matching Python's shared `StorageCredential.creds` + `self.timestamp` contract (`file_transfer_agent.py:316-326`):

- `cached_creds()` — storage clients seed their upload/download loops from this before the first attempt, so file #2..N pick up the refresh from file #1 without the wasted round-trip.
- `current_epoch()` + `refresh(seen_epoch)` — concurrent `ExpiredToken` events coalesce: the first caller advances the epoch and performs the server round-trip; later callers find `cache.epoch > seen_epoch` and get the cached creds without re-executing the SQL. This is the direct port of Python's `cur_timestamp < self.timestamp` guard.

Worst case drops from `O(N)` wasted round-trips + `O(N)` refreshes to 1 + 1.

## Test plan

- [x] `cached_creds_is_none_before_first_refresh` — trait contract on a fresh refresher.
- [x] `refresh_with_stale_epoch_returns_cached_without_round_trip` — direct epoch-coalescing test.
- [x] `second_upload_uses_cached_creds_and_skips_refresh` — end-to-end: file #2 starts from the cache and never hits ExpiredToken.
- [x] All 12 pre-existing s3_transfer tests still pass.
- [x] Full `sf_core` suite (724 tests) passes; `cargo clippy --workspace --lib --tests` clean.
- [ ] E2E put/get batch against a real account to confirm the coalescing under real STS rotation.

## Stacking

Based on `pczajka/aws-token-refreshal` (PR #1131). Review/land #1131 first; this PR will auto-retarget to `main` once it merges.

### DIFF
--- a/sf_core/src/apis/database_driver_v1/query.rs
+++ b/sf_core/src/apis/database_driver_v1/query.rs
@@ -117,12 +117,29 @@ async fn perform_put_get(
 /// credentials when S3 reports `ExpiredToken`. Re-execution is serialized
 /// against other queries on the same connection by the `Mutex` inside
 /// `RefreshContext`.
+///
+/// One refresher instance is shared across every file in a PUT/GET batch via
+/// `Arc<dyn CredentialRefresher>`; the cache holds the last successful refresh
+/// plus an ever-increasing version so files 2..N pick up the fresh creds
+/// without another round-trip and concurrent refresh calls coalesce.
 pub(super) struct PutGetCredentialRefresher {
     conn: Arc<Mutex<super::connection::Connection>>,
     command: String,
     query_parameters: QueryParameters,
     http_client: Client,
     retry_policy: RetryPolicy,
+    // `std::sync::Mutex` (not `tokio::sync::Mutex`): the critical sections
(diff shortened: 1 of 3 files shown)


sfc-gh-turbaszek (approved):
Looks reasonable 👌

sfc-gh-pczajka:
Aborted - to be added with actual parallelism change

PR #1137, opened 2026-05-13
feat(session): detect body-level 390112 in read_response_json

sfc-gh-pczajka:
## Summary
- Snowflake GS signals expired sessions with HTTP 200 + `{"success":false,"code":"390112"}` in addition to HTTP 401. Before this change, only the HTTP 401 branch routed to `RefreshContext`; every endpoint except logout surfaced an opaque error for the body-code case.
- Introduces `pub(crate) trait SnowflakeResponseEnvelope` implemented by all response types flowing through `read_response_json`, and short-circuits on a 390112 body code so the existing single-flight refresh machinery (`RefreshContext`) fires uniformly across auth, query, query-status, abort-query, heartbeat, and telemetry.
- `AuthResponse` opts out of the body-code check (login produces session tokens and cannot legitimately emit 390112). `AbortQueryResponse` uses the trait default (no `code` field). The existing HTTP 401 branch stays in place as a belt-and-suspenders safeguard.

## Scope

This is PR1 of a two-PR session-renewal series. Out of scope here:
- Other error codes (390111, 390114, 390195, 390318)
- Re-authentication on master-token expiry
- OAuth refresh-token flow and token-cache wiring
- Configurable heartbeat frequency (PR2, coming next)

## Test plan
- [x] `cargo fmt && cargo clippy -p sf_core --all-targets -- -D warnings`
- [x] `cargo test -p sf_core --lib` — 704 passed, 0 failed
- [x] `cargo test -p sf_core rest::snowflake` — 113 passed (3 new envelope tests in `mod.rs`, new 390112/other-code split in `heartbeat.rs`, new 390112 test in `telemetry.rs`)
- [x] `cargo test -p sf_core database_driver_v1::connection` — 30 passed, including new end-to-end `heartbeat_refreshes_session_on_390112` that mocks heartbeat → 390112 → `/session/token-request` → refreshed heartbeat succeeds
- [x] Pre-existing e2e failures requiring `PARAMETER_PATH` confirmed to exist on `main`, unrelated to this PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1138, opened 2026-05-13
feat(session): add client-driven heartbeat with configurable frequency

sfc-gh-pczajka:
## Summary
- Wires `spawn_heartbeat_task` into `connection_init` so the session-token keepalive loop actually runs when users opt in (it shipped in #743 but had no production caller until now).
- Two new connection options mirror Python and the old driver:
  - `client_session_keep_alive` — enable the background task (default: off)
  - `client_session_keep_alive_heartbeat_frequency` — interval hint in seconds
- `compute_heartbeat_interval` now applies Python's `[master_validity / 16, master_validity / 4]` clamp on top of the existing 3600 s ceiling, so user hints outside the safe window snap to the nearest bound instead of being silently over- or under-cadenced.
- `cleanup_connection` cancels and awaits the task, retiring the `SNOW-2881763` TODO.

## Scope

PR2 of the two-PR session-renewal series — depends on #1137. Out of scope here:
- Master-token re-authentication, OAuth refresh-token flow, ID / MFA token cache wiring (future work).
- Proactive session refresh via `session_expires_at`.

## Test plan
- [x] `cargo fmt && cargo clippy -p sf_core --all-targets -- -D warnings`
- [x] `cargo test -p sf_core --lib` — 715 passed, 0 failed (+11 over `main` since PR1)
- [x] `cargo test -p sf_core --lib config::heartbeat` — 5 passed (parsing + validation)
- [x] `cargo test -p sf_core --lib apis::database_driver_v1::heartbeat` — 12 passed (interval clamp + task lifecycle)
- [x] `cargo test -p sf_core --lib apis::database_driver_v1::connection::tests::initialize_spawns_heartbeat_task_when_enabled_and_cleanup_stops_it` — passes; spawns against a `wiremock` server, asserts ≥1 `/session/heartbeat` hit, then confirms `cleanup_connection` drops the handle.

### DIFF
--- a/sf_core/src/apis/database_driver_v1/connection.rs
+++ b/sf_core/src/apis/database_driver_v1/connection.rs
@@ -20,6 +20,7 @@ use super::validation::{
 };
 use crate::config::ParamStore;
 use crate::config::connection_config::ConnectionConfig;
+use crate::config::heartbeat::HeartbeatConfig;
 use crate::config::logout::LogoutConfig;
 use crate::config::param_registry::{ParamKey, ParamScope, param_names};
 use crate::config::resolver;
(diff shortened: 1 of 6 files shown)


sf_core/src/apis/database_driver_v1/connection.rs
```diff
@@ -269,6 +270,12 @@ impl DatabaseDriverV1 {
                     .map(|v| v.eq_ignore_ascii_case("true"))
                     .unwrap_or(false);
 
+                // User hint for heartbeat cadence. Clamp (see
+                // `compute_heartbeat_interval`) happens at spawn-time.
+                let heartbeat_config = HeartbeatConfig::from_settings(&resolved_snapshot)
```
sfc-gh-jszczerbinski, line 275: Why are we not resolving full hearbeat and config here and resolving it? keep_alive is resolved separately without param def. Please make the config resolution local to one part of the code

sfc-gh-jszczerbinski (changes requested):

sfc-gh-pczajka:
Implemented in #1203

PR #1152, opened 2026-05-13
ci(python): cache rustup and target/ for Linux build_wheels

sfc-gh-turbaszek:
## Summary

Linux `build_wheels` reinstalled rustup (~1-2 min) and recompiled sf_core from scratch (~4-6 min) on every run because the manylinux Docker path only mounted the cargo registry — macOS/Windows already cache `target/` via the `cargo-cache` composite action. This PR closes that gap.

**Changes to `.github/workflows/_build-python-wheels.yml`:**
- Restore/save `target/` via existing `cargo-cache` composite action (per-arch `shared-key` so `manylinux_x86_64` and `manylinux_aarch64` don't collide).
- Cache `~/.rustup` + `~/.cargo/bin` in a separate step keyed only on a version marker (decoupled from `Cargo.lock`).
- Mount the rustup dirs into the Docker container; conditional install (`if [ ! -x /root/.cargo/bin/cargo ]`) with `--profile minimal` to shrink the cached toolchain.
- Hoist the post-build `chown` into its own `if: always()` step so cache save still runs after build failures.

Expected warm run: **~10 min → ~4-5 min** on Linux `build_wheels`.

## Rollback

Two independent kill-switches:
- Rustup cache: bump the `-v1` suffix in the new cache keys.
- Registry + target cache: bump `CACHE_VERSION` in `.github/actions/cargo-cache/action.yml`.

## Test plan

- [ ] First run after merge is cold: expect baseline ~10 min, logs show `"No cached rustup; installing..."` and cache miss on both restores.
- [ ] Both `Save Rust cache (Linux)` and `Save rustup toolchain cache (Linux)` execute post-job; no "Unable to reserve cache" warnings.
- [ ] Second run (warm cache): logs show `"Using cached rustup toolchain"` and cache hits on both restores.
- [ ] Warm `build_wheels (manylinux_x86_64)` and `build_wheels (manylinux_aarch64)` jobs complete in ~4-5 min.
- [ ] `cargo build` output shows `Compiling` only for changed crates, not the full dep tree.
- [ ] Wheels still pass `auditwheel show` manylinux_2_28 policy check.
- [ ] Watch for ENOSPC on `ubuntu-24.04-arm` runner; if observed, reduce `max-target-size-mb` via action input.

### DIFF
--- a/.github/workflows/_build-python-wheels.yml
+++ b/.github/workflows/_build-python-wheels.yml
@@ -247,36 +247,72 @@ jobs:
       # ── sf_core build: Linux (Docker + manylinux for glibc compat) ──
       # Cannot use job-level container: because cibuildwheel needs Docker.
       # Instead, build sf_core via docker run with the manylinux image.
-      - name: Restore Cargo registry cache (Linux)
+      #
+      # Caching strategy: cargo-cache (registry + target/) runs on the HOST,
+      # and its dirs are volume-mounted into the manylinux container so the
+      # in-container `cargo build` reads/writes them directly. The rustup
+      # toolchain (~/.rustup + ~/.cargo/bin) is cached separately — keyed
+      # only on a version marker so Cargo.lock churn doesn't invalidate it.
+      - name: Restore Rust cache (Linux)
         if: runner.os == 'Linux'
+        id: rust-cache-linux
         continue-on-error: true
-        uses: actions/cache@v5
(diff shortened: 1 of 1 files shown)


sfc-gh-akolodziejczyk (approved):
LGTM, if you want to test behavior on the merge queue, you can use `ci:scope-merge` label

PR #1157, opened 2026-05-13
SNOW-3530279: fix pyformat percent escaping with empty params

sfc-gh-asolarski:
## Summary
- Implement `_interpolate_empty_sequences` flag on the Connection class, matching the reference connector's behavior
- When the flag is `False` (default): `%%` is preserved with empty params — no behavior change for direct cursor users
- When the flag is `True` (set by SA dialect in `pre_exec`/`post_exec`): `%%` is unescaped to `%` even with empty params
- This is the mechanism SQLAlchemy uses to unescape doubled percent signs after pyformat compilation

## Context
SQLAlchemy uses `pyformat` paramstyle, which doubles literal `%` to `%%` during query compilation. The SA dialect toggles `connection._interpolate_empty_sequences = True` around execution so the connector unescapes `%%` → `%` via Python's `query % params` formatting. The universal-driver was missing this flag entirely, so `%%` always reached the server verbatim, causing syntax errors in queries with literal percent signs (modulo, `LIKE '%x%'`, stage paths `@%table`).

## Test plan
- **Unit tests**: Verify `interpolate_query` preserves `%%` by default with empty params, unescapes when `interpolate_empty_sequences=True`, and always interpolates when params are non-empty
- **Integration tests** (`TestInterpolateEmptySequences`): Run against both old and new connector:
  - Flag defaults to `False`
  - `%%` causes syntax error with empty params when flag is off (default)
  - `%%` is unescaped and query succeeds when flag is on
  - Works with both empty dict and empty list
  - Normal parameterized queries work regardless of flag
  - `%%` with actual params always unescapes (no flag needed)

### DIFF
--- a/python/src/snowflake/connector/_internal/binding_converters.py
+++ b/python/src/snowflake/connector/_internal/binding_converters.py
@@ -515,7 +515,13 @@ class ClientSideBindingConverter:
             return tuple(cls.process_single_param(param) for param in params)
 
     @classmethod
-    def interpolate_query(cls, query: str, params: Sequence[Any] | Mapping[str, Any] | None) -> str:
+    def interpolate_query(
+        cls,
+        query: str,
+        params: Sequence[Any] | Mapping[str, Any] | None,
+        *,
+        interpolate_empty_sequences: bool = False,
+    ) -> str:
         """Interpolate parameters into query using Python % formatting.
 
         This is the main entry point for client-side binding, mirroring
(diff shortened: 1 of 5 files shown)


sfc-gh-asolarski:
* **#1157** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1157?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1157?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sfc-gh-turbaszek (approved):

PR #1166, opened 2026-05-14
feat(s3): refresh STS credentials on ExpiredToken during PUT/GET

sfc-gh-pczajka:
## Summary

When a PUT/GET runs against an internal S3 stage, Snowflake GS hands out a short-lived STS session token in `stageInfo.creds`. If the token expires mid-transfer, S3 replies with HTTP 400 + `<Error><Code>ExpiredToken</Code></Error>`. Previously the driver surfaced an opaque upload/download error; Python and ODBC handle it by re-executing the original PUT/GET SQL to obtain fresh `stageInfo.creds` and retrying the S3 call once. This change ports that behavior.

Stacked on **#1137** — that PR adds body-level `390112` detection in `read_response_json`, so a refreshed PUT/GET SQL can itself transparently trigger session renewal if GS says the session is expired. The two PRs are independent but complementary.

## Shape

Three layers, each with one job:

- **`file_manager/s3_transfer.rs`** — detects `ExpiredToken` via `ProvideErrorMetadata::code()` (the SDK already parses the XML) and flags it as `StsExpiredToken`. Close-but-different codes (`InvalidToken`, `AccessDenied`, `SignatureDoesNotMatch`, 5xx, throttling) keep their existing paths.
- **`file_manager/mod.rs`** — takes `Option<&mut dyn StageCredsRefresher>`. On `StsExpiredToken` it `take()`s the refresher, calls `refresh()`, swaps `stage_info.creds`, and retries once. Passing `&mut Option<&mut dyn ...>` into the inner calls makes the first `take()` drain the option for the whole batch — one refresh per PUT/GET command, enforced by ownership rather than a state flag. A refreshed credential heals the remaining files in a batch, matching Python's shared `StorageCredential`.
- **`apis/database_driver_v1/query.rs` + `statement.rs`** — builds a `SnowflakeStageCredsRefresher` that re-issues the original PUT/GET SQL via `snowflake_query_with_client` and returns `stageInfo.creds` from the response. Non-PUT/GET queries pass `None`.

## Scope

- S3 only. GCS has its own `TokenExpired` fast-fail path; Azure has no equivalent signal.
- Exactly one refresh per PUT/GET command. A second `ExpiredToken` after the retry propagates.
- AWS secret and session token stay in `SensitiveString` throughout.

## Out of scope / follow-ups

- **Parallel upload refresh.** The file loop is serial today; if it's parallelized, `StageInfo.creds` will need `Arc<RwLock<...>>` and the refresher will need its own single-flight gate (similar to PR #1137's `RefreshContext`).
- **Wiremock-based e2e test for the refresh path.** The existing `put_get_source_compression` wiremock test short-circuits before the S3 call. A real end-to-end test needs a fake S3 HTTP server the AWS SDK can sign against — meaningful scaffolding, left for a dedicated PR. The detector is unit-tested against real `aws_sdk_s3::Error` values.

## Test plan

- [x] `cargo fmt && cargo clippy -p sf_core --all-targets -- -D warnings` — clean
- [x] `cargo test -p sf_core --lib` — 746 passed, 0 failed (3 new detector tests in `s3_transfer.rs`: expired-token hit, close-but-different codes miss, missing-code miss)
- [x] `cargo test -p sf_core --test integration_tests put_get` — passes (existing wiremock test)
- [x] Pre-existing e2e failures requiring `PARAMETER_PATH` confirmed on main, unrelated to this PR

## References

- Python: `snowflake/connector/s3_storage_client.py::_has_expired_token`, `file_transfer_agent.py::StorageCredential.update`
- ODBC: `libsnowflakeclient/cpp/SnowflakeS3Client.cpp::handleError`, `FileTransferAgent.cpp::renewToken`

### DIFF
--- a/sf_core/src/apis/database_driver_v1/query.rs
+++ b/sf_core/src/apis/database_driver_v1/query.rs
@@ -1,4 +1,5 @@
 use super::ColumnMetadata;
+use super::connection::{Connection, RefreshContext};
 use super::global_state::{PutGetResultsetFlavor, WrapperPresets};
 use crate::arrow_utils::ArrowUtilsError;
 use crate::arrow_utils::{boxed_arrow_reader, create_schema};
@@ -7,7 +8,10 @@ use crate::chunks::{
     schema_only_reader, single_chunk_reader,
 };
 use crate::file_manager;
-use crate::file_manager::{DownloadResult, UploadResult, download_files, upload_files};
+use crate::file_manager::{
+    CloudCredentials, DownloadResult, StageCredsCache, StageCredsRefreshError, UploadResult,
+    download_files, upload_files,
+};
 use crate::query_types::RowType;
 use crate::rest;
 use arrow::array::{Array, Int64Array, RecordBatchReader, StringArray};
@@ -15,8 +19,10 @@ use arrow::datatypes::DataType;
 use arrow::error::ArrowError;
 use reqwest::Client;
 use rest::snowflake::query_response::{self, QueryResponseError, RowsetData};
-use snafu::{IntoError, Location, ResultExt, Snafu};
+use snafu::{IntoError, Location, OptionExt, ResultExt, Snafu};
 use std::sync::Arc;
+use std::time::{Duration, Instant};
+use tokio::sync::Mutex;
 
 const PUT_GET_ROWSET_TEXT_LENGTH: u64 = 10000;
 const PUT_GET_ROWSET_FIXED_LENGTH: u64 = 64;
@@ -35,18 +41,47 @@ const ODBC_PUT_ENCRYPTION_LITERAL: &str = "ENCRYPTED";
 /// from legacy libsnowflakeclient. See `ODBC_PUT_ENCRYPTION_LITERAL`.
 const ODBC_GET_ENCRYPTION_LITERAL: &str = "DECRYPTED";
 
+/// Inputs the refresher needs to re-issue the original PUT/GET SQL against GS.
+///
+/// The connection handle is held instead of a snapshot session token: a long
+/// upload batch can outlive its session, and reading the token freshly per
+/// refresh (via `RefreshContext::execute_with_refresh`) lets PR #1137's
+/// session-renewal path heal a 390112 transparently.
+#[derive(Clone)]
+pub struct StageCredsRefreshContext {
+    pub sql: String,
+    pub query_parameters: crate::config::rest_parameters::QueryParameters,
+    pub conn: Arc<Mutex<Connection>>,
+}
+
 /// Executes a PUT/GET file transfer and returns a `RowsetData` variant holding the results.
+///
+/// When `stage_creds_refresh_context` is `Some`, an S3 `ExpiredToken` during a
+/// file transfer triggers a re-issue of the original PUT/GET SQL to obtain fresh
+/// STS credentials and the operation is retried. Non-PUT/GET callers pass `None`.
 pub(super) async fn perform_put_get_transfer(
     command: &str,
     data: &query_response::Data,
     wrapper_presets: &WrapperPresets,
+    stage_creds_refresh_context: Option<StageCredsRefreshContext>,
 ) -> Result<RowsetData, QueryResponseProcessingError> {
+    // Seed the refresher's cache with the initial creds.
+    let initial_creds = data
+        .stage_info_creds()
+        .context(FileTransferPreparationSnafu)?;
+    let mut refresher = stage_creds_refresh_context
+        .zip(initial_creds)
+        .map(|(ctx, initial_creds)| SnowflakeStageCredsRefresher::new(ctx, initial_creds));
+    let refresher_handle = refresher
+        .as_mut()
+        .map(|r| r as &mut dyn file_manager::StageCredsRefresher);
+
     match command {
         "UPLOAD" => {
             let file_upload_data = data
                 .to_file_upload_data(wrapper_presets.put_get_resultset_flavor.clone())
                 .context(FileTransferPreparationSnafu)?;
-            let upload_results = upload_files(&file_upload_data)
+            let upload_results = upload_files(&file_upload_data, refresher_handle)
                 .await
                 .context(FileUploadSnafu)?;
             Ok(RowsetData::Upload(upload_results))
@@ -61,7 +96,7 @@ pub(super) async fn perform_put_get_transfer(
                         FileTransferPreparationSnafu.into_error(e)
                     }
                 })?;
-            let download_results = download_files(file_download_data)
+            let download_results = download_files(file_download_data, refresher_handle)
                 .await
                 .context(FileDownloadSnafu)?;
             Ok(RowsetData::Download(download_results))
@@ -71,6 +106,136 @@ pub(super) async fn perform_put_get_transfer(
         }
         .fail(),
     }
+}
+
+/// Window during which repeated `refresh()` calls return without hitting GS.
+/// Matches ODBC's `FileTransferAgent.cpp` `m_lastRefreshTokenSec` gate (10
+/// minutes), which coalesces rapid-fire refreshes from concurrent uploads.
+const REFRESH_COALESCE_WINDOW: Duration = Duration::from_secs(10 * 60);
+
+/// Refreshes stage credentials by re-executing the original PUT/GET SQL
+/// against Snowflake GS, matching Python's `StorageCredential.update` and
+/// ODBC's `FileTransferAgent::renewToken`. GS returns a brand-new `stageInfo`
+/// per query, so we take only the creds and leave bucket/region/key_prefix
+/// untouched. The new creds land in the shared `StageCredsCache` so every
+/// in-flight transfer in the batch picks them up on its next attempt.
+///
+/// The refresh re-issues the PUT/GET SQL through `RefreshContext::execute_with_refresh`
+/// — if the session token has itself expired by the time we reach this point
+/// (e.g. a long batch upload), the 390112 detection from PR #1137 transparently
+/// renews the session before retrying the SQL.
+///
+/// A 10-minute coalescing window short-circuits subsequent refresh calls
+/// without re-issuing the SQL, keeping us well-behaved against a burst of
+/// `ExpiredToken` responses (long batch upload, concurrent parts in a future
+/// parallel implementation) without either capping retries artificially or
+/// hammering GS.
+struct SnowflakeStageCredsRefresher {
+    ctx: StageCredsRefreshContext,
+    cache: StageCredsCache,
+    last_refresh_at: Option<Instant>,
+}
+
+impl SnowflakeStageCredsRefresher {
+    fn new(ctx: StageCredsRefreshContext, initial_creds: CloudCredentials) -> Self {
+        Self {
+            ctx,
+            cache: StageCredsCache::new(initial_creds),
+            last_refresh_at: None,
+        }
+    }
+}
+
+/// Returns `true` if a refresh recorded at `last` is still considered fresh
+/// at `now` and a new fetch should be coalesced. Extracted so the
+/// time-window logic can be unit-tested without a real `Instant::now()`.
+fn should_coalesce(last: Option<Instant>, now: Instant) -> bool {
+    last.is_some_and(|at| now.saturating_duration_since(at) < REFRESH_COALESCE_WINDOW)
+}
+
+impl file_manager::StageCredsRefresher for SnowflakeStageCredsRefresher {
+    fn refresh(&mut self) -> file_manager::RefreshFuture<'_> {
+        Box::pin(async move {
+            // Coalesce rapid-fire refreshes: if we already fetched creds
+            // within the window, the cache still holds them — nothing to do.
+            if should_coalesce(self.last_refresh_at, Instant::now()) {
+                tracing::debug!("Stage creds refresh coalesced; cache holds recent creds");
+                return Ok(());
+            }
+
+            tracing::info!("Refreshing stage credentials by re-executing PUT/GET SQL");
+            let creds = fetch_fresh_stage_creds(&self.ctx).await?;
+            self.cache.store(creds);
+            self.last_refresh_at = Some(Instant::now());
+            Ok(())
+        })
+    }
+
+    fn cache(&self) -> &StageCredsCache {
+        &self.cache
+    }
+}
+
+/// Re-issues the original PUT/GET SQL through `RefreshContext::execute_with_refresh`
+/// and extracts the fresh `stageInfo.creds` from the response. Going through
+/// `execute_with_refresh` means a session-token expiry mid-batch is healed
+/// transparently by PR #1137's 390112 detection before the SQL is retried.
+async fn fetch_fresh_stage_creds(
+    ctx: &StageCredsRefreshContext,
+) -> Result<CloudCredentials, StageCredsRefreshError> {
+    use crate::file_manager::types::stage_creds_refresh_error::*;
+
+    // `from_arc` is used (not `new`) so that a `close()` raced against an
+    // in-flight refresh is rejected, consistent with the original query path.
+    let mut refresh_ctx = RefreshContext::from_arc(&ctx.conn)
+        .await
+        .context(QueryFailedSnafu)?;
+    // `from_arc` already validates that `http_client` is present (via the
+    // is_closed check + `RefreshContext::new`), so this lookup just clones it.
+    let http_client = ctx
+        .conn
+        .lock()
+        .await
+        .http_client
+        .clone()
+        .expect("http_client present after RefreshContext::from_arc succeeded");
+
+    let query_input = rest::snowflake::QueryInput::new(ctx.sql.clone());
+    let response = refresh_ctx
+        .execute_with_refresh(|session_token| {
+            let http_client = http_client.clone();
+            let query_parameters = ctx.query_parameters.clone();
+            let query_input = query_input.clone();
+            async move {
+                rest::snowflake::snowflake_query_with_client(
+                    &http_client,
+                    query_parameters,
+                    session_token.reveal(),
+                    query_input,
+                    &crate::config::retry::RetryPolicy::default(),
+                    rest::snowflake::QueryExecutionMode::Blocking,
+                )
+                .await
+            }
+        })
+        .await
+        .context(QueryFailedSnafu)?;
+
+    if !response.success {
+        return Err(ServerRejectedSnafu {
+            message: response
+                .message
+                .unwrap_or_else(|| "Unknown error".to_string()),
+        }
+        .build());
+    }
+
+    // The re-issued PUT/GET carries the fresh stageInfo on the response.
+    response
+        .data
+        .stage_info_creds()
+        .context(InvalidStageInfoSnafu)?
+        .context(MissingStageInfoSnafu)
 }
 
 /// Builds an Arrow `RecordBatchReader` from the stored `RowsetData`.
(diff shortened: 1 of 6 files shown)


sf_core/src/file_manager/mod.rs
```diff
@@ -34,67 +37,102 @@ pub async fn upload_files(data: &UploadData) -> Result<Vec<UploadResult>, FileMa
...
+                match upload_to_s3_or_skip(
+                    prepared.clone(),
+                    &stage_info,
+                    file_metadata.target.as_str(),
+                    data.overwrite,
+                )
+                .await
+                {
+                    Ok(s) => break s,
+                    Err(UploadFileError::StsExpiredToken { .. }) if refresher.is_some() => {
```
sfc-gh-pczajka: This retry loop should live inside upload_to_s3_or_skip. It contains s3-specific code, which should not live in generic file_transfer

sf_core/src/file_manager/mod.rs
```diff
@@ -198,30 +239,58 @@ pub async fn download_files(
...
-            .await
-            .context(GcsDownloadSnafu)?,
-        LocationType::Azure => download_from_azure(&data.stage_info, data.src_location.as_str())
-            .await
-            .context(AzureDownloadSnafu)?,
+    refresher: &mut Option<&mut dyn StageCredsRefresher>,
+) -> Result<(DownloadResult, Option<CloudCredentials>), FileManagerError> {
+    let mut stage_info = data.stage_info.clone();
+    let mut refreshed_creds: Option<CloudCredentials> = None;
+    let (raw_data, digest, file_metadata) = loop {
```
sfc-gh-pczajka: Same here - let's move that logic to s3 module

sf_core/src/apis/database_driver_v1/query.rs
```diff
@@ -45,9 +45,32 @@ pub async fn process_query_response(
...
+/// Processes a query response, optionally wiring a stage-credential refresher
+/// for the PUT/GET path. When `refresh_context` is `Some`, an S3 `ExpiredToken`
+/// during the first file transfer triggers a re-issue of the original PUT/GET
+/// SQL to obtain fresh STS credentials and the operation is retried once.
+pub async fn process_query_response_with_refresh_context(
+    data: &query_response::Data,
+    http_client: &Client,
+    prefetch_config: &PrefetchConfig,
+    wrapper_presets: &WrapperPresets,
+    refresh_context: Option<StageRefreshContext>,
```
sfc-gh-pczajka: Let's name it better - stage_creds_refresh_context

sf_core/src/file_manager/mod.rs
sfc-gh-pczajka: Moved the retry loop into `s3_transfer::upload_to_s3_or_skip` — `file_manager/mod.rs` is now a thin dispatcher with no S3-specific branches. `StsExpiredToken` is caught inside the S3 module and never escapes.

sf_core/src/file_manager/mod.rs
sfc-gh-pczajka: Same: moved into `s3_transfer::download_from_s3`.

sf_core/src/apis/database_driver_v1/query.rs
sfc-gh-pczajka: Renamed `StageRefreshContext` -> `StageCredsRefreshContext` and `process_query_response_with_refresh_context` -> `process_query_response_with_stage_creds_refresh_context`.

sfc-gh-pczajka:
* **#1204** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1204?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1166** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1166?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1166?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1137** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1137?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sf_core/src/apis/database_driver_v1/query.rs
```diff
@@ -45,9 +46,38 @@ pub async fn process_query_response(
     http_client: &Client,
     prefetch_config: &PrefetchConfig,
     wrapper_presets: &WrapperPresets,
+) -> Result<QueryResult, QueryResponseProcessingError> {
+    process_query_response_with_stage_creds_refresh_context(
```
sfc-gh-boler: Why `process_query_response_with_stage_creds_refresh_context` is a separate function instead of the body of `process_query_response`?

sf_core/src/file_manager/s3_transfer.rs
```diff
@@ -319,6 +435,20 @@ pub enum UploadFileError {
         #[snafu(implicit)]
         location: Location,
     },
+    /// Emitted by `upload_to_s3` when S3 surfaces `ExpiredToken`. The wrapping
```
sfc-gh-boler: This variant still does leak, because it's a public enum. `upload_to_s3` is private, so it could return a private error too

sf_core/src/file_manager/s3_transfer.rs
```diff
@@ -354,6 +484,19 @@ pub enum DownloadFileError {
         #[snafu(implicit)]
         location: Location,
     },
+    /// Emitted by `do_get_object` when S3 surfaces `ExpiredToken`. Caught and
+    /// handled by `download_from_s3`'s retry loop; never leaks to callers.
+    #[snafu(display("S3 temporary credentials expired (ExpiredToken)"))]
```
sfc-gh-boler: The same comment as for UploadFileError

sfc-gh-boler (commented):
I glanced at this and the first question that comes to my mind is why are we implementing a new refresh mechaism instead of integrating with RefreshContext::execute_with_refresh?

sf_core/src/file_manager/s3_transfer.rs
sfc-gh-pczajka: Fixed

sf_core/src/file_manager/s3_transfer.rs
sfc-gh-pczajka: Fixed

sf_core/src/apis/database_driver_v1/query.rs
sfc-gh-pczajka: Only because Rust does not seem to have defaults. This argument is put/get-specific, IMO it's more readable that other callers does not need to pass "None"

sfc-gh-pczajka:
@sfc-gh-boler Answering:
> I glanced at this and the first question that comes to my mind is why are we implementing a new refresh mechaism instead of integrating with RefreshContext::execute_with_refresh?

Those are two different mechanisms, refreshing two different tokens. execute_with_refresh refreshes session token using master token. This PR refreshes AWS-specific temporary S3 token, used only to authenticate to S3 stages (using session token to refresh itself)

sf_core/src/apis/database_driver_v1/query.rs
```diff
@@ -70,6 +90,94 @@ pub(super) async fn perform_put_get_transfer(
...
+        })
+    }
+}
+
+async fn fetch_fresh_stage_creds(ctx: &StageCredsRefreshContext) -> file_manager::RefreshResult {
+    let query_input = rest::snowflake::QueryInput::new(ctx.sql.clone());
+    let response = rest::snowflake::snowflake_query_with_client(
+        &ctx.http_client,
+        ctx.query_parameters.clone(),
+        ctx.session_token.reveal(),
```
sfc-gh-boler: Since this operates on a session token obtained at the start of the upload, it can also expire. So the stage token refresh should handle session token refresh via RefreshContext::execute_with_refresh. Otherwise this PR builds up on #1137 but doesn't capitalize on it.

sf_core/src/apis/database_driver_v1/query.rs
sfc-gh-boler: Looks like it's a moot point. But in generate, taking Option or a config struct may be often better than expanding the public API surface.

sf_core/src/file_manager/mod.rs
```diff
@@ -34,67 +37,88 @@ pub async fn upload_files(data: &UploadData) -> Result<Vec<UploadResult>, FileMa
...
 
-pub async fn upload_single_file(data: SingleUploadData) -> Result<UploadResult, FileManagerError> {
+/// Uploads one file. On S3 stages, the `refresher` (if any) is used to refresh
+/// STS credentials on `ExpiredToken`; see `s3_transfer::upload_to_s3_or_skip`
+/// for the refresh semantics. Returns any refreshed credentials so the caller
+/// can propagate them to subsequent files in the batch.
+pub async fn upload_single_file(
+    data: SingleUploadData,
+    refresher: &mut Option<&mut dyn StageCredsRefresher>,
+) -> Result<(UploadResult, Option<CloudCredentials>), FileManagerError> {
```
sfc-gh-boler: It is quite iffy that upload_single_file returns potentially refreshed credentials. Especially as part of a public API. Could the refresher keep them instead?

sf_core/src/file_manager/types.rs
```diff
@@ -241,3 +241,19 @@ pub(super) fn percent_encode_path(s: &str) -> String {
...
+pub type RefreshFuture<'a> =
+    std::pin::Pin<Box<dyn std::future::Future<Output = RefreshResult> + Send + 'a>>;
+
+/// Refreshes stage credentials after an AWS STS `ExpiredToken`.
+///
+/// The file manager calls `refresh` at most once per PUT/GET command: a second
+/// `ExpiredToken` propagates instead of retrying again. Implementations decide
+/// the refresh strategy (the Snowflake wrapper re-executes the original
+/// PUT/GET SQL). Callers that don't need refresh pass `None` and fall back to
+/// the pre-existing single-shot behavior.
```
sfc-gh-boler: Is this still true?

sf_core/src/file_manager/s3_transfer.rs
```diff
@@ -100,34 +174,69 @@ async fn upload_to_s3(
...
+    }
 }
 
 /// Downloads a file from S3 and returns the data with optional encryption metadata.
 /// For SSE stages the metadata headers will be absent and `None` is returned.
+///
+/// On AWS `ExpiredToken` the `refresher` (if any) is invoked to fetch fresh
+/// STS credentials and the download is retried. See `upload_to_s3_or_skip`
+/// for the refresh semantics.
+#[allow(clippy::type_complexity)]
```
sfc-gh-boler: Instead of returning a large tuple and soothing the linter, let's return a proper struct.

sf_core/src/file_manager/s3_transfer.rs
```diff
@@ -165,7 +274,45 @@ pub async fn download_from_s3(
...
+    Ok((data, digest, file_metadata, refreshed_creds))
+}
+
+/// Outcome of a single S3 GET attempt. Mirrors `S3UploadOutcome`: the
+/// `StsExpired` arm is internal to this module and never escapes through
+/// `DownloadFileError`. The `Got` payload is boxed because
+/// `GetObjectOutput` is ~800 bytes — far larger than the `aws_sdk_s3::Error`
+/// in `StsExpired`.
+#[derive(Debug)]
+enum S3DownloadOutcome {
```
sfc-gh-boler: Could we merge download/upload outcome enums into one?

sf_core/src/file_manager/types.rs
```diff
@@ -241,3 +241,19 @@ pub(super) fn percent_encode_path(s: &str) -> String {
     }
     encoded
 }
+
+pub type RefreshResult = Result<CloudCredentials, Box<dyn std::error::Error + Send + Sync>>;
```
sfc-gh-boler: Without error trace and snafu we are losing visibility. We should have an error enum here.

sfc-gh-boler (changes requested):

sfc-gh-boler:
> @sfc-gh-boler Answering:
> 
> > I glanced at this and the first question that comes to my mind is why are we implementing a new refresh mechaism instead of integrating with RefreshContext::execute_with_refresh?
> 
> Those are two different mechanisms, refreshing two different tokens. execute_with_refresh refreshes session token using master token. This PR refreshes AWS-specific temporary S3 token, used only to authenticate to S3 stages (using session token to refresh itself)

Sure, they are two different tokens, but the structure of the refresh is the same in both. We should generalize the pattern instead of building a new loops with their own coalesing and token passing. While it is not directly the scope of this PR, I think it would be great to do the refactoring first and then build this PR on top of that.

sf_core/src/apis/database_driver_v1/query.rs
sfc-gh-pczajka: Addressed in 6cc44047: refresh now goes through `RefreshContext::execute_with_refresh`, so a session-token expiry mid-batch is healed by PR #1137's 390112 detection before the re-issued PUT/GET SQL retries.

Also followed up with #1204 — a separate refactor that extracts the shared refresh-and-retry shape (this path + the session path) into a generic `sf_core::refresh` module, per your earlier note.

sf_core/src/file_manager/mod.rs
sfc-gh-pczajka: Done in 6cc44047. The refresher now owns a `StageCredsCache` (`Arc<RwLock<CloudCredentials>>`); `upload_single_file` / `download_single_file` no longer return refreshed creds. The file-transfer layer reads the current creds via `cache.snapshot()` before each S3 attempt and the refresher writes new ones via `cache.store()`.

sf_core/src/file_manager/types.rs
sfc-gh-pczajka: Stale, fixed in 6cc44047. The doc now reads "may be called multiple times per PUT/GET command (e.g. a long batch upload where the refreshed token itself expires) — implementations are expected to coalesce rapid-fire calls themselves."

sf_core/src/file_manager/s3_transfer.rs
sfc-gh-pczajka: Done in 6cc44047. The 4-tuple is replaced with `DownloadFromS3Result { data, digest, file_metadata }`; the `clippy::type_complexity` allow is gone.

sf_core/src/file_manager/s3_transfer.rs
sfc-gh-pczajka: Done in 6cc44047 — collapsed `S3UploadOutcome` and `S3DownloadOutcome` into a single generic `S3Outcome<T>` (later renamed to `S3AttemptError<E>` in #1204).

sf_core/src/file_manager/types.rs
sfc-gh-pczajka: Done in 6cc44047. `RefreshResult = Result<.., Box<dyn Error>>` is replaced with a snafu enum `StageCredsRefreshError` (`QueryFailed`, `ServerRejected`, `MissingStageInfo`) so refresh failures carry error_trace instrumentation through the pipeline.

sfc-gh-pczajka:
@sfc-gh-boler 
>  While it is not directly the scope of this PR, I think it would be great to do the refactoring first and then build this PR on top of that.

I've created a follow-up PR (less hussle) with refactor, will continue works on Monday - #1204

sf_core/src/file_manager/s3_transfer.rs
```diff
@@ -27,24 +36,87 @@ const REQUEST_TIMEOUT_SECS: u64 = 300;
...
+}
+
+/// Asks the refresher (if any) to fetch new stage credentials and updates
+/// `stage_info.creds` from the refresher's cache. Returns:
+/// - `Ok(true)` — credentials rotated; the caller should retry.
+/// - `Ok(false)` — no refresher, or the refresher returned the same creds
+///   (e.g. inside its coalescing window). The caller should propagate the
+///   original AWS error rather than loop.
+/// - `Err(e)` — the refresh itself failed; this is a terminal error.
+async fn try_refresh_creds<E>(
```
sfc-gh-boler: Let's drop the generic E and return `Result<bool, StageCredsRefreshError>` directly, then `.context(upload_file_error::StageCredsRefreshSnafu)?` / `.context(download_file_error::StageCredsRefreshSnafu)?` at the two call sites.

As it is, the manual `From` impls break source location tracking.

sf_core/src/apis/database_driver_v1/query.rs
```diff
@@ -35,18 +41,52 @@ const ODBC_PUT_ENCRYPTION_LITERAL: &str = "ENCRYPTED";
...
 ) -> Result<RowsetData, QueryResponseProcessingError> {
+    // Seed the refresher's cache with the initial creds. The flavor argument
+    // here is irrelevant — we only consume `stage_info.creds`.
+    let initial_creds = if let Ok(d) = data.to_file_upload_data(PutGetResultsetFlavor::Python) {
+        Some(d.stage_info.creds)
+    } else if let Ok(d) = data.to_file_download_data() {
+        Some(d.stage_info.creds)
+    } else {
+        None
+    };
```
sfc-gh-boler: Why do we need to build this object from data if `data` is already available and we should be able to access `creds` directly?

sf_core/src/apis/database_driver_v1/query.rs
```diff
@@ -71,6 +111,137 @@ pub(super) async fn perform_put_get_transfer(
...
+    if !response.success {
+        return Err(ServerRejectedSnafu {
+            message: response
+                .message
+                .unwrap_or_else(|| "Unknown error".to_string()),
+        }
+        .build());
+    }
+
+    // The re-issued PUT/GET carries the fresh stageInfo in the usual shape.
```
sfc-gh-boler: Similar to the earlier question about accessing `creds` directly.

sf_core/src/file_manager/types.rs
```diff
@@ -247,3 +249,89 @@ pub(super) fn percent_encode_path(s: &str) -> String {
...
+}
+
+#[derive(Debug, Snafu, error_trace::ErrorTrace)]
+#[snafu(module, visibility(pub(crate)))]
+pub enum StageCredsRefreshError {
+    /// The Snowflake query that re-issues PUT/GET to obtain new stage
+    /// credentials failed.
+    #[snafu(display("Failed to re-execute PUT/GET SQL during stage credentials refresh"))]
+    QueryFailed {
+        source: Box<dyn std::error::Error + Send + Sync>,
```
sfc-gh-boler: Use `ApiError` instead of Box.

sf_core/src/apis/database_driver_v1/query.rs
```diff
@@ -71,6 +111,137 @@ pub(super) async fn perform_put_get_transfer(
...
+        Self {
+            ctx,
+            cache: StageCredsCache::new(initial_creds),
+            last_refresh_at: None,
+        }
+    }
+}
+
+impl file_manager::StageCredsRefresher for SnowflakeStageCredsRefresher {
+    fn refresh(&mut self) -> file_manager::RefreshFuture<'_> {
```
sfc-gh-boler, line 157: Would be good to have unit tests for the 10-minute coalescing window and the creds_unchanged short-circuit. A fake StageCredsRefresher that records call count, plus an injectable clock for the window, would cover both without needing AWS mocks.

sfc-gh-boler (changes requested):

sf_core/src/file_manager/s3_transfer.rs
sfc-gh-pczajka: Done in 4fc9f37e. `try_refresh_creds` now returns `Result<bool, StageCredsRefreshError>` directly; both call sites wrap with `.context(upload_file_error::StageCredsRefreshSnafu)` / `.context(download_file_error::StageCredsRefreshSnafu)`. The two manual `From<StageCredsRefreshError>` impls (which stamped a default `Location`) are removed, so source-location tracking now goes through snafu.

sf_core/src/apis/database_driver_v1/query.rs
sfc-gh-pczajka: Done in 4fc9f37e. Added `Data::stage_info_creds()` on the response shape; `perform_put_get_transfer` now reads creds directly off `data.stage_info_creds()` instead of building a full `UploadData`/`DownloadData` just to get at `creds`.

sf_core/src/apis/database_driver_v1/query.rs
sfc-gh-pczajka: Done in 4fc9f37e (same `Data::stage_info_creds()` helper). `fetch_fresh_stage_creds` now extracts the rotated creds directly off the refresh response instead of going through `to_file_upload_data`. Added an `InvalidStageInfo` variant on `StageCredsRefreshError` for parse failures, distinct from `MissingStageInfo`.

sf_core/src/file_manager/types.rs
sfc-gh-pczajka: Done in 4fc9f37e. `StageCredsRefreshError::QueryFailed` now carries `Box<ApiError>` (with `#[snafu(source(from(ApiError, Box::new)))]`) instead of `Box<dyn std::error::Error + Send + Sync>`, so the full error chain from `RefreshContext::execute_with_refresh` flows through `error_trace`.

sf_core/src/apis/database_driver_v1/query.rs
sfc-gh-pczajka: Done in 4fc9f37e (+11 unit tests). The 10-minute coalescing logic is extracted as `should_coalesce(last: Option<Instant>, now: Instant) -> bool`, and tests cover before-first-refresh / inside-window / exactly-at-boundary (must NOT coalesce — strictly less-than) / past-window. `creds_unchanged` is covered for same-key, different-key, and non-S3 variants. `try_refresh_creds` is driven via a `FakeRefresher` with an arming hook + atomic call counter, covering no-refresher, rotated-creds, and coalesced-creds (cache unchanged → caller propagates the original AWS error). No AWS mocks needed.

sfc-gh-boler (approved):
lgtm

PR #619, opened 2026-03-19
SNOW-3235553 Parameter Array Attributes

sfc-gh-daniszewski:
## Description

Implements ODBC parameter array execution support by adding six new statement attributes for handling multiple parameter sets in a single execution:

- `SQL_ATTR_PARAMSET_SIZE` - Controls the number of parameter sets to execute (default: 1)
- `SQL_ATTR_PARAM_BIND_TYPE` - Specifies column-wise (0) or row-wise binding layout
- `SQL_ATTR_PARAM_BIND_OFFSET_PTR` - Pointer to offset applied to parameter binding pointers
- `SQL_ATTR_PARAM_STATUS_PTR` - Pointer to per-parameter status array (receives `SQL_PARAM_SUCCESS` or `SQL_PARAM_UNUSED`)
- `SQL_ATTR_PARAMS_PROCESSED_PTR` - Pointer to count of parameters processed
- `SQL_ATTR_PARAM_OPERATION_PTR` - Pointer to per-parameter operation array (allows skipping rows with `SQL_PARAM_IGNORE`)

When `PARAMSET_SIZE > 1`, the driver converts ODBC parameter arrays into JSON array bindings where each parameter's value becomes a JSON array. The implementation handles both column-wise and row-wise binding layouts, respects operation arrays for skipping rows, and writes appropriate status information after execution.

Adds `ArrayStatusPtr` field to APD and ARD descriptors with corresponding get/set operations in the descriptor API. Extends parameter binding conversion to support array execution through `odbc_bindings_to_json_array()` function that processes multiple parameter sets and generates appropriate JSON for the server.

## Testing

Adds comprehensive C++ test suite covering all six parameter array attributes including default values, set/get operations, and functional array execution scenarios. Tests verify proper handling of multiple parameter sets, status reporting, and parameter counting functionality.

PR #617, opened 2026-03-19
SNOW-3235556 Snowflake Custom Statement Attributes

sfc-gh-daniszewski:
## Description

Adds support for Snowflake-specific ODBC statement attributes:

- **`SQL_SF_STMT_ATTR_LAST_QUERY_ID`**: Read-only attribute that returns the query ID of the last executed statement. Returns an empty string before any execution and updates with each new query execution.

- **`SQL_SF_STMT_ATTR_MULTI_STATEMENT_COUNT`**: Configurable attribute for multi-statement execution control. Accepts values of -1 (auto-detect, default), 0 (single statement), or N > 0 (expect exactly N statements). Values less than -1 are rejected with an error.

Updates the statement attribute base value from a hardcoded `2000100` to use the standard ODBC driver attribute base (`0x4000 + 0x106`) for better compatibility. The multi-statement count setting is applied via RPC calls before query execution in both direct execution and prepare/execute paths.

  ## Testing                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                              
  - Added comprehensive C++ test suite covering both attributes                                                                                                                                                                                                                               
  - Tests verify default values, get/set operations, error handling for invalid values, and behavior across multiple executions                                                                                                                                                               
  - `SQL_SF_STMT_ATTR_LAST_QUERY_ID` get before execution and get after `SQLExecDirect` run on both drivers (identical behavior); `SQLPrepare`+`SQLExecute` path and all `SQL_SF_STMT_ATTR_MULTI_STATEMENT_COUNT` assertions are gated with `NEW_DRIVER_ONLY` (BD#55)                         
  - All 9 tests pass on both new and reference (old) drivers

PR #560, opened 2026-03-17
SNOW-3235550 Connection, Session & Database Attributes

sfc-gh-daniszewski:
### TL;DR

Implemented comprehensive support for ODBC connection attributes including access mode, autocommit, transaction isolation, current catalog, connection status, and packet size settings.

### What changed?

Added support for 9 new ODBC connection attributes:
- **SQL_ATTR_ACCESS_MODE** - Read/write mode setting (advisory only)
- **SQL_ATTR_AUTOCOMMIT** - Transaction autocommit behavior with session parameter integration
- **SQL_ATTR_TXN_ISOLATION** - Transaction isolation level (returns READ_COMMITTED with warnings for unsupported levels)
- **SQL_ATTR_CURRENT_CATALOG** - Database switching via `USE DATABASE` statements; returns 24000 if a cursor is open; server ANSI SQL state (e.g. `3D000`) is now propagated through the error chain
- **SQL_ATTR_QUIET_MODE** - Window handle pointer storage
- **SQL_ATTR_PACKET_SIZE** - Network packet size (pre-connection only)
- **SQL_ATTR_CONNECTION_DEAD** - Connection status indicator (read-only)
- **SQL_ATTR_AUTO_IPD** - Automatic IPD population flag (read-only, returns FALSE)

Also fixed `from_protobuf_error` to forward the server's ANSI SQL state from `DriverException.sql_state` through the ODBC error chain, so server-originated errors (e.g. `3D000` for an invalid catalog name) surface with the correct SQLSTATE instead of being discarded.

### How to test?

Run the new test suite `odbc_api_conn_attr_tests` which includes comprehensive test cases for:
- Setting and getting each attribute in connected/disconnected states
- Error handling for read-only attributes and invalid values
- Warning generation for unsupported transaction isolation levels
- Database switching functionality
- Connection status detection

### Why make this change?

These connection attributes are essential ODBC functionality required for compatibility with standard ODBC applications. Many database tools and frameworks rely on these attributes to configure connection behavior, manage transactions, and detect connection status.

### Known limitations

**`SQL_ATTR_TXN_ISOLATION` set — HY011 if a transaction is open** (`connection.rs`)\
Requires either a server round-trip (`SELECT CURRENT_TRANSACTION()`) or local transaction state tracking threaded through statement execution. Tracked in SNOW-3240589. Practically harmless: Snowflake only supports READ_COMMITTED and the set path never executes anything on the server, so the missing check has no behavioral impact.

PR #1171, opened 2026-05-14
SNOW-3235553 Parameter Array Attributes

sfc-gh-daniszewski:
### TL;DR

Adds full ODBC parameter array execution support (`SQL_ATTR_PARAMSET_SIZE > 1`) for both `SQLExecute` and `SQLExecDirect`, including per-set status reporting, operation filtering, and row-wise binding.

### What changed?

**Parameter array execution (`SQLExecute` / `SQLExecDirect`)**
When `SQL_ATTR_PARAMSET_SIZE` is greater than 1, both `exec_direct` and `execute` now iterate over each parameter set, binding and executing the query once per set. The last successful response is used to populate statement state. If all sets fail, the last error is returned. If some sets fail and at least one succeeds, `SQL_SUCCESS_WITH_INFO` with `Warning::RowError` is returned.

**Per-set status and progress tracking**
- `IPD.array_status_ptr` (`SQL_ATTR_PARAM_STATUS_PTR`) is written with `SQL_PARAM_SUCCESS`, `SQL_PARAM_ERROR`, or `SQL_PARAM_UNUSED` for each set.
- `IPD.rows_processed_ptr` (`SQL_ATTR_PARAMS_PROCESSED_PTR`) is updated incrementally after each set so progress is observable mid-batch.
- Sets marked `SQL_PARAM_IGNORE` via `APD.array_status_ptr` (`SQL_ATTR_PARAM_OPERATION_PTR`) are skipped and written as `SQL_PARAM_UNUSED`; they are not counted in `PARAMS_PROCESSED_PTR`.

**Pointer arithmetic for array sets**
`odbc_bindings_to_json` now accepts `set_idx` and `bind_offset` parameters. For each parameter, the data and indicator pointers are advanced using column-wise stride (falling back to the type's fixed size when `buffer_length == 0`) or row-wise stride when `APD.bind_type > 0`.

**New statement attributes**
`SQL_ATTR_PARAMSET_SIZE` (22), `SQL_ATTR_PARAM_BIND_TYPE` (18), `SQL_ATTR_PARAM_BIND_OFFSET_PTR` (17), `SQL_ATTR_PARAM_OPERATION_PTR` (19), `SQL_ATTR_PARAM_STATUS_PTR` (20), and `SQL_ATTR_PARAMS_PROCESSED_PTR` (21) are now handled in both `get_stmt_attr` and `set_stmt_attr`. Setting `PARAMSET_SIZE` to 0 coerces the value to 1 and returns `SQL_SUCCESS_WITH_INFO` (SQLSTATE 01S02).

**APD descriptor**
`ApdDescriptor` gains an `array_status_ptr` field (`SQL_DESC_ARRAY_STATUS_PTR`) for `SQL_ATTR_PARAM_OPERATION_PTR`, with get/set support in the descriptor API.

**Diagnostic records**
Failed parameter sets append a diagnostic record with SQLSTATE `ErrorInRow` (01S01) identifying the 1-based set index.

**Behavior differences documented**
BD#57 (PARAMSET_SIZE=0 coercion) and BD#58 (PARAMS_PROCESSED_PTR counting for ignored sets) are documented in `BehaviorDifferences.yaml`.

### How to test?

A new end-to-end test suite `param_array_attrs.cpp` covers:
- Default values and round-trip get/set for all six new statement attributes.
- `PARAMSET_SIZE > 1` with `SQLExecute`: verifies row count, `PARAMS_PROCESSED_PTR`, and `PARAM_STATUS_PTR`.
- `SQL_PARAM_IGNORE` skipping a set and writing `SQL_PARAM_UNUSED` in the status array.
- `PARAMSET_SIZE == 1` still writes `PARAMS_PROCESSED_PTR` and `PARAM_STATUS_PTR`.
- Two bound parameters with array execution.
- `SQL_C_CHAR` (VARCHAR) array parameters.
- `SQL_NULL_DATA` indicator in one array slot.
- `SQLExecDirect` with `PARAMSET_SIZE > 1`.
- Row-wise binding (`SQL_ATTR_PARAM_BIND_TYPE > 0`) with array execution.

Unit tests in `param_binding.rs` cover `set_idx` pointer arithmetic for column-wise arrays and the `buffer_length == 0` fixed-size fallback.

### Why make this change?

Many ODBC applications (ETL tools, bulk-insert drivers, ORMs) rely on parameter arrays to batch DML statements efficiently. Without this support, each row required a separate `SQLExecute` call, and attributes like `SQL_ATTR_PARAMSET_SIZE`, `SQL_ATTR_PARAM_STATUS_PTR`, and `SQL_ATTR_PARAMS_PROCESSED_PTR` were silently ignored, breaking conformance with the ODBC 3.x specification.

### DIFF
--- a/odbc/src/api/statement.rs
+++ b/odbc/src/api/statement.rs
@@ -14,7 +14,7 @@ use crate::api::query_type::{QueryType, ResultKind};
 use crate::api::runtime::global;
 use crate::api::{
     ApdRecord, Connection, ConnectionState, DaeContext, ExecutionOrigin, FreeStmtOption, IpdRecord,
-    OdbcResult, ParamDirection, ParamValue, SQL_CONCUR_LOCK, SQL_CONCUR_READ_ONLY,
+    OdbcError, OdbcResult, ParamDirection, ParamValue, SQL_CONCUR_LOCK, SQL_CONCUR_READ_ONLY,
     SQL_CONCUR_VALUES, SQL_INSENSITIVE, SQL_NONSCROLLABLE, SQL_NOSCAN_OFF, SQL_NOSCAN_ON,
     SQL_RD_OFF, SQL_RD_ON, SQL_SCROLLABLE, SQL_SENSITIVE, SQL_UNSPECIFIED, SqlType, StatementInner,
     StatementState, stmt_from_handle,
@@ -63,26 +63,31 @@ pub fn exec_direct<E: OdbcEncoding>(
     statement_handle: sql::Handle,
     statement_text: *const E::Char,
     text_length: sql::Integer,
+    warnings: &mut crate::conversion::warning::Warnings,
 ) -> OdbcResult<()> {
     let query = E::read_string(statement_text, text_length)?;
-    exec_direct_impl(statement_handle, &query)
+    exec_direct_impl(statement_handle, &query, warnings)
 }
 
-fn exec_direct_impl(statement_handle: sql::Handle, statement_text: &str) -> OdbcResult<()> {
+fn exec_direct_impl(
+    statement_handle: sql::Handle,
+    statement_text: &str,
+    warnings: &mut crate::conversion::warning::Warnings,
+) -> OdbcResult<()> {
     let guard = stmt_from_handle(statement_handle)?;
     let dbc = guard.conn()?;
-    let mut conn = dbc.connection.lock();
-    let mut inner = guard.inner.lock();
-    tracing::debug!("exec_direct: statement_handle={:?}", statement_handle);
-
-    // Validate connection before committing to NeedData state.
-    let conn_handle = match &conn.state {
-        ConnectionState::Connected { conn_handle, .. } => *conn_handle,
-        ConnectionState::Disconnected => {
-            tracing::error!("exec_direct: connection is disconnected");
-            return DisconnectedSnafu.fail();
+    let conn_handle = {
+        let conn = dbc.connection.lock();
+        match &conn.state {
+            ConnectionState::Connected { conn_handle, .. } => *conn_handle,
+            ConnectionState::Disconnected => {
+                tracing::error!("exec_direct: connection is disconnected");
+                return DisconnectedSnafu.fail();
+            }
         }
     };
+    let mut inner = guard.inner.lock();
+    tracing::debug!("exec_direct: statement_handle={:?}", statement_handle);
 
     if inner.state.as_ref().is_need_data() {
         return InvalidDuringDaeSnafu.fail();
@@ -114,72 +119,210 @@ fn exec_direct_impl(statement_handle: sql::Handle, statement_text: &str) -> Odbc
         return DaeRequiredSnafu.fail();
     }
 
-    let (bindings, _json_owner) = apply_parameter_bindings(&inner.apd, &inner.ipd, false, None)?;
     let stmt_handle = guard.stmt_handle;
     let query_timeout = inner.query_timeout;
     let effective_query = statement_text.to_string();
     let multi_statement_count = inner.multi_statement_count;
 
-    let token = CancellationToken::new();
-    *guard.active_cancel.lock() = Some(token.clone());
+    let array_size = inner.apd.array_size.max(1);
+    let bind_offset = if inner.apd.bind_offset_ptr.is_null() {
+        0
+    } else {
+        unsafe { *inner.apd.bind_offset_ptr as isize }
+    };
+    let param_status_ptr = inner.ipd.array_status_ptr;
+    let rows_processed_ptr = inner.ipd.rows_processed_ptr;
+    let param_operation_ptr = inner.apd.array_status_ptr;
 
-    let response = global().context(OdbcRuntimeSnafu)?.block_on(async |c| {
-        tokio::select! {
-            biased;
-            _ = token.cancelled() => Err(OperationCanceledSnafu.build()),
-            result = async {
-                if multi_statement_count >= 0 {
-                    let mut options = std::collections::HashMap::new();
-                    options.insert(
-                        "multi_statement_count".to_string(),
-                        ConfigSetting {
-                            value: Some(config_setting::Value::IntValue(
-                                multi_statement_count as i64,
-                            )),
-                        },
-                    );
-                    c.statement_set_options(StatementSetOptionsRequest {
+    if array_size == 1 {
+        // Honour SQL_PARAM_IGNORE for the single set (mirrors array-loop behaviour).
+        if !param_operation_ptr.is_null()
+            && unsafe { param_operation_ptr.read() } == SQL_PARAM_IGNORE
+        {
+            unsafe { write_param_status(param_status_ptr, 0, SQL_PARAM_UNUSED) };
+            if !rows_processed_ptr.is_null() {
+                unsafe { rows_processed_ptr.write(0) };
+            }
+            set_state(
+                &mut inner,
+                StatementState::DmlExecuted {
+                    rows_affected: 0,
+                    schema: arrow::datatypes::Schema::empty().into(),
+                    origin: ExecutionOrigin::Direct,
+                },
+            );
+            inner.rows_returned = 0;
+            return Ok(());
+        }
+
+        let (bindings, _json_owner) =
+            apply_parameter_bindings(&inner.apd, &inner.ipd, false, None, 0, bind_offset)?;
+
+        let token = CancellationToken::new();
+        *guard.active_cancel.lock() = Some(token.clone());
+
+        let response = global().context(OdbcRuntimeSnafu)?.block_on(async |c| {
+            tokio::select! {
+                biased;
+                _ = token.cancelled() => Err(OperationCanceledSnafu.build()),
+                result = async {
+                    if multi_statement_count >= 0 {
+                        let mut options = std::collections::HashMap::new();
+                        options.insert(
+                            "multi_statement_count".to_string(),
+                            ConfigSetting {
+                                value: Some(config_setting::Value::IntValue(
+                                    multi_statement_count as i64,
+                                )),
+                            },
+                        );
+                        c.statement_set_options(StatementSetOptionsRequest {
+                            stmt_handle: Some(stmt_handle),
+                            options,
+                        })
+                        .await?;
+                    }
+
+                    c.statement_set_sql_query(StatementSetSqlQueryRequest {
                         stmt_handle: Some(stmt_handle),
-                        options,
+                        query: effective_query,
                     })
                     .await?;
-                }
 
-                c.statement_set_sql_query(StatementSetSqlQueryRequest {
+                    c.statement_execute_query(StatementExecuteQueryRequest {
+                        stmt_handle: Some(stmt_handle),
+                        bindings,
+                        timeout_seconds: if query_timeout > 0 {
+                            Some(query_timeout.min(u32::MAX as sql::ULen) as u32)
+                        } else {
+                            None
+                        },
+                    })
+                    .await
+                } => result.map_err(Into::into),
+            }
+        });
+
+        *guard.active_cancel.lock() = None;
+
+        tracing::info!("exec_direct: response={:?}", response);
+        let response = match response {
+            Ok(r) => r,
+            Err(e) => {
+                if let Some(qid) = e.query_id() {
+                    inner.last_query_id = Some(qid.to_owned());
+                }
+                // write_param_status is null-safe; if the app set PARAM_STATUS_PTR
+                // it guaranteed an array of at least array_size (1) elements.
+                unsafe { write_param_status(param_status_ptr, 0, SQL_PARAM_ERROR) };
+                if !rows_processed_ptr.is_null() {
+                    unsafe { rows_processed_ptr.write(1) };
+                }
+                return Err(e);
+            }
+        };
+
+        update_numeric_settings(&conn_handle, &mut dbc.connection.lock().numeric_settings)?;
+        let apply_result =
+            apply_execute_response(&mut inner, conn_handle, response, ExecutionOrigin::Direct);
+        // write_param_status is null-safe; if the app set PARAM_STATUS_PTR
+        // it guaranteed an array of at least array_size (1) elements.
+        unsafe { write_param_status(param_status_ptr, 0, SQL_PARAM_SUCCESS) };
+        if !rows_processed_ptr.is_null() {
+            unsafe { rows_processed_ptr.write(1) };
+        }
+        // Re-propagate any error that is NOT NoMoreData (which just means a DML
+        // statement affected 0 rows — a valid outcome, not a failure).
+        match apply_result {
+            Ok(()) | Err(OdbcError::NoMoreData { .. }) => {}
+            Err(e) => return Err(e),
+        }
+    } else {
+        // Parameter array execution: send the query text once, then run one
+        // RPC per set via the shared helper.
+        let token = CancellationToken::new();
+        *guard.active_cancel.lock() = Some(token.clone());
+        let set_query_result = global().context(OdbcRuntimeSnafu)?.block_on(async |c| {
+            tokio::select! {
+                biased;
+                _ = token.cancelled() => Err(OperationCanceledSnafu.build()),
+                result = c.statement_set_sql_query(StatementSetSqlQueryRequest {
                     stmt_handle: Some(stmt_handle),
                     query: effective_query,
-                })
-                .await?;
-
-                c.statement_execute_query(StatementExecuteQueryRequest {
-                    stmt_handle: Some(stmt_handle),
-                    bindings,
-                    timeout_seconds: if query_timeout > 0 {
-                        Some(query_timeout.min(u32::MAX as sql::ULen) as u32)
-                    } else {
-                        None
-                    },
-                })
-                .await
-            } => result.map_err(Into::into),
-        }
-    });
-
-    *guard.active_cancel.lock() = None;
-
-    tracing::info!("exec_direct: response={:?}", response);
-    let response = match response {
-        Ok(r) => r,
-        Err(e) => {
-            if let Some(qid) = e.query_id() {
-                inner.last_query_id = Some(qid.to_owned());
+                }) => result.map_err(Into::into),
             }
-            return Err(e);
-        }
-    };
+        });
+        *guard.active_cancel.lock() = None;
+        set_query_result?;
 
-    update_numeric_settings(&conn_handle, &mut conn.numeric_settings)?;
-    apply_execute_response(&mut inner, conn_handle, response, ExecutionOrigin::Direct)?;
+        let ArrayLoopResult {
+            last_response,
+            any_error,
+            total_rows_affected,
+        } = execute_param_array_loop(
+            &mut inner,
+            &guard.active_cancel,
+            stmt_handle,
+            array_size,
+            bind_offset,
+            param_status_ptr,
+            rows_processed_ptr,
+            param_operation_ptr,
+            false,
+            query_timeout,
+            multi_statement_count,
+        )?;
+
+        if let Some(response) = last_response {
+            update_numeric_settings(&conn_handle, &mut dbc.connection.lock().numeric_settings)?;
+            // NoMoreData from apply_execute_response means the last set was a
+            // DML statement that affected 0 rows — that is a valid outcome and
+            // must not override SQL_SUCCESS/SQL_SUCCESS_WITH_INFO for the batch.
+            match apply_execute_response(&mut inner, conn_handle, response, ExecutionOrigin::Direct)
+            {
+                Ok(()) | Err(OdbcError::NoMoreData { .. }) => {
+                    // Overwrite the single-set rows_affected with the batch total.
+                    inner
+                        .state
+                        .transition_or_err::<(), ()>(|s| {
+                            Ok((
+                                match s {
+                                    StatementState::DmlExecuted { schema, origin, .. } => {
+                                        StatementState::DmlExecuted {
+                                            rows_affected: total_rows_affected,
+                                            schema,
+                                            origin,
+                                        }
+                                    }
+                                    other => other,
+                                },
+                                (),
+                            ))
+                        })
+                        .ok();
+                }
+                Err(e) => return Err(e),
+            }
+            if any_error {
+                warnings.push(crate::conversion::warning::Warning::RowError);
+            }
+        } else if !any_error {
+            // All sets were SQL_PARAM_IGNORE — no RPC was ever sent. Transition
+            // the statement to DmlExecuted so it is not left in Created/Prepared.
+            set_state(
+                &mut inner,
+                StatementState::DmlExecuted {
+                    rows_affected: 0,
+                    schema: arrow::datatypes::Schema::empty().into(),
+                    origin: ExecutionOrigin::Direct,
+                },
+            );
+            if !rows_processed_ptr.is_null() {
+                unsafe { rows_processed_ptr.write(0) };
+            }
+        }
+    }
+
     inner.rows_returned = 0;
     Ok(())
 }
@@ -484,7 +627,10 @@ fn prepare_impl(statement_handle: sql::Handle, query: &str) -> OdbcResult<()> {
 }
 
 /// Execute a prepared statement
-pub fn execute(statement_handle: sql::Handle) -> OdbcResult<()> {
+pub fn execute(
+    statement_handle: sql::Handle,
+    warnings: &mut crate::conversion::warning::Warnings,
+) -> OdbcResult<()> {
     tracing::debug!("execute: statement_handle={:?}", statement_handle);
     let guard = stmt_from_handle(statement_handle)?;
     let mut inner = guard.inner.lock();
@@ -544,70 +690,191 @@ pub fn execute(statement_handle: sql::Handle) -> OdbcResult<()> {
             }
         }
     };
-    let (bindings, _json_owner) = apply_parameter_bindings(
-        &inner.apd,
-        &inner.ipd,
-        is_prepared,
-        inner.prepared_param_count,
-    )?;
 
     let stmt_handle = guard.stmt_handle;
     let query_timeout = inner.query_timeout;
     let multi_statement_count = inner.multi_statement_count;
 
-    let token = CancellationToken::new();
-    let _cancel_guard = ActiveCancelGuard::arm(&guard.active_cancel, token.clone());
-
-    let response = global().context(OdbcRuntimeSnafu)?.block_on(async |c| {
-        tokio::select! {
-            biased;
-            _ = token.cancelled() => Err(OperationCanceledSnafu.build()),
-            result = async {
-                if multi_statement_count >= 0 {
-                    let mut options = std::collections::HashMap::new();
-                    options.insert(
-                        "multi_statement_count".to_string(),
-                        ConfigSetting {
-                            value: Some(config_setting::Value::IntValue(
-                                multi_statement_count as i64,
-                            )),
-                        },
-                    );
-                    c.statement_set_options(StatementSetOptionsRequest {
-                        stmt_handle: Some(stmt_handle),
-                        options,
-                    })
-                    .await?;
-                }
-                c.statement_execute_query(StatementExecuteQueryRequest {
-                    stmt_handle: Some(stmt_handle),
-                    bindings,
-                    timeout_seconds: if query_timeout > 0 {
-                        Some(query_timeout.min(u32::MAX as sql::ULen) as u32)
-                    } else {
-                        None
-                    },
-                })
-                .await
-            } => result.map_err(Into::into),
-        }
-    });
-
-    let response = match response {
-        Ok(r) => r,
-        Err(e) => {
-            if let Some(qid) = e.query_id() {
-                inner.last_query_id = Some(qid.to_owned());
-            }
-            return Err(e);
-        }
+    let array_size = inner.apd.array_size.max(1);
+    let bind_offset = if inner.apd.bind_offset_ptr.is_null() {
+        0
+    } else {
+        unsafe { *inner.apd.bind_offset_ptr as isize }
     };
+    let param_status_ptr = inner.ipd.array_status_ptr;
+    let rows_processed_ptr = inner.ipd.rows_processed_ptr;
+    let param_operation_ptr = inner.apd.array_status_ptr;
 
-    tracing::info!("execute: Successfully executed statement");
-    let mut settings = dbc.connection.lock().numeric_settings;
-    update_numeric_settings(&conn_handle, &mut settings)?;
-    dbc.connection.lock().numeric_settings = settings;
-    apply_execute_response(&mut inner, conn_handle, response, origin)?;
+    if array_size == 1 {
+        // Honour SQL_PARAM_IGNORE for the single set (mirrors array-loop behaviour).
+        if !param_operation_ptr.is_null()
+            && unsafe { param_operation_ptr.read() } == SQL_PARAM_IGNORE
+        {
+            unsafe { write_param_status(param_status_ptr, 0, SQL_PARAM_UNUSED) };
+            if !rows_processed_ptr.is_null() {
+                unsafe { rows_processed_ptr.write(0) };
+            }
+            set_state(
+                &mut inner,
+                StatementState::DmlExecuted {
+                    rows_affected: 0,
+                    schema: arrow::datatypes::Schema::empty().into(),
+                    origin,
+                },
+            );
+            inner.rows_returned = 0;
+            return Ok(());
+        }
+
+        let (bindings, _json_owner) = apply_parameter_bindings(
+            &inner.apd,
+            &inner.ipd,
+            is_prepared,
+            inner.prepared_param_count,
+            0,
+            bind_offset,
+        )?;
+
+        let token = CancellationToken::new();
+        let _cancel_guard = ActiveCancelGuard::arm(&guard.active_cancel, token.clone());
+
+        let response = global().context(OdbcRuntimeSnafu)?.block_on(async |c| {
+            tokio::select! {
+                biased;
+                _ = token.cancelled() => Err(OperationCanceledSnafu.build()),
+                result = async {
+                    if multi_statement_count >= 0 {
+                        let mut options = std::collections::HashMap::new();
+                        options.insert(
+                            "multi_statement_count".to_string(),
+                            ConfigSetting {
+                                value: Some(config_setting::Value::IntValue(
+                                    multi_statement_count as i64,
+                                )),
+                            },
+                        );
+                        c.statement_set_options(StatementSetOptionsRequest {
+                            stmt_handle: Some(stmt_handle),
+                            options,
+                        })
+                        .await?;
+                    }
+                    c.statement_execute_query(StatementExecuteQueryRequest {
+                        stmt_handle: Some(stmt_handle),
+                        bindings,
+                        timeout_seconds: if query_timeout > 0 {
+                            Some(query_timeout.min(u32::MAX as sql::ULen) as u32)
+                        } else {
+                            None
+                        },
+                    })
+                    .await
+                } => result.map_err(Into::into),
+            }
+        });
+
+        let response = match response {
+            Ok(r) => r,
+            Err(e) => {
+                if let Some(qid) = e.query_id() {
+                    inner.last_query_id = Some(qid.to_owned());
+                }
+                // write_param_status is null-safe; if the app set PARAM_STATUS_PTR
+                // it guaranteed an array of at least array_size (1) elements.
+                unsafe { write_param_status(param_status_ptr, 0, SQL_PARAM_ERROR) };
+                if !rows_processed_ptr.is_null() {
+                    unsafe { rows_processed_ptr.write(1) };
+                }
+                return Err(e);
+            }
+        };
+
+        tracing::info!("execute: Successfully executed statement");
+        update_numeric_settings(&conn_handle, &mut dbc.connection.lock().numeric_settings)?;
+        let apply_result = apply_execute_response(&mut inner, conn_handle, response, origin);
+        // write_param_status is null-safe; if the app set PARAM_STATUS_PTR
+        // it guaranteed an array of at least array_size (1) elements.
+        unsafe { write_param_status(param_status_ptr, 0, SQL_PARAM_SUCCESS) };
+        if !rows_processed_ptr.is_null() {
+            unsafe { rows_processed_ptr.write(1) };
+        }
+        // Re-propagate any error that is NOT NoMoreData (which just means a DML
+        // statement affected 0 rows — a valid outcome, not a failure).
+        match apply_result {
+            Ok(()) | Err(OdbcError::NoMoreData { .. }) => {}
+            Err(e) => return Err(e),
+        }
+    } else {
+        let loop_result = execute_param_array_loop(
+            &mut inner,
+            &guard.active_cancel,
+            stmt_handle,
+            array_size,
+            bind_offset,
+            param_status_ptr,
+            rows_processed_ptr,
+            param_operation_ptr,
+            is_prepared,
+            query_timeout,
+            multi_statement_count,
+        );
+        inner.rows_returned = 0;
+        let ArrayLoopResult {
+            last_response,
+            any_error,
+            total_rows_affected,
+        } = loop_result?;
+
+        if let Some(response) = last_response {
+            tracing::info!("execute: Successfully executed statement (array mode)");
+            update_numeric_settings(&conn_handle, &mut dbc.connection.lock().numeric_settings)?;
+            // NoMoreData from apply_execute_response means the last set was a
+            // DML statement that affected 0 rows — that is a valid outcome and
+            // must not override SQL_SUCCESS/SQL_SUCCESS_WITH_INFO for the batch.
+            match apply_execute_response(&mut inner, conn_handle, response, origin) {
+                Ok(()) | Err(OdbcError::NoMoreData { .. }) => {
+                    // Overwrite the single-set rows_affected with the batch total.
+                    inner
+                        .state
+                        .transition_or_err::<(), ()>(|s| {
+                            Ok((
+                                match s {
+                                    StatementState::DmlExecuted { schema, origin, .. } => {
+                                        StatementState::DmlExecuted {
+                                            rows_affected: total_rows_affected,
+                                            schema,
+                                            origin,
+                                        }
+                                    }
+                                    other => other,
+                                },
+                                (),
+                            ))
+                        })
+                        .ok();
+                }
+                Err(e) => return Err(e),
+            }
+            if any_error {
+                warnings.push(crate::conversion::warning::Warning::RowError);
+            }
+        } else if !any_error {
+            // All sets were SQL_PARAM_IGNORE — no RPC was ever sent. Transition
+            // the statement to DmlExecuted so it is not left in Created/Prepared.
+            set_state(
+                &mut inner,
+                StatementState::DmlExecuted {
+                    rows_affected: 0,
+                    schema: arrow::datatypes::Schema::empty().into(),
+                    origin,
+                },
+            );
+            if !rows_processed_ptr.is_null() {
+                unsafe { rows_processed_ptr.write(0) };
+            }
+        }
+    }
+
     inner.rows_returned = 0;
     Ok(())
 }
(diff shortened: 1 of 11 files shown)


sfc-gh-daniszewski:
* **#1269** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1269?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1232** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1232?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1171** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1171?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1171?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#617** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/617?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#619](https://github.com/snowflakedb/universal-driver/pull/619) <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/619?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#591** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/591?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#573** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/573?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#560** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/560?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#519** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/519?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

odbc/src/api/statement.rs
```diff
@@ -575,67 +740,218 @@ pub fn execute(statement_handle: sql::Handle) -> OdbcResult<()> {
...
+        }
+    } else {
+        // Parameter array execution: execute once per set, discard all but the last response.
+        let mut sets_processed: usize = 0;
+        let mut last_response: Option<ExecuteQueryResponse> = None;
+        let mut last_error: Option<OdbcError> = None;
+        let mut any_error = false;
+
+        // Send the query text once before the loop (it stays set on the server handle).
+        if let Some(query) = query_to_send {
```
sfc-gh-mhofman: **[A1 | Correctness | High] `multi_statement_count` is silently dropped in array mode**

In `execute` (~L743) and `exec_direct_impl` (~L127), the single-set path calls `c.statement_set_options(... multi_statement_count ...)` before the execute call, but the array branch never does. If an application sets `SQL_ATTR_MULTI_STATEMENT_COUNT` and `SQL_ATTR_PARAMSET_SIZE > 1`, the multi-statement option is silently ignored. Even if this combination is unusual, the silent drop is a bug — the array branch should either honor it or fail with `HY024`/`HYC00`.

Additional risk: a stale `multi_statement_count` from a prior single-set execute on the same stmt handle persists into the array execution because the array branch doesn't reset it.

odbc/src/api/statement.rs
```diff
@@ -575,67 +740,218 @@ pub fn execute(statement_handle: sql::Handle) -> OdbcResult<()> {
...
+                    sets_processed += 1;
+                    if !rows_processed_ptr.is_null() {
+                        unsafe { rows_processed_ptr.write(sets_processed as sql::ULen) };
+                    }
+                    any_error = true;
+                    continue;
+                }
+            };
+
+            let token = CancellationToken::new();
```
sfc-gh-mhofman: **[A2 | Correctness | High] Cancellation token recreated per-set**

A `CancellationToken::new()` is created in every loop iteration and written into `guard.active_cancel`. If `SQLCancel` is called between sets, it cancels the previous set's token (which has already completed). A single batch-wide token would give consistent SQLCancel semantics. Currently, an application calling `SQLCancel` once may need to call it many times to actually stop a long batch.

Spec: "the processing on the statement" — for an array batch that means the whole batch, not the in-flight per-set RPC. Promoted to **HIGH** after panel review (was Medium in initial analysis).

odbc/src/api/statement.rs
```diff
@@ -575,67 +740,218 @@ pub fn execute(statement_handle: sql::Handle) -> OdbcResult<()> {
...
+            match exec_result {
+                Ok(resp) => {
+                    // write_param_status is null-safe; see comment above.
+                    unsafe { write_param_status(param_status_ptr, set_idx, SQL_PARAM_SUCCESS) };
+                    last_response = Some(resp);
+                }
+                Err(e) => {
+                    tracing::error!("execute array: set_idx={set_idx} execution failed: {e}");
+                    if let Some(qid) = e.query_id() {
+                        inner.last_query_id = Some(qid.to_owned());
```
sfc-gh-mhofman: **[A3 | Correctness | Medium] `last_query_id` semantics in array mode**

On per-set error, `inner.last_query_id = Some(qid)` is set. When the batch finishes and `apply_execute_response` is applied for `last_response`, `stmt.last_query_id = Some(query_id)` overwrites with the LAST SUCCESSFUL set's query id. Earlier failed sets' query ids are lost.

This is not necessarily wrong, but it should be documented (and possibly exposed as part of the diag record so support can correlate per-set failures with server query IDs).

odbc/src/api/statement.rs
```diff
@@ -575,67 +740,218 @@ pub fn execute(statement_handle: sql::Handle) -> OdbcResult<()> {
...
+        if let Some(response) = last_response {
+            tracing::info!("execute: Successfully executed statement (array mode)");
+            let mut settings = dbc.connection.lock().numeric_settings;
+            update_numeric_settings(&conn_handle, &mut settings)?;
+            dbc.connection.lock().numeric_settings = settings;
+            // NoMoreData from apply_execute_response means the last set was a
+            // DML statement that affected 0 rows — that is a valid outcome and
+            // must not override SQL_SUCCESS/SQL_SUCCESS_WITH_INFO for the batch.
+            match apply_execute_response(&mut inner, conn_handle, response, origin) {
+                Ok(()) | Err(OdbcError::NoMoreData { .. }) => {}
```
sfc-gh-mhofman: **[A4 | Correctness | Medium] `apply_execute_response` returning `NoMoreData` is swallowed**

The comment correctly notes that NoMoreData from a 0-row last-set DML batch is a valid outcome and shouldn't override SUCCESS. But for `UPDATE … WHERE col = ?` batches where the *last* set's predicate matches nothing, the driver silently turns `SQL_NO_DATA` into `SQL_SUCCESS` — losing information that the caller relies on. Acceptable trade-off for batch INSERT, but worth at least a `tracing::debug!`, and ideally summing `rows_affected` across sets to detect the "every set affected 0 rows" case.

See also A5 (SQLRowCount returns only the last set's row count).

sfc-gh-mhofman (commented):
Automated review pass — findings A1..T14 from `analysis_pr_1171_review.md`. Each inline comment is tagged with `[ID | Category | Severity]`. See `final_pr_1171.md` for the consolidated panel review with re-rated severities and additional findings.

odbc/src/api/statement.rs
```diff
@@ -575,67 +740,218 @@ pub fn execute(statement_handle: sql::Handle) -> OdbcResult<()> {
...
+            // Write the running count after each set so a cancellation handler
+            // or diagnostic inspector can observe progress mid-batch.
+            if !rows_processed_ptr.is_null() {
+                unsafe { rows_processed_ptr.write(sets_processed as sql::ULen) };
+            }
+            match exec_result {
+                Ok(resp) => {
+                    // write_param_status is null-safe; see comment above.
+                    unsafe { write_param_status(param_status_ptr, set_idx, SQL_PARAM_SUCCESS) };
+                    last_response = Some(resp);
```
sfc-gh-mhofman: **[A5 | Correctness | High] `SQLRowCount` returns only the last set's row count**

"Last response wins" means `SQLRowCount` after a 1000-set `INSERT ... VALUES (?)` reports 1, not 1000. Most ETL tools rely on `SQLRowCount` to verify expected vs actual rows.

Per ODBC spec `SQL_PARAM_ARRAY_ROW_COUNTS = SQL_PARC_BATCH` is supposed to return one row count per set, and `SQL_PARC_NO_BATCH` returns one rolled-up count. The current implementation matches neither — it returns only the last set's count. The driver should either (a) sum `rows_affected` across all successful sets and apply that, or (b) expose per-set row counts.

There is also no `SQLGetInfo` advertisement of `SQL_PARAM_ARRAY_ROW_COUNTS` in this PR, so applications cannot detect this behavior at runtime.

odbc/src/api/statement.rs
```diff
@@ -575,67 +740,218 @@ pub fn execute(statement_handle: sql::Handle) -> OdbcResult<()> {
...
+
+        if let Some(response) = last_response {
+            tracing::info!("execute: Successfully executed statement (array mode)");
+            let mut settings = dbc.connection.lock().numeric_settings;
+            update_numeric_settings(&conn_handle, &mut settings)?;
+            dbc.connection.lock().numeric_settings = settings;
+            // NoMoreData from apply_execute_response means the last set was a
+            // DML statement that affected 0 rows — that is a valid outcome and
+            // must not override SQL_SUCCESS/SQL_SUCCESS_WITH_INFO for the batch.
+            match apply_execute_response(&mut inner, conn_handle, response, origin) {
```
sfc-gh-mhofman, line 834: **[A6 | Correctness | High] `SELECT` with `PARAMSET_SIZE > 1` is undefined**

"Last response wins" means only the last set's result set is exposed via `SQLFetch`/`SQLMoreResults`. The ODBC spec defines `SQL_PARAM_ARRAY_SELECTS = SQL_PAS_NO_SELECT` to indicate this is unsupported.

The driver does not advertise `SQL_PAS_NO_SELECT` and does not return `SQL_ERROR` for SELECT with arrays. There is also no test for this scenario. Aggravated by a result-set handle leak: every intermediate set's `ResultSetHandle` is overwritten by `last_response = Some(resp)` without being released — for a 1000-set SELECT, that's 999 leaked server-side cursors.

odbc/src/api/statement.rs
```diff
@@ -112,75 +117,228 @@ fn exec_direct_impl(statement_handle: sql::Handle, statement_text: &str) -> Odbc
...
             }
-            return Err(e);
         }
-    };
+    }
 
-    update_numeric_settings(&conn_handle, &mut conn.numeric_settings)?;
-    apply_execute_response(&mut inner, conn_handle, response, ExecutionOrigin::Direct)?;
     inner.rows_returned = 0;
+    inner.last_sent_max_rows = Some(max_rows);
```
sfc-gh-mhofman: **[A7 | Other | Low] `inner.last_sent_max_rows = Some(max_rows);` added to `exec_direct_impl`**

Behavior change unrelated to the parameter-array feature. Either intentional cleanup or accidental leakage from the array refactor.

Also: `exec_direct` has no concept of a "previous query" (every call is fresh), so this write is never read on the `exec_direct` path — it's effectively dead. Either delete it or call it out in the PR description.

odbc/src/api/statement.rs
```diff
@@ -61,12 +61,17 @@ pub fn exec_direct<E: OdbcEncoding>(
...
 
-fn exec_direct_impl(statement_handle: sql::Handle, statement_text: &str) -> OdbcResult<()> {
+fn exec_direct_impl(
+    statement_handle: sql::Handle,
+    statement_text: &str,
+    warnings: &mut crate::conversion::warning::Warnings,
+) -> OdbcResult<()> {
     let guard = stmt_from_handle(statement_handle)?;
     let dbc = guard.conn()?;
     let mut conn = dbc.connection.lock();
```
sfc-gh-mhofman: **[A8 | Other | Low] Lock held for entire batch**

`exec_direct_impl` locks `dbc.connection` (next line) and `guard.inner` at the top and holds them for the entire batch loop. For large batches this can starve other operations on the same connection.

The `execute` path does NOT hold `conn` across the batch (it re-acquires briefly). Inconsistent. Fix: rework `exec_direct_impl` to use the same pattern as `execute` (read `conn_handle` into a local, drop the conn guard before the loop).

odbc/src/api/statement.rs
```diff
@@ -1562,6 +1985,48 @@ pub fn set_stmt_attr(
             inner.ard.bind_offset_ptr = ptr;
             Ok(())
         }
+        StmtAttr::ParamsetSize => {
+            let size = value_ptr as usize;
+            tracing::debug!("set_stmt_attr: ParamsetSize = {}", size);
+            if size == 0 {
+                tracing::warn!("set_stmt_attr: ParamsetSize value 0 is invalid; coercing to 1");
+                inner.apd.array_size = 1;
```
sfc-gh-mhofman, line 1818: **[A9 | Other | Low] `SQL_ATTR_PARAMSET_SIZE = 0` SQLSTATE choice**

The PR coerces 0 → 1 with `01S02` (Option value changed). Per Microsoft spec the canonical behavior for an out-of-range `PARAMSET_SIZE` is `SQL_ERROR` + `HY024` (Invalid attribute value).

The PR's behavior matches what the old Snowflake driver does (BD#57 documents this), and matches existing patterns in this file for `SQL_ATTR_CURSOR_SCROLLABLE`, `SQL_ATTR_CURSOR_SENSITIVITY` (substituting to a supported value + `01S02`). Considered spec-correct after panel review — keep as is, but please cap upper bound to prevent negative SQLLEN → huge usize → hang (separate finding).

**Note:** After panel review this was downgraded to INFO/Documented behavior; the existing `01S02` is fine.

odbc/src/api/statement.rs
```diff
@@ -1562,6 +1985,48 @@ pub fn set_stmt_attr(
...
+                warnings.push(Warning::OptionValueChanged);
+            } else {
+                inner.apd.array_size = size;
+            }
+            Ok(())
+        }
+        StmtAttr::ParamBindType => {
+            let val = value_ptr as sql::ULen;
+            tracing::debug!("set_stmt_attr: ParamBindType = {}", val);
+            inner.apd.bind_type = val;
```
sfc-gh-mhofman, line 1828: **[A10 | Other | Low] `ParamBindOffsetPtr` typed as `*mut sql::Len` (signed)**

Per ODBC, `SQL_ATTR_PARAM_BIND_OFFSET_PTR` points to `SQLULEN` (unsigned), not `SQLLEN`. The code dereferences and casts to `isize`. Same width, so the bytes read are identical, but a negative-looking value would behave as a huge offset.

The existing `SQL_ATTR_ROW_BIND_OFFSET_PTR` uses the same pattern, so this is just a consistent inaccuracy in the codebase, not a regression. Worth a Rust unit test pinning behavior with a negative `bind_offset` to document the contract.

odbc/src/api/statement.rs
```diff
@@ -892,6 +1208,113 @@ fn apply_parameter_bindings(
     Ok((Some(bindings), Some(json_string)))
 }
 
+/// Like `apply_parameter_bindings` but targets a specific set within a parameter array.
+///
+/// `set_idx` is 0-based; `bind_offset` is the dereferenced value of
+/// `APD.bind_offset_ptr` (pass 0 when the pointer is null).
+fn apply_parameter_bindings_for_set(
```
sfc-gh-mhofman: **[A11 | Other | Low] `apply_parameter_bindings_for_set` does not log success**

The original `apply_parameter_bindings` logs `"apply_parameter_bindings: Found X bound parameters"` and `"Successfully bound parameters"`. The new `_for_set` function skips both `tracing::info!` calls. Minor observability regression — the moment `array_size > 1`, the tracing for parameter binding goes silent.

Resolved naturally if R2 (unify both functions) is applied.

odbc/src/api/statement.rs
```diff
@@ -112,75 +117,228 @@ fn exec_direct_impl(statement_handle: sql::Handle, statement_text: &str) -> Odbc
...
-                c.statement_set_sql_query(StatementSetSqlQueryRequest {
+        update_numeric_settings(&conn_handle, &mut conn.numeric_settings)?;
+        apply_execute_response(&mut inner, conn_handle, response, ExecutionOrigin::Direct)?;
+        // write_param_status is null-safe; if the app set PARAM_STATUS_PTR
+        // it guaranteed an array of at least array_size (1) elements.
+        unsafe { write_param_status(param_status_ptr, 0, SQL_PARAM_SUCCESS) };
+        if !rows_processed_ptr.is_null() {
+            unsafe { rows_processed_ptr.write(1) };
+        }
+    } else {
```
sfc-gh-mhofman, line 240: **[R1 | Refactor | High] Massive duplication between `exec_direct_impl` and `execute`**

The array-execution loop (~120 lines) is almost byte-for-byte duplicated between `exec_direct_impl` (this block, ~L213-338) and `execute` (~L824-953). Only meaningful differences: `ExecutionOrigin::Direct` vs the cached `origin`; `query_to_send: Option<String>`; `is_prepared`/`prepared_param_count`.

A helper such as `execute_param_array_loop(...) -> OdbcResult<(Option<Response>, bool /* any_error */)>` would unify these. Even better: extract a single `execute_one_set(...)` helper for the per-set body so the single-set branch can call it with `set_idx = 0` and the array branch can iterate.

Without this refactor, every future correctness fix (A1/A2/A4/A5/A6) becomes a double-edit — guaranteed source of drift.

odbc/src/api/statement.rs
sfc-gh-mhofman: **[R2 | Refactor | High] `apply_parameter_bindings` and `apply_parameter_bindings_for_set` are 95% identical**

The only differences are the extra `set_idx`/`bind_offset` parameters forwarded to `odbc_bindings_to_json` and dropped tracing calls.

Unify by adding `set_idx: usize` and `bind_offset: isize` to `apply_parameter_bindings` (defaulting to 0, 0 at the single-set call site) and delete `apply_parameter_bindings_for_set`. Pure mechanical dedup, no behavioural risk, also fixes A11 (observability regression) for free.

odbc/src/api/statement.rs
```diff
@@ -112,75 +117,228 @@ fn exec_direct_impl(statement_handle: sql::Handle, statement_text: &str) -> Odbc
...
+        // Parameter array execution: execute once per set, discard all but the last response.
+        let mut sets_processed: usize = 0;
+        let mut last_response: Option<ExecuteQueryResponse> = None;
+        let mut last_error: Option<OdbcError> = None;
+        let mut any_error = false;
+
+        // Send the query text once before the loop (it stays set on the server handle).
+        let token = CancellationToken::new();
```
sfc-gh-mhofman, line 243: **[R3 | Refactor | Medium] Single-set and array-set divergence on cancellation/exec_request**

The single-set path uses one `tokio::select! { ... }` block that chains `statement_set_options` → `statement_set_sql_query` → `statement_execute_query`. The array path splits these into separate `block_on`/`select!` blocks.

Semantic difference for cancellation:
- Single-set: a SQLCancel between set_options and execute_query is caught.
- Array: a SQLCancel between the (pre-loop) `set_sql_query` and the first iteration's `execute_query` is **not** caught (`active_cancel` is cleared between them).

This is subtle and warrants a test, or unifying the loop body with a single `select!`. Subsumed by the A2 fix (batch-wide token).

odbc/src/api/statement.rs
```diff
@@ -892,6 +1208,113 @@ fn apply_parameter_bindings(
...
+        class_origin: ClassOrigin::Odbc3_0,
+        native_error: err.to_native_error(),
+        row_number: Some((set_idx + 1) as sql::Integer),
+        column_number: None,
+        connection_name: String::new(),
+        message_text: format!("Error in parameter set {}: {err}", set_idx + 1),
+    });
+}
+
+// SQL_PARAM_* status codes written to IPD.array_status_ptr.
```
sfc-gh-mhofman, line 1351: **[R4 | Refactor | Low] `SQL_PARAM_*` constants placement**

The new constants are defined locally in `statement.rs`. These are ODBC standard codes — should be referenced from the `sql` (or `odbc_sys`) crate if available, or at least placed in a shared module (`api/types/odbc_types.rs` already defines many constants) with doc comments referencing the spec line.

Also: `SQL_PARAM_DIAG_UNAVAILABLE = 1` overlaps numerically with `SQL_PARAM_IGNORE` (different array, different semantic) — the comment partially covers it but a typed split would be safer (see R9). Promoted to MEDIUM after panel review.

odbc/src/api/statement.rs
```diff
@@ -1562,6 +1985,48 @@ pub fn set_stmt_attr(
             inner.ard.bind_offset_ptr = ptr;
             Ok(())
         }
+        StmtAttr::ParamsetSize => {
```
sfc-gh-mhofman, line 1804: **[R5 | Refactor | Low] Boilerplate get/set for ParamXxx attributes**

The six new `StmtAttr::ParamXxx` branches in both `set_stmt_attr` and `get_stmt_attr` (24 small almost-identical blocks). Two helpers like `set_pointer_attr<T>(value_ptr, &mut field)` / `get_pointer_attr<T>(value_ptr, string_length_ptr, field)` would collapse most of them — and also fold in ~20 existing `RowXxx` branches for a repo-wide cleanup (~50–80 lines deletable).

While doing this, also fix the inconsistent `string_length_ptr` writes: `ParamOperationPtr`, `ParamStatusPtr`, `ParamsProcessedPtr`, `ParamBindOffsetPtr` omit the write; the others include it. Promoted to MEDIUM after panel review.

odbc/src/api/statement.rs
```diff
@@ -575,67 +740,218 @@ pub fn execute(statement_handle: sql::Handle) -> OdbcResult<()> {
...
+                        },
+                    }) => result.map_err(Into::into),
+                }
+            });
+            *guard.active_cancel.lock() = None;
+
+            sets_processed += 1;
+            // Write the running count after each set so a cancellation handler
+            // or diagnostic inspector can observe progress mid-batch.
+            if !rows_processed_ptr.is_null() {
```
sfc-gh-mhofman: **[R6 | Refactor | Low] Inline write of `rows_processed_ptr` is duplicated 4× per fn**

```rust
if !rows_processed_ptr.is_null() {
    unsafe { rows_processed_ptr.write(sets_processed as sql::ULen) };
}
```

Extract a small `write_rows_processed(ptr, count)` analogous to `write_param_status` (already in this PR). 8 occurrences × 3 lines collapses to 8 × 1 line + 7-line helper.

odbc/src/api/statement.rs
```diff
@@ -112,75 +117,228 @@ fn exec_direct_impl(statement_handle: sql::Handle, statement_text: &str) -> Odbc
...
+            }
+        });
+        *guard.active_cancel.lock() = None;
+        set_query_result?;
+
+        for set_idx in 0..array_size {
+            // Honour SQL_PARAM_IGNORE on APD.array_status_ptr.
+            if !param_operation_ptr.is_null()
+                // SAFETY: null checked above; app guarantees array has array_size elements.
+                && unsafe { param_operation_ptr.add(set_idx).read() } == SQL_PARAM_IGNORE
```
sfc-gh-mhofman: **[R7 | Refactor | Low] `unsafe` blocks could be tightened**

Several places do `unsafe { param_operation_ptr.add(set_idx).read() } == SQL_PARAM_IGNORE` inline. Pull the read into a tiny `read_op(ptr, idx) -> u16` helper (or `is_param_set_ignored(ptr, idx) -> bool`) to centralize the safety comment.

Additionally: `write_param_status` is `unsafe fn` with an internal null check — it could be a safe function (push the `unsafe` inside the actual write) which eliminates ~12 redundant `unsafe { }` blocks at call sites. Also use `write_unaligned` since the application-supplied pointer is not guaranteed aligned. Promoted to MEDIUM after panel review.

odbc/src/api/statement.rs
```diff
@@ -892,6 +1208,113 @@ fn apply_parameter_bindings(
...
+/// size the array to at least `SQL_ATTR_PARAMSET_SIZE`).
+unsafe fn write_param_status(ptr: *mut u16, set_idx: usize, status: u16) {
+    if !ptr.is_null() {
+        unsafe { ptr.add(set_idx).write(status) };
+    }
+}
+
+/// Add a diagnostic record identifying the 1-based parameter-set row that
+/// failed during array execution (SQLSTATE 01S01 per ODBC §13.2).
+fn add_param_set_error_diag(stmt: &mut StatementInner, set_idx: usize, err: &OdbcError) {
```
sfc-gh-mhofman, line 1338: **[R8 | Refactor | Low] `add_param_set_error_diag` signature**

Takes `&OdbcError` only for its `to_native_error()`/`Display`. Could take the native error and message directly to decouple from `OdbcError`.

Alternatively, move the helper onto `DiagnosticInfo` as `add_row_error(set_idx, &OdbcError)` so the SQLSTATE / class-origin choices live in one place next to the rest of the diagnostic logic.

odbc/src/api/statement.rs
```diff
@@ -892,6 +1208,113 @@ fn apply_parameter_bindings(
...
+        message_text: format!("Error in parameter set {}: {err}", set_idx + 1),
+    });
+}
+
+// SQL_PARAM_* status codes written to IPD.array_status_ptr.
+const SQL_PARAM_SUCCESS: u16 = 0;
+const SQL_PARAM_ERROR: u16 = 5;
+const SQL_PARAM_UNUSED: u16 = 7;
+// SQL_PARAM_IGNORE is the value placed in APD.array_status_ptr to skip a set.
+const SQL_PARAM_IGNORE: u16 = 1;
```
sfc-gh-mhofman, line 1356: **[R9 | Refactor | Low] Magic `SQL_PARAM_IGNORE` value lives in two semantic spaces**

The constant is used to *read* from `APD.array_status_ptr` and is in the same numeric space as `SQL_PARAM_PROCEED = 0` (operation codes). It must NOT be confused with the status codes used in `IPD.array_status_ptr` (`SQL_PARAM_SUCCESS / ERROR / UNUSED / DIAG_UNAVAILABLE / SUCCESS_WITH_INFO`).

The comment is clear, but separating into two enums (`ParamOpStatus` for APD, `ParamSetStatus` for IPD) would make the cross-namespace mistake a compile error. Particularly important because `SQL_PARAM_DIAG_UNAVAILABLE = 1` collides numerically with `SQL_PARAM_IGNORE = 1` on different arrays. Promoted to MEDIUM after panel review.

odbc/src/api/statement.rs
```diff
@@ -562,6 +717,16 @@ pub fn execute(statement_handle: sql::Handle) -> OdbcResult<()> {
     let sql_text = inner.sql_text.clone();
     let multi_statement_count = inner.multi_statement_count;
 
+    let array_size = inner.apd.array_size.max(1);
```
sfc-gh-mhofman, line 698: **[R10 | Refactor | Low] `array_size: usize` should be `sql::ULen`**

Driver mixes `usize` and `sql::ULen` here. On 32-bit Windows these differ. The current usage casts both ways. Consistency would help.

Also ties into M1/T13: `inner.apd.array_size: usize` accepts any `usize` value, so passing `(SQLULEN)-1` via `SQLSetStmtAttr` sets it to `0xFFFFFFFFFFFFFFFE` and produces an effectively infinite loop. Either store as `sql::ULen` with an explicit clamp, or validate at `set_stmt_attr` time with `HY024` above a sane bound.

odbc_tests/tests/e2e/query/param_array_attrs.cpp
```diff
@@ -0,0 +1,640 @@
...
+
+  // Then PARAMS_PROCESSED_PTR should be 3
+  CHECK(processed == 3);
+
+  // And 3 rows should be in the table
+  auto sel = conn.execute("SELECT COUNT(*) FROM param_array_exec_1");
+  ret = SQLFetch(sel.getHandle());
+  REQUIRE_ODBC(ret, sel);
+  CHECK(get_data<SQL_C_LONG>(sel, 1) == 3);
+}
```
sfc-gh-mhofman, line 303: **[T1 | Tests | High] No partial-success E2E test**

The PR claims "If some sets fail and at least one succeeds, `SQL_SUCCESS_WITH_INFO` with `Warning::RowError` is returned." There is **no** test that exercises this path.

A typical test would: insert into a `UNIQUE` column where the middle set duplicates an earlier row, then verify:
- `ret == SQL_SUCCESS_WITH_INFO`
- `status[0] == SQL_PARAM_SUCCESS`, `status[1] == SQL_PARAM_ERROR`, `status[2] == SQL_PARAM_SUCCESS`
- `processed == 3`
- `SQLGetDiagRec` returns `01S01` with `row_number = 2`

Use `REQUIRE_THAT(OdbcResult(ret, stmt), OdbcMatchers::IsSuccessWithInfo() && OdbcMatchers::HasSqlState("01S01"));` (the file already includes `odbc_matchers.hpp` at line 15 but never uses it).

odbc_tests/tests/e2e/query/param_array_attrs.cpp
```diff
@@ -0,0 +1,640 @@
...
+
```
sfc-gh-mhofman, line 304: **[T2 | Tests | High] No all-fail E2E test**

The PR claims "If all sets fail, the last error is returned." No test for this.

A test should: bind a NULL into a `NOT NULL` column for all sets, verify `SQL_ERROR`, all `SQL_PARAM_ERROR` statuses in the status array, and a diagnostic record for each failed set via `SQLGetDiagRec`.

odbc_tests/tests/e2e/query/param_array_attrs.cpp
```diff
@@ -0,0 +1,640 @@
...
+  auto sel = conn.execute("SELECT COUNT(*) FROM param_array_exec_3");
+  ret = SQLFetch(sel.getHandle());
+  REQUIRE_ODBC(ret, sel);
+  CHECK(get_data<SQL_C_LONG>(sel, 1) == 2);
+
+  // PARAMS_PROCESSED_PTR: new driver counts only non-ignored sets (2);
+  // old driver counts all sets including ignored ones (3) (BD#58)
+  NEW_DRIVER_ONLY("BD#58") { CHECK(processed == 2); }
+  OLD_DRIVER_ONLY("BD#58") { CHECK(processed == 3); }
+}
```
sfc-gh-mhofman, line 374: **[T3 | Tests | High] No `SQLGetDiagRec` / `SQL_DIAG_ROW_NUMBER` test**

The new `add_param_set_error_diag` produces per-set `01S01` records with `row_number = set_idx + 1`. There is no test verifying that this record is retrievable, has the right SQLSTATE, has the right row number (via `SQLGetDiagField(SQL_DIAG_ROW_NUMBER)`), and has the human-readable message.

The file includes `get_diag_rec.hpp` at line 13 but never uses it — that include is signal that diagnostic-level assertions were intended.

odbc_tests/tests/e2e/query/param_array_attrs.cpp
```diff
@@ -0,0 +1,640 @@
...
+  SQLLEN offset = 0;
+  SQLRETURN ret = SQLSetStmtAttr(stmt.getHandle(), SQL_ATTR_PARAM_BIND_OFFSET_PTR, &offset, 0);
+  REQUIRE(ret == SQL_SUCCESS);
+
+  // Then get should return the same pointer
+  SQLLEN* ptr = nullptr;
+  ret = SQLGetStmtAttr(stmt.getHandle(), SQL_ATTR_PARAM_BIND_OFFSET_PTR, &ptr, 0, nullptr);
+  REQUIRE(ret == SQL_SUCCESS);
+  CHECK(ptr == &offset);
+}
```
sfc-gh-mhofman, line 159: **[T4 | Tests | Medium] No bind-offset E2E test**

The bind-offset pointer is set/get tested here but never exercised end-to-end. A real test should bind an array starting at offset 0, set `BIND_OFFSET_PTR` to e.g. `2 * sizeof(SQLINTEGER)`, execute `PARAMSET_SIZE = 3`, and verify the driver picks up the offset elements.

The Rust unit tests for `odbc_bindings_to_json` also pass `bind_offset = 0` only — the whole `bind_offset` parameter is brand-new in this PR and has zero non-zero test coverage. **Promoted to HIGH** after panel review.

odbc_tests/tests/e2e/query/param_array_attrs.cpp
```diff
@@ -0,0 +1,640 @@
...
+  ret = SQLFetch(sel.getHandle());
+  REQUIRE_ODBC(ret, sel);
+  CHECK(get_data<SQL_C_LONG>(sel, 1) == 42);
+  ret = SQLFetch(sel.getHandle());
+  REQUIRE_ODBC(ret, sel);
+  CHECK(get_data<SQL_C_LONG>(sel, 1) == 43);
+  ret = SQLFetch(sel.getHandle());
+  REQUIRE_ODBC(ret, sel);
+  CHECK(get_data<SQL_C_LONG>(sel, 1) == 44);
+}
```
sfc-gh-mhofman, line 641: **[T5 | Tests | Medium] No SELECT-with-array test**

The driver does not advertise `SQL_PAS_NO_SELECT`. A test should either:
- Show that SELECT with array works (and what semantics: last set only, or per-set result sets), OR
- Show that SELECT with array returns `SQL_ERROR` HYC00, OR
- Document the gap with a `# TODO(SNOW-3235553-followup)` style comment.

Linked to A6 (undefined behavior + leaked cursors).

odbc_tests/tests/e2e/query/param_array_attrs.cpp
```diff
@@ -0,0 +1,640 @@
...
+  // When SQLExecute is called with PARAMSET_SIZE = 3 and SQL_C_CHAR parameters
+  ret = SQLExecute(stmt.getHandle());
+  REQUIRE_ODBC(ret, stmt);
+
+  // Then 3 rows should be present
+  auto sel = conn.execute("SELECT COUNT(*) FROM param_array_varchar");
+  ret = SQLFetch(sel.getHandle());
+  REQUIRE_ODBC(ret, sel);
+  CHECK(get_data<SQL_C_LONG>(sel, 1) == 3);
+}
```
sfc-gh-mhofman, line 532: **[T6 | Tests | Medium] No SQL_C_WCHAR array test**

Only `SQL_C_CHAR` is tested. Real ETL drivers commonly use `SQL_C_WCHAR` (Unicode). Stride math for wide chars must be verified separately because `value_type.fixed_size()` returns different values for `Char` vs `WChar`.

Also: the existing `SQL_C_CHAR` test only verifies row count, not the string contents — a future regression where the driver inserts wrong bytes (e.g. null-padded for the BUF=20 fixed-width buffer) would not be caught. Add `SELECT s FROM ... ORDER BY s` and assert each string.

odbc_tests/tests/e2e/query/param_array_attrs.cpp
```diff
@@ -0,0 +1,640 @@
...
+
```
sfc-gh-mhofman, line 533: **[T7 | Tests | Medium] No SQL_C_NUMERIC / SQL_C_BINARY array test**

Type-specific stride math may regress for these (and for `SQL_C_TYPE_TIMESTAMP`). Add at least one array test per C type that has its own `fixed_size()` value.

odbc_tests/tests/e2e/query/param_array_attrs.cpp
sfc-gh-mhofman, line 641: **[T8 | Tests | Medium] No row-wise binding with multiple parameters**

The row-wise test at L602 has a single parameter. With one parameter the column-wise and row-wise paths produce identical pointer arithmetic for the value buffer; the `bind_type` branch is exercised but the *interesting* part (advancing column 2 inside a struct, padding between fields) is not.

Add a row-wise test with a 2-field struct (e.g. `{ SQLINTEGER, SQLLEN, char[20], SQLLEN }`) to exercise the bind_type path more thoroughly. Pin the struct layout with `static_assert(sizeof(Row) == ...)`.

odbc_tests/tests/e2e/query/param_array_attrs.cpp
```diff
@@ -0,0 +1,640 @@
...
+  // Then 3 rows should be in the table with values 42, 43, 44
+  auto sel = conn.execute("SELECT val FROM param_array_exec_5 ORDER BY val");
+  ret = SQLFetch(sel.getHandle());
+  REQUIRE_ODBC(ret, sel);
+  CHECK(get_data<SQL_C_LONG>(sel, 1) == 42);
+  ret = SQLFetch(sel.getHandle());
+  REQUIRE_ODBC(ret, sel);
+  CHECK(get_data<SQL_C_LONG>(sel, 1) == 43);
+  ret = SQLFetch(sel.getHandle());
+  REQUIRE_ODBC(ret, sel);
```
sfc-gh-mhofman, line 639: **[T9 | Tests | Medium] No cancellation test**

`SQLCancel` during a long batch should abort the loop. Given finding A2 (per-set token), there is no test to detect the issue.

A loose test that calls `SQLCancel` from another thread and verifies the batch returns *eventually* would be enough to surface the per-set token bug.

odbc_tests/tests/e2e/query/param_array_attrs.cpp
```diff
@@ -0,0 +1,640 @@
...
+  CHECK(get_data<SQL_C_LONG>(sel, 1) == 44);
```
sfc-gh-mhofman, line 640: **[T10 | Tests | Low] No `multi_statement_count` + array test**

Documenting that the driver does or does not support this combo. Currently the array branch silently drops the option (see A1).

odbc_tests/tests/e2e/query/param_array_attrs.cpp
sfc-gh-mhofman, line 641: **[T11 | Tests | Low] No DAE + array test**

The driver currently rejects DAE in array mode (via `find_dae_params`). A test should confirm this and pin down the SQLSTATE.

Note: `find_dae_params` only inspects set 0's indicator — if DAE is set on later sets, the rejection silently misroutes (see panel finding X7). Worth a test verifying the rejection covers all set indices.

odbc_tests/tests/e2e/query/param_array_attrs.cpp
```diff
@@ -0,0 +1,640 @@
...
+  REQUIRE_ODBC(ret, stmt);
+
+  // When SQLExecute is called with the default array_size == 1
+  ret = SQLExecute(stmt.getHandle());
+  REQUIRE_ODBC(ret, stmt);
+
+  // Then both pointers should be written: processed == 1, status[0] == SQL_PARAM_SUCCESS
+  CHECK(processed == 1);
+  CHECK(status[0] == SQL_PARAM_SUCCESS);
+}
```
sfc-gh-mhofman, line 410: **[T12 | Tests | Low] `PARAMSET_SIZE == 1` + `PARAM_OPERATION_PTR = SQL_PARAM_IGNORE` on the only set**

Edge case: array of size 1 with the single set ignored. Should result in 0 rows inserted, `SQL_PARAM_UNUSED` in the status array, `processed == 0`.

Note: the single-set branch at `statement.rs:137-212` does NOT consult `param_operation_ptr` — it unconditionally executes. So setting `SQL_PARAM_IGNORE` on the only set has no effect today, which contradicts the spec. Either always run the loop, or add a `SQL_PARAM_IGNORE` check to the single-set path.

odbc_tests/tests/e2e/query/param_array_attrs.cpp
```diff
@@ -0,0 +1,640 @@
...
+  // driver returns SQL_SUCCESS and leaves the value as 0 (BD#57)
+  NEW_DRIVER_ONLY("BD#57") {
+    REQUIRE(ret == SQL_SUCCESS_WITH_INFO);
+    SQLULEN value = 0;
+    ret = SQLGetStmtAttr(stmt.getHandle(), SQL_ATTR_PARAMSET_SIZE, &value, 0, nullptr);
+    REQUIRE(ret == SQL_SUCCESS);
+    CHECK(value == 1);
+  }
+  OLD_DRIVER_ONLY("BD#57") { REQUIRE(ret == SQL_SUCCESS); }
+}
```
sfc-gh-mhofman, line 90: **[T13 | Tests | Low] `SQLSetStmtAttr SQL_ATTR_PARAMSET_SIZE` negative value**

What happens if the app passes `-1`? Cast to `usize` yields huge value → effectively infinite loop. Should be rejected with `HY024`.

A test that calls `SQLSetStmtAttr(SQL_ATTR_PARAMSET_SIZE, (SQLPOINTER)-1, 0)` and asserts the response would close this. Also useful as documentation of the chosen behavior.

odbc_tests/tests/e2e/query/param_array_attrs.cpp
sfc-gh-mhofman, line 641: **[T14 | Tests | Low/Informational] No Gherkin coverage**

PRs in this repo sometimes add Gherkin feature files when integration behavior is feature-level. ODBC parameter arrays would map nicely to a `Scenario Outline:` with `Examples:`. Not required (the panel review marks this as INFO — skip; existing parameter-array tests live in C++ E2E and adding `.feature` would create a second source of truth), but flagged here for completeness.

sfc-gh-mhofman (changes requested):

sfc-gh-daniszewski:
Thanks for the thorough review! Here is the status of each finding after the latest push:

### Fixed in this iteration

**A1** (multi_statement_count silently dropped in array mode) — Fixed. `execute_param_array_loop` receives and forwards `multi_statement_count` to each set's RPC call, matching the single-set path.

**A2** (cancellation token recreated per-set) — Fixed. A single `batch_token` is created once in `execute_param_array_loop` and stored in `active_cancel` before the loop. All sets share it, so one `SQLCancel` call cancels the entire batch.

**A5** (SQLRowCount returns only last set's count) — Fixed. `execute_param_array_loop` accumulates `total_rows_affected` across all sets and the caller overwrites the execute response's `rows_affected` with the batch total, so `SQLRowCount` reports the sum.

**A6** (SQL_PARAM_ARRAY_SELECTS / SQL_PARAM_ARRAY_ROW_COUNTS not advertised) — Fixed. Added `InfoType::ParamArraySelects` (147) and `InfoType::ParamArrayRowCounts` (153); both return `SQL_PAS_BATCH = 1` / `SQL_PARC_BATCH = 1`. Updated `get_info_tests.cpp` to remove `SKIP_NEW_DRIVER_NOT_IMPLEMENTED` and assert driver-specific expected values. Added BehaviorDifferences entry #60.

**A7** (dead `last_sent_max_rows` write in exec_direct_impl) — No longer present. This field was removed by a separately-merged commit (SNOW-3573289) and the restack dropped it entirely.

**A8** (conn lock held for entire batch) — Fixed. `exec_direct_impl` now extracts `conn_handle` in a short-lived scope and drops the lock before the loop.

**R1** (duplication between exec_direct_impl and execute) — Fixed. Both paths call `execute_param_array_loop` for `array_size > 1`.

**R2** (apply_parameter_bindings_for_set duplication) — Fixed. Single `apply_parameter_bindings` with `set_idx`/`bind_offset` parameters; no separate `_for_set` variant.

**T1** (partial-success test) — Added at line 647: inserts into a UNIQUE column where set 1 duplicates set 0; verifies `SQL_SUCCESS_WITH_INFO`, `status[1] == SQL_PARAM_ERROR`, processed == 3, and that the surviving rows are present.

**T2** (all-fail test) — Added at line 704: inserts 3 NULLs into a NOT NULL column; verifies `SQL_ERROR`, all `SQL_PARAM_ERROR` statuses, processed == 3.

**T3** (SQLGetDiagField row number) — Added at line 758: verifies per-set `01S01` diagnostic records with correct `SQL_DIAG_ROW_NUMBER` values.

**T4** (bind-offset E2E test) — Added at line 803: sets `BIND_OFFSET_PTR = 2 * sizeof(SQLINTEGER)`, executes PARAMSET_SIZE=3, verifies the driver picks up elements [2,3,4] of the array.

**T5** (SELECT with array) — Added at line 850: documents last-set-only semantics; verifies `SQLFetch` returns only the last set's value.

**T6** (SQL_C_WCHAR array) — Added at line 889: inserts 3 wide-char strings and verifies stored values.

**T7** (SQL_C_TYPE_TIMESTAMP array) — Added at line 933.

**T8** (row-wise binding with 2 parameters) — Added at line 969: `{ SQLINTEGER, SQLLEN, char[20], SQLLEN }` struct, verifies both column values per row.

**T11** (DAE rejection) — Added at line 1029: verifies `SQL_DATA_AT_EXEC` with PARAMSET_SIZE > 1 still produces `SQL_NEED_DATA`.

**T12** (PARAMSET_SIZE == 1 with SQL_PARAM_IGNORE) — Added at line 1058: the single-set path now consults `param_operation_ptr`, consistent with the multi-set path.

**T13** (PARAMSET_SIZE = (SQLULEN)-1) — Added at line 1106: verified `SQL_ERROR` + `HY024` is returned.

---

### Intentional design choices / deferred

**A3** (last_query_id semantics) — The last successful set's query_id is exposed; earlier failed sets' IDs appear in the per-set `01S01` diagnostic message text, which is sufficient for support correlation.

**A4** (NoMoreData from last-set DML swallowed) — Kept as `SQL_SUCCESS`; a `tracing::debug!` added for the case. Returning `SQL_NO_DATA` would break callers on batch UPDATEs where the last predicate matches 0 rows, which is the normal case.

**A9** (SQL_ATTR_PARAMSET_SIZE = 0 SQLSTATE choice) — Kept as `01S02` to match old-driver behavior (BD#58) and be consistent with the `SQL_ATTR_CURSOR_SCROLLABLE` / `SQL_ATTR_CURSOR_SENSITIVITY` coercion patterns in the same file.

**A10** (ParamBindOffsetPtr signed type) — Same pattern as the existing `SQL_ATTR_ROW_BIND_OFFSET_PTR`; pre-existing inaccuracy, not a regression from this PR.

**A11** — Resolved by R2 (single function retains the tracing calls).

**R3** (cancellation gap between set_sql_query and execute_query) — Small; both calls are inside a `tokio::select!` against `batch_token`. Acceptable for this iteration.

**R4–R10** (refactoring suggestions) — Noted for a follow-up cleanup PR to keep this PR's scope manageable.

**T9** (cancellation test) — A2 is verifiable by code inspection; a cross-thread timing test deferred.

**T10** (multi_statement_count + array test) — A1 is fixed; a dedicated combo test deferred.

**T14** (Gherkin coverage) — Not added; C++ E2E is the primary test layer for ODBC.

sfc-gh-daniszewski:
Closing as moved to https://github.com/snowflake-eng/universal-driver/pull/59

PR #1175, opened 2026-05-14
Align CLIENT_APP_ID and CLIENT_ENVIRONMENT.APPLICATION with old connector behavior

sfc-gh-pczajka:
##  Summary                                                                                                                                  
         
  Separate CLIENT_APP_ID (driver family — used by the server for feature gating) from CLIENT_ENVIRONMENT.APPLICATION (the user-facing      
  application name) across both the Python and ODBC wrappers. This restores the two-field shape the old connectors have always had.        
   
  - Rust core: add client_application to ClientInfo, register it as a first-class ParamDef, and route it to CLIENT_ENVIRONMENT.APPLICATION 
  in the login request. client_app_id continues to drive both CLIENT_APP_ID and the User-Agent header.
  - Python wrapper: stop conflating application with client_app_id. The user's application=... kwarg now flows to client_application;      
  client_app_id is pinned to "PythonConnector".                                                                                            
  - ODBC wrapper: route the APPLICATION connection-string key and SQL_SF_CONN_ATTR_APPLICATION to the canonical client_application setting
  (previously written to an unregistered "application" key that sf_core silently dropped).                                                 
                                                                       
##  Behavior comparison                                                                                                                      
                                                                       
  User passes application="MyApp" (or ODBC equivalent APPLICATION=MyApp). Wire values shown for the login request:                         
```
  ┌───────────────────────────────────────────────────┬────────────────────┬───────────────────────────────────────┐                       
  │                      Driver                       │   CLIENT_APP_ID    │    CLIENT_ENVIRONMENT.APPLICATION     │
  ├───────────────────────────────────────────────────┼────────────────────┼───────────────────────────────────────┤                       
  │ Old Python connector (snowflake-connector-python) │ PythonConnector    │ MyApp                                 │
  ├───────────────────────────────────────────────────┼────────────────────┼───────────────────────────────────────┤
  │ Old ODBC driver (snowflake-odbc)                  │ ODBC               │ MyApp                                 │                       
  ├───────────────────────────────────────────────────┼────────────────────┼───────────────────────────────────────┤                       
  │ Universal — Python (before)                       │ ❌ MyApp           │ MyApp                                 │                       
  ├───────────────────────────────────────────────────┼────────────────────┼───────────────────────────────────────┤                       
  │ Universal — ODBC (before)                         │ ODBC               │ ❌ ODBC (user value silently dropped) │
  ├───────────────────────────────────────────────────┼────────────────────┼───────────────────────────────────────┤                       
[...]

PR #1177, opened 2026-05-14
NO-SNOW: Add internal_application_name and internal_application_version connection kwargs

sfc-gh-pczajka:
##  Summary                                                                                                 

Add `internal_application_name` and `internal_application_version` connection kwargs to the Python wrapper. Tools that re-host the Python driver — SnowSQL, Snow CLI — use these to identify themselves to the server instead of "PythonConnector". The old Python connector supports both; this restores parity.
                                                                                                                                           
##  Logic vs. old connector                                                                                                                  
   
  **Scenario**: caller passes internal_application_name="SnowSQL", internal_application_version="1.2.3". Wire values shown for the login  request:                   
```                                                                  
  ┌───────────────────────────────────────────────────┬───────────────────┬────────────────────┐                                           
  │ Driver                                            │ CLIENT_APP_ID     │ CLIENT_APP_VERSION │                                           
  ├───────────────────────────────────────────────────┼───────────────────┼────────────────────┤                                           
  │ Old Python connector (snowflake-connector-python) │ SnowSQL           │ 1.2.3              │                                           
  ├───────────────────────────────────────────────────┼───────────────────┼────────────────────┤                                           
  │ Universal — Python (before)                       │ ❌ PythonConnector │ ❌ <core default>  │                                          
  ├───────────────────────────────────────────────────┼───────────────────┼────────────────────┤                                           
  │ Universal — Python (this PR)                      │ ✅ SnowSQL        │ ✅ 1.2.3            │                                           
  └───────────────────────────────────────────────────┴───────────────────┴────────────────────┘                                           
 ```
 
  The "before" row is what the universal driver does today: the kwargs aren't recognized, so they get forwarded to the Rust core as unknown options (warning, then ignored) and the wire values fall back to defaults — SnowSQL can't identify itself. 
                                                                                                                     
  ODBC is intentionally omitted: the old ODBC driver never exposed these kwargs, and this PR doesn't add them to the new ODBC wrapper either.
                                                                                                                                           
##  Relationship to #1175                                                                                                                    
   
  Stacked on #1175 — must merge after it.                                                                                                  
                             
  #1175 separates the two identity slots (CLIENT_APP_ID vs. CLIENT_ENVIRONMENT.APPLICATION) and pins client_app_id/client_app_version as wrapper-internal params (caller can't set them directly — _INTERNAL_PARAMS raises ProgrammingError). This PR plugs into that:
 *  internal_application_name / internal_application_version are the wrapper-owned levers that populate those internal slots, mirroring the old connector's contract.  

*  application= (CLIENT_ENVIRONMENT.APPLICATION) and internal_application_name (CLIENT_APP_ID) stay independent — set both and they land in different fields, exactly like in the old connector.
                                                                                                                                           
##  Rust core changes                                                                                                                        
   
  Single param added to `sf_core/src/config/param_registry.rs`:                                                                              
                             
  - client_app_version — ParamScope::Connection, ValueType::String, no default. Consumed by rest_parameters.rs to populate ClientInfo.version; when unset the Rust core falls back to its compiled-in CARGO_PKG_VERSION (so omitting internal_application_version doesn't change behavior).                                                                                                                
                             
client_app_id and application were already registered by #1175 — this PR only adds the version slot.                                     
   
##  Test plan                                                                                                                                
                             
  - Unit tests in test_connection_config.py: override paths, omission paths, popping from _extra, client_app_version rejected as user      
  kwarg.                     
  - Unit tests in test_connection.py (TestInternalApplicationName): override behaviour, defaults, independence from application, combined  
  usage.                                                                                                                                   
  - WireMock integration tests in test_client_identity.py validating actual login-request payload JSON for CLIENT_APP_ID /
  CLIENT_APP_VERSION / CLIENT_ENVIRONMENT.APPLICATION independence.                                                                        
  - All sf_core lib tests pass.

### DIFF
--- a/python/src/snowflake/connector/_internal/connection_config_mixin.py
+++ b/python/src/snowflake/connector/_internal/connection_config_mixin.py
@@ -268,8 +268,16 @@ class ConnectionConfigMixin:
         * ``application`` - validates against ``_APPLICATION_RE`` and defaults
           to ``_APPLICATION_NAME``. The value is forwarded to the Rust core as
           ``application`` and lands in CLIENT_ENVIRONMENT.APPLICATION on the
-          wire. ``client_app_id`` (CLIENT_APP_ID) is always the driver name so
-          server-side feature gating tied to the client type keeps working.
+          wire. ``client_app_id`` (CLIENT_APP_ID) is the driver name by default
+          so server-side feature gating tied to the client type keeps working;
+          tools that re-host the driver (SnowSQL, Snow CLI) can override it via
+          ``internal_application_name``.
+        * ``internal_application_name`` / ``internal_application_version`` -
+          popped from kwargs and mapped to ``client_app_id`` (CLIENT_APP_ID)
+          and ``client_app_version`` (CLIENT_APP_VERSION) respectively. When
+          omitted, ``client_app_id`` defaults to ``_APPLICATION_NAME`` and
+          ``client_app_version`` is left unset so the Rust core falls back to
+          its compiled-in default.
         * ``autocommit`` - type-checked (must be ``bool``), then merged into
           ``session_parameters["AUTOCOMMIT"]``.
         """
@@ -278,11 +286,20 @@ class ConnectionConfigMixin:
                 "Cannot pass both a ConnectionConfig object and keyword arguments. Use one or the other."
             )
 
+        internal_app_name: Any = None
+        internal_app_version: Any = None
         if config is None:
             if connection_name is not None:
                 kwargs["connection_name"] = connection_name
             if connections_file_path is not None:
                 kwargs["connections_file_path"] = connections_file_path
+            # Pop the ``internal_application_*`` overrides before
+            # ``from_kwargs`` runs: they are wrapper-internal levers that
+            # ultimately populate ``client_app_id`` / ``client_app_version``,
+            # which ``_INTERNAL_PARAMS`` forbids end users from setting
+            # directly.
+            internal_app_name = kwargs.pop("internal_application_name", None)
+            internal_app_version = kwargs.pop("internal_application_version", None)
             config = cls.from_kwargs(**kwargs)
         else:
             if connection_name is not None:
@@ -303,11 +320,20 @@ class ConnectionConfigMixin:
                 raise ProgrammingError(f"Invalid application name: {application!r}")
         else:
             raise ProgrammingError(f"Invalid application parameter (must be a non-empty string): {application!r}")
-        # CLIENT_APP_ID is the driver identity — always the driver name.
+        # CLIENT_APP_ID is the driver identity. Defaults to the driver name,
+        # but can be overridden via ``internal_application_name`` (used e.g.
+        # by SnowSQL / Snow CLI to identify themselves to the server).
         # CLIENT_ENVIRONMENT.APPLICATION (server-side ``application``) carries
-        # the user-facing application name. This mirrors the old connector
-        # where the two values are independent.
-        config.client_app_id = cls._APPLICATION_NAME  # type: ignore[attr-defined]
+        # the user-facing application name; the two values are independent.
+        if isinstance(internal_app_name, str) and internal_app_name:
+            config.client_app_id = internal_app_name  # type: ignore[attr-defined]
+        else:
+            config.client_app_id = cls._APPLICATION_NAME  # type: ignore[attr-defined]
+        # CLIENT_APP_VERSION is only sent when the caller explicitly overrides
+        # it via ``internal_application_version`` — otherwise the Rust core
+        # falls back to its compiled-in default.
+        if isinstance(internal_app_version, str) and internal_app_version:
+            config.client_app_version = internal_app_version  # type: ignore[attr-defined]
 
         if config.autocommit is not None:
             if not isinstance(config.autocommit, bool):
--- a/python/src/snowflake/connector/_internal/connection_config_mixin.py
+++ b/python/src/snowflake/connector/_internal/connection_config_mixin.py
@@ -23,6 +23,7 @@ from typing import Any, ClassVar, TypeVar
 from snowflake.connector._internal._private_key_helper import normalize_private_key
 from snowflake.connector._internal.protobuf_gen.database_driver_v1_pb2 import ConfigSetting
 from snowflake.connector.errors import ProgrammingError
+from snowflake.connector.version import __version__ as _DRIVER_VERSION
 
 
 _Self = TypeVar("_Self", bound="ConnectionConfigMixin")
@@ -276,8 +277,9 @@ class ConnectionConfigMixin:
           popped from kwargs and mapped to ``client_app_id`` (CLIENT_APP_ID)
           and ``client_app_version`` (CLIENT_APP_VERSION) respectively. When
           omitted, ``client_app_id`` defaults to ``_APPLICATION_NAME`` and
-          ``client_app_version`` is left unset so the Rust core falls back to
-          its compiled-in default.
+          ``client_app_version`` defaults to the Python driver's own
+          ``__version__`` — matching the old connector, which sends its own
+          version on the wire when the caller does not override it.
         * ``autocommit`` - type-checked (must be ``bool``), then merged into
           ``session_parameters["AUTOCOMMIT"]``.
         """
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: python/src/snowflake/connector/connection.py [hunk 1]
```diff
@@ -211,8 +211,13 @@ def __init__(
         self._session_parameters = SessionParametersProxy(self.db_api, self.conn_handle)
         self._connection_info = ConnectionInfoProxy(self.db_api, self.conn_handle)
 
+        # ``internal_application_name`` / ``internal_application_version`` are
+        # consumed by ``ConnectionConfig.from_connection_args`` to override
+        # CLIENT_APP_ID / CLIENT_APP_VERSION; strip them here so they don't
+        # surface again via ``conn.kwargs``.
+        _stripped_keys = {"internal_application_name", "internal_application_version"}
```
sfc-gh-turbaszek: Are we sure we can stop sending those fields in requests? Do we fully understand the story behind the keys?

python/src/snowflake/connector/connection_config.py
```diff
@@ -286,6 +289,7 @@ class ConnectionConfig:
             "auto_cleanup",
             "autocommit",
             "client_app_id",
+            "client_app_version",
```
sfc-gh-turbaszek: Overriding client_app_id or client_app_version in ConnectionConfig should probably raise an error similar to what we do for `internal` values in old driver

sfc-gh-turbaszek (changes requested):
Please confirm how the field is used e2e with server and if we can simply drop it / rewrite it

python/src/snowflake/connector/_internal/connection_config_mixin.py
```diff
@@ -268,8 +268,16 @@ def from_connection_args(
...
+          wire. ``client_app_id`` (CLIENT_APP_ID) is the driver name by default
+          so server-side feature gating tied to the client type keeps working;
+          tools that re-host the driver (SnowSQL, Snow CLI) can override it via
+          ``internal_application_name``.
+        * ``internal_application_name`` / ``internal_application_version`` -
+          popped from kwargs and mapped to ``client_app_id`` (CLIENT_APP_ID)
+          and ``client_app_version`` (CLIENT_APP_VERSION) respectively. When
+          omitted, ``client_app_id`` defaults to ``_APPLICATION_NAME`` and
+          ``client_app_version`` is left unset so the Rust core falls back to
+          its compiled-in default.
```
sfc-gh-pczajka: It should fall back to driver's version. Please confirm with old driver code

python/tests/integ/telemetry/test_client_identity.py
```diff
@@ -80,3 +80,50 @@ def test_custom_application_only_affects_client_environment_application(int_test
...
+
+def test_internal_application_version_overrides_client_app_version(int_test_connection_factory, wiremock):
+    wiremock.add_mapping("auth/login_success_jwt.json")
+
+    connection = int_test_connection_factory(
+        server_url=wiremock.http_url(),
+        internal_application_version="1.2.3",
+    )
+    try:
+        data = _login_request_data(wiremock)
```
sfc-gh-turbaszek:
```suggestion
        data = _lget_ogin_request_data(wiremock)
```
nit

sfc-gh-turbaszek (approved):

python/src/snowflake/connector/connection.py
sfc-gh-pczajka: Outdated — that whole `_stripped_keys` / `self.kwargs` block has been removed in the latest revision. The wrapper no longer keeps a shadow copy of constructor kwargs; `internal_application_name` / `internal_application_version` are popped inside `ConnectionConfig.from_connection_args`, mapped to `client_app_id` / `client_app_version`, and from there flow through `to_options` to the Rust core like every other parameter. There is nothing left to strip after `__init__`, so the question is moot.

Re: "do we fully understand the story behind the keys?" — yes:
- In the legacy connector, `internal_application_name` / `internal_application_version` are the kwargs that re-hosters (SnowSQL, Snow CLI) use to set `CLIENT_APP_ID` / `CLIENT_APP_VERSION` on the wire (see `DEFAULT_CONFIGURATION` in the old `connection.py` — defaults to `CLIENT_NAME` / `CLIENT_VERSION`, written to the login body in `auth/_auth.py`).
- In the universal driver, `client_app_id` / `client_app_version` are wrapper-internal (forbidden via `_INTERNAL_PARAMS` since #1175). This PR adds the user-facing `internal_application_*` kwargs that populate them, restoring parity.

python/src/snowflake/connector/connection_config.py
sfc-gh-pczajka: Done — addressed in #1175 and validated here. `_INTERNAL_PARAMS = {"client_app_id", "client_app_version"}` in `ConnectionConfigMixin.from_kwargs` raises `ProgrammingError` if either is passed as a user kwarg, mirroring the old connector's `INTERNAL_PARAMS` guard.

Tests:
- `tests/unit/test_connection_config.py::test_client_app_id_kwarg_rejected`
- `tests/unit/test_connection_config.py::test_client_app_version_kwarg_rejected`

python/tests/integ/telemetry/test_client_identity.py
sfc-gh-pczajka: The current code already uses `_login_request_data(wiremock)` (defined at the top of the file and used by every test in it). The suggestion text `_lget_ogin_request_data` looks like a typo in the suggestion — declining as a no-op.

sfc-gh-pczajka:
@sfc-gh-turbaszek re: "confirm how the field is used e2e with server and if we can simply drop it / rewrite it" — confirmed; we cannot drop. Summary of the e2e flow:

**Wire fields, default behaviour:**
- `CLIENT_APP_ID` — sent in the JSON body of `POST /session/v1/login-request` under `data.CLIENT_APP_ID`. Server uses this for client-family feature gating (e.g. driver-specific behaviour and metric attribution). Defaults to `"PythonConnector"`.
- `CLIENT_APP_VERSION` / `CLIENT_APP_VERSION_FULL` — same login body. `_FULL` is the raw `__version__`; `CLIENT_APP_VERSION` is the digits-only stripped form (`sf_core/src/rest/snowflake/mod.rs::strip_version_suffix`) required by server-side feature gating that does numeric comparisons.

**Why re-hosters need to override them (the case for `internal_application_*`):**
Tools that re-host the Python connector — SnowSQL and Snow CLI — must identify themselves on the wire as their own product, not as `PythonConnector`. The server-side telemetry and feature gates differ for these tools. The legacy connector exposes `internal_application_name` / `internal_application_version` exactly for this purpose (`DEFAULT_CONFIGURATION` in old `connection.py`, written into the login body by `auth/_auth.py::_authenticate`). Without the overrides, every Snow CLI / SnowSQL session would mis-attribute as `PythonConnector` — that's a behavioural regression vs. the old driver, hence this PR.

**Why we can't just rewrite `application` to do this:**
`application` already maps to `CLIENT_ENVIRONMENT.APPLICATION` — a *separate* wire field that carries the user-customisable label. In the old connector the two are independent: a user can set `application="MyApp"` while a tool also sets `internal_application_name="SnowSQL"`, landing in different fields. Conflating them would break that contract (and is what #1175 fixed in the opposite direction by splitting them).

**E2e validation (wiremock integration tests in `python/tests/integ/telemetry/test_client_identity.py`):**
- `test_internal_application_name_overrides_client_app_id` — caller passes `internal_application_name="SnowSQL"`; asserts `CLIENT_APP_ID == "SnowSQL"`, `CLIENT_ENVIRONMENT.APPLICATION == "PythonConnector"`.
- `test_internal_application_name_with_application_uses_both_independently` — both kwargs together; asserts `CLIENT_APP_ID == "SnowSQL"` and `CLIENT_ENVIRONMENT.APPLICATION == "SNOWCLI.STAGE.COPY"` independently.
- `test_internal_application_version_overrides_client_app_version` — caller passes `internal_application_version="1.2.3"`; asserts `CLIENT_APP_VERSION == "1.2.3"`.

The default-fallback case is covered by `test_login_request_contains_correct_client_identity` in the same file (asserts `CLIENT_APP_VERSION_FULL == __version__` when the kwarg is omitted).

**Bug fix in latest revision:** the previous commit left `client_app_version` unset when `internal_application_version` was omitted, falling back to `sf_core`'s `CARGO_PKG_VERSION` (which is `"0.0.0"` since the crate is unversioned). Old connector defaults to the Python driver's own `CLIENT_VERSION`. Fixed: `from_connection_args` now defaults `client_app_version` to `snowflake.connector.version.__version__` when the kwarg is omitted, matching old-connector wire behaviour. Tests updated accordingly.

python/tests/integ/telemetry/test_client_identity.py
sfc-gh-pczajka: Sorry, I misread the suggestion as a typo. Renamed `_login_request_data` → `_get_login_request_data` (file-wide).

PR #1194, opened 2026-05-14
SNOW-3535355: python - fix connector logger propagation for external log capture

sfc-gh-asolarski:
### TL;DR

Fix log propagation so that `snowflake.connector` and `snowflake.connector._core` loggers propagate to the root logger by default, enabling `pytest caplog` and application-level handlers to capture connector logs without extra configuration.

### What changed?

Both `snowflake.connector` and `snowflake.connector._core` loggers now have `propagate=True` by default (matching standard Python library logging practice). Previously, propagation was disabled at module load, which silently blocked `pytest caplog` and any root-level log handlers from seeing connector output.

`setup_logging()` now only disables propagation on a logger when it actually attaches a dedicated `StreamHandler` to that logger. This prevents duplicate output (once via the dedicated handler, once via the root handler chain) while leaving propagation intact when the application has configured its own handlers. The level is always updated regardless.

The `register_default_logger_callback` export was removed from the top-level `__init__.py`, and the startup info log now goes through `get_connector_logger()` instead of `logging.getLogger("snowflake.connector")` directly.

E2E tests in `test_close.py` were refactored to instantiate `WiremockClient` inline via a context manager rather than relying on a shared `wiremock` fixture parameter. This makes each test self-contained and eliminates state bleed between test phases. The best-effort error handling test was also simplified to use `caplog` directly now that propagation is enabled, removing the need for a temporary `FileHandler` workaround.

A new test file `test_log_propagation.py` was added covering:

- `caplog` capturing logs from `snowflake.connector`, `snowflake.connector._core`, and child loggers
- The Rust→Python FFI callback path producing records visible to `caplog`
- `setup_logging()` producing no duplicate output even when a root-level handler is present

### Why make this change?

The previous `propagate=False` default caused silent failures in downstream libraries (e.g. snowflake-sqlalchemy) whose tests asserted on `caplog.text` for connector log output. Because propagation was disabled, those records never reached the root logger and `caplog` captured nothing, causing test assertions to fail. The fix aligns the connector with the standard Python logging guidance for libraries: add a `NullHandler` and leave propagation on, letting the application or test framework decide where logs go.

### DIFF
--- a/python/src/snowflake/connector/__init__.py
+++ b/python/src/snowflake/connector/__init__.py
@@ -10,7 +10,6 @@ from __future__ import annotations
 from typing import Any
 
 from . import util_text  # noqa: F401 - backward compatibility re-exports
-from ._internal.api_client.c_api import register_default_logger_callback  # noqa: F401
 from ._internal.decorators import pep249
 from .connection import Connection, SnowflakeConnection
 from .connection_config import ConnectionConfig
--- a/python/src/snowflake/connector/_internal/api_client/__init__.py
+++ b/python/src/snowflake/connector/_internal/api_client/__init__.py
@@ -1,7 +1,6 @@
 """Initialize the core API with the default logger callback."""
 
-import logging
-
+from ..logging import get_connector_logger
 from .c_api import c_logger_callback, sf_core_init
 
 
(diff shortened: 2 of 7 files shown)


sfc-gh-asolarski:
* **#1194** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1194?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1194?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sfc-gh-turbaszek (approved):

PR #1200, opened 2026-05-15
SNOW-3521142: python - expose TelemetryClient

sfc-gh-asolarski:


sfc-gh-asolarski:
* **#1200** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1200?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1200?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1031** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1031?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

protobuf/database_driver_v1.proto
```diff
@@ -781,6 +781,16 @@ message TelemetrySendWrapperErrorRequest {
   string error_source = 3;
 }
 
+message TelemetrySendUserLogRequest {
```
sfc-gh-turbaszek: Imo `UserLog` is misleading. This is not produced by users and it's not a "log" per-se. How about `JsonRequest`?

python/src/snowflake/connector/_internal/telemetry.py
```diff
@@ -1,43 +1,58 @@
...
 
-    def __init__(self, db_api: DatabaseDriverClient, conn_handle: ConnectionHandle) -> None:
+    def __init__(
+        self,
+        db_api: DatabaseDriverClient | None = None,
+        conn_handle: ConnectionHandle | None = None,
+    ) -> None:
         self._db_api = db_api
         self._conn_handle = conn_handle
+        self._enabled = True
```
sfc-gh-turbaszek, line 39: There should be probably round trip from core. Te core checks telemetry enebled on server and uses this accordingly  
  
On the second hand the core handles it correctly, should python be stub only?

protobuf/database_driver_v1.proto
sfc-gh-asolarski: done

python/src/snowflake/connector/_internal/telemetry.py
sfc-gh-asolarski: You're right — the core is the authoritative gatekeeper here. It checks `CLIENT_TELEMETRY_ENABLED` from the server session parameters and only registers the session for export when enabled.

We don't sync that parameter to Python (it would be a stale point-in-time read that can't track mid-session changes). The Python `_enabled` flag only serves the explicit programmatic disable use-case (`connection.telemetry_enabled = False`) — it avoids unnecessary RPC serialization overhead on every `@api_telemetry`-decorated call. When server-side telemetry is disabled, RPCs from Python become cheap no-ops (core has no session registered, returns immediately).

sfc-gh-fpawlowski (commented):
Overall looks good, left few comments 👍

python/src/snowflake/connector/_internal/telemetry.py
```diff
@@ -17,22 +18,37 @@
...
-    sends runtime events — sf_core attaches the stored identity automatically.
+    Server-side disablement (``CLIENT_TELEMETRY_ENABLED=false``) is handled
+    entirely by the Rust core which skips session registration — RPCs from
+    Python become cheap no-ops. The Python flag only serves the explicit
+    programmatic disable use-case, avoiding unnecessary RPC overhead.
     """
 
     def __init__(self, conn_handle: ConnectionHandle) -> None:
         self._conn_handle = conn_handle
+        self._enabled = True
```
sfc-gh-fpawlowski, line 39: Nit: we may have more such mirroring use cases over time - seems error-prone if they end up not synched. Structure that would propagate any changes in the param to core param with appropriate name may be needed in the future.

python/src/snowflake/connector/connection.py
```diff
@@ -572,10 +574,14 @@ def _get_connection_info(self, include_master_token: bool = False) -> Connection
             include_master_token=include_master_token,
         )
 
+    @property
+    def _telemetry(self) -> _PublicTelemetryClient:
```
sfc-gh-fpawlowski, line 578: is it no longer backward compatibility decorated?

python/src/snowflake/connector/telemetry.py
```diff
@@ -1,19 +1,108 @@
-"""BACKWARD COMPATIBILITY MODULE ONLY"""
+"""Public telemetry API.
```
sfc-gh-fpawlowski, line 1: Is publicly available telemetry smth we wish to have exposed or we are have it due to backward compat?

python/src/snowflake/connector/_internal/telemetry.py
```diff
@@ -51,3 +68,22 @@ def send_wrapper_error(self, exception_type: str, error_source: str) -> None:
             )
         except Exception:
             logger.debug("Failed to send wrapper_error telemetry", exc_info=True)
+
+    def send_user_log(self, telemetry_data: TelemetryData) -> None:
```
sfc-gh-fpawlowski, line 72: Is this name correct or should it be more related to adding a log to the batch?

python/src/snowflake/connector/telemetry.py
```diff
@@ -1,19 +1,108 @@
...
+
+    Buffering and flush are owned by the Rust core.
+    """
+
+    def __init__(
+        self,
+        core: CoreTelemetryClient | None = None,
+        **kwargs: Any,
+    ) -> None:
+        self._core = core
```
sfc-gh-fpawlowski, line 86: Name core may be confusing with UD core -> maybe core_telemetry_client or smth like that, since we use composition over inheritance ?

sf_core/src/protobuf/apis/database_driver_v1/mod.rs
```diff
@@ -940,6 +940,43 @@ impl DatabaseDriver for DatabaseDriverImpl {
 
         Ok(TelemetrySendResponse {})
     }
+
+    #[instrument(name = "DatabaseDriverV1::telemetry_send_json", skip(self, input))]
+    async fn telemetry_send_json(
```
sfc-gh-fpawlowski, line 945: is it correct naming if we are not actually sending the request?

sfc-gh-fpawlowski (commented):
Re-added comments as they seem to be missing in the previous review somehow.

PR #25, opened 2025-08-05
SNOW-2247757: Large result set in Python

sfc-gh-jszczerbinski:

PR #1203, opened 2026-05-15
feat(session): wire client_session_keep_alive[_heartbeat_frequency]

sfc-gh-pczajka:
## Summary
- Register `CLIENT_SESSION_KEEP_ALIVE` and `CLIENT_SESSION_KEEP_ALIVE_HEARTBEAT_FREQUENCY` in the sf_core param registry; thread the user-provided heartbeat frequency through `compute_heartbeat_interval` with `[master_validity/16, master_validity/4]` clamping (Python parity, defaults to `master/16` when unspecified, capped at 1h).
- Mirror both values into login session parameters so the server sees them on first contact (matches old Python connector behavior).
- Implement the matching `Connection.client_session_keep_alive[_heartbeat_frequency]` properties on the Python wrapper (previously `NotImplementedError`); regenerate `connection_config.py`.

## Test plan
- [x] `cargo test -p sf_core --lib` (790 pass; 6 new in `heartbeat.rs`, 1 new in `param_registry.rs`)
- [x] `pytest tests/unit/test_connection.py tests/unit/test_connection_config.py` (191 pass; 4 new in `test_connection_config.py`, 3 new in `test_connection.py`)
- [ ] integration test against a live account with `client_session_keep_alive=True` and varying `client_session_keep_alive_heartbeat_frequency` values (in/below/above range)

### DIFF
--- a/python/src/snowflake/connector/connection.py
+++ b/python/src/snowflake/connector/connection.py
@@ -645,10 +645,33 @@ class Connection(ErrorHandlerMixin):
         """Whether to keep the session active with periodic heartbeat requests."""
         return self.config.client_session_keep_alive
 
+    @client_session_keep_alive.setter
+    @backward_compatibility
+    def client_session_keep_alive(self, value: bool | None) -> None:
+        """Setter retained for compatibility with snowflake-connector-python.
+
+        The Universal Driver consumes ``client_session_keep_alive`` only at
+        connect time, so assigning to this attribute on a live ``Connection``
+        does not start or stop the heartbeat task. Pass the value as a
+        constructor kwarg instead.
+        """
+        self.config.client_session_keep_alive = value
+
     @property
     def client_session_keep_alive_heartbeat_frequency(self) -> int | None:
         """The frequency in seconds of heartbeat requests when session keep-alive is enabled."""
         return self.config.client_session_keep_alive_heartbeat_frequency
+
+    @client_session_keep_alive_heartbeat_frequency.setter
+    @backward_compatibility
+    def client_session_keep_alive_heartbeat_frequency(self, value: int | None) -> None:
+        """Setter retained for compatibility with snowflake-connector-python.
+
+        The Universal Driver consumes the heartbeat frequency only at connect
+        time, so post-connect assignments do not reschedule the running
+        heartbeat task. Pass the value as a constructor kwarg instead.
+        """
+        self.config.client_session_keep_alive_heartbeat_frequency = value
 
     @property
     def client_prefetch_threads(self) -> int:
--- a/python/src/snowflake/connector/connection.py
+++ b/python/src/snowflake/connector/connection.py
@@ -643,20 +643,12 @@ class Connection(ErrorHandlerMixin):
     @property
     def client_session_keep_alive(self) -> bool | None:
         """Whether to keep the session active with periodic heartbeat requests."""
-        raise NotImplementedError("client_session_keep_alive is not yet implemented")
-
-    @client_session_keep_alive.setter
-    def client_session_keep_alive(self, value: bool) -> None:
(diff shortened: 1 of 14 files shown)


python/src/snowflake/connector/connection_config.py
```diff
@@ -169,6 +165,12 @@ class ConnectionConfig(ConnectionConfigMixin):
     client_prefetch_threads: int | None = 4
     """Number of concurrent chunk prefetch threads for result set downloading. Default: 4"""
 
+    client_session_keep_alive: bool | None = None
+    """Keep the session alive with periodic heartbeat requests"""
+
+    client_session_keep_alive_heartbeat_frequency: int | None = None
+    """Heartbeat frequency in seconds (clamped to master/16..master/4)"""
```
sfc-gh-pczajka: Lets reword that, for example:
```suggestion
    """Heartbeat frequency in seconds (clamped to interval master_token_validity/16..master_token_validity/4)"""
```

python/src/snowflake/connector/connection_config.py
```diff
@@ -169,6 +165,12 @@ class ConnectionConfig(ConnectionConfigMixin):
     client_prefetch_threads: int | None = 4
     """Number of concurrent chunk prefetch threads for result set downloading. Default: 4"""
 
+    client_session_keep_alive: bool | None = None
+    """Keep the session alive with periodic heartbeat requests"""
```
sfc-gh-pczajka: What is the default behavior? Why not to use meaningful value instead of "None"?

python/src/snowflake/connector/connection_config.py
sfc-gh-pczajka: Done in e75c9165 — expanded to spell out master_token_validity.

python/src/snowflake/connector/connection_config.py
sfc-gh-pczajka: Defaulted to False in e75c9165. Rationale: the param has only two real states (heartbeat on / off) and Python's None vs False treatment was a quirk we don't need to inherit; explicit False is forwarded to the server now too.

sf_core/src/apis/database_driver_v1/heartbeat.rs
```diff
@@ -47,14 +47,34 @@ impl Drop for HeartbeatHandle {
...
+    let master = master_validity.unwrap_or(DEFAULT_MASTER_VALIDITY);
+    let max = master / 4;
+    let min = master / 16;
+
+    let interval = match user_frequency_secs {
+        None => min,
+        Some(secs) => Duration::from_secs(secs).clamp(min, max),
+    };
+
+    interval.min(MAX_HEARTBEAT_INTERVAL)
```
sfc-gh-asolarski:
```suggestion
  user_frequency_secs
      .map_or(min, |s| Duration::from_secs(s).clamp(min, max))
      .min(MAX_HEARTBEAT_INTERVAL)
```

python/src/snowflake/connector/connection.py
sfc-gh-asolarski, line 1: we removed setters there, is it intentional (because this is initialization only param?)?

If yes we should add a behavioral difference entry, as the old python driver used to expose those setters

sfc-gh-asolarski (changes requested):

sf_core/src/apis/database_driver_v1/heartbeat.rs
sfc-gh-pczajka: Done in 8376baf2.

python/src/snowflake/connector/connection.py
sfc-gh-pczajka: Documented as BehaviorDifferences #25 - they were essentially a no-op in old connector

sfc-gh-asolarski (approved):

PR #1204, opened 2026-05-15
refactor(refresh): extract generic refresh-and-retry pattern

sfc-gh-pczajka:
## Summary

Extract the shared shape of the two refresh-and-retry loops in the codebase
(`RefreshContext::execute_with_refresh` for session tokens; the per-attempt
S3 loops in `s3_transfer.rs` for STS credentials) into a new
`sf_core::refresh` module. Both existing implementations now live behind a
single generic helper.

Stacked on **#1166** because that PR's STS work was the second concrete
data point that motivated this refactor — see boler's review note there.
Doing the abstraction now (with two real implementations to extract from)
rather than before #1166 (with only one) means the API actually fits both
cases.

## Shape

* **`refresh.rs`** — new top-level module:
  * `Refresher<Resource, Err>` trait. `current()` reads the resource for
    the next attempt; `should_refresh(&Err) -> bool` decides which errors
    are recoverable; `refresh() -> Ok(bool)` rotates the resource and
    returns whether anything actually changed.
  * `execute_with_refresh<R, ...>(refresher, op)` — the shared loop. On
    error, calls `should_refresh`; if recoverable, calls `refresh()` —
    `Ok(true)` retries, `Ok(false)` propagates. Termination is
    refresher-driven, so coalescing-window and one-shot state machines
    both fit naturally.
  * Type parameters (not associated types) so `dyn Refresher<R, E>` is
    object-safe. The S3 file-transfer path needs that.

* **`RefreshContext`** (session) — now impls `Refresher<SensitiveString,
  ApiError>`. The existing `RefreshState` state machine maps directly
  onto the `Ok(bool)` return: first refresh returns `Ok(true)` and
  flips to `Refreshed`; second returns `Ok(false)`. Public method
  signatures unchanged; the imperative loop body is replaced with
  a call to `refresh::execute_with_refresh`.

* **STS path** — `s3_transfer` introduces `StageCredsAdapter<'a, E>`,
  a thin bridge from the file-manager-facing `StageCredsRefresher`
  trait (with its shared `StageCredsCache`) to `Refresher<CloudCredentials,
  S3AttemptError<E>>`. The adapter tracks the AWS key id it last handed
  out, so a refresh inside the 10-minute coalescing window reports
  `Ok(false)` and the helper propagates rather than spinning.
  `S3AttemptError<E>` keeps `StsExpired` an internal-only signal —
  `UploadFileError` / `DownloadFileError` still carry no refresh-
  internal variants.

## What didn't fit cleanly

* **No-refresher path.** When the file-transfer caller passes
  `refresher: None`, there's nothing to coalesce or retry on — we just
  do one attempt and propagate. Forcing that through the generic
  helper would require either a no-op refresher impl or an
  `unreachable!` in `current()`. Cleaner to dispatch on the `Option`
  up-front: `Some(_)` goes through the helper, `None` does a single
  call. Two lines on each call site, zero conceptual cost.

* **Session-token error mapping.** `RefreshContext::execute_with_refresh`
  takes a closure that returns `RestError` but the public method
  returns `ApiError` (the existing signature). The wrapper closure
  converts in flight: `f(token).await.map_err(|e| QuerySnafu.into_error(e))`.
  Once that conversion happens up-front, `should_refresh` inspects an
  `ApiError::Query{source: RestError::InvalidSnowflakeResponse{SessionExpired}}`
  pattern. A bit nested but it's matching the existing error shape.

## Test plan

- [x] `cargo fmt && cargo clippy -p sf_core --all-targets -- -D warnings` — clean
- [x] `cargo test -p sf_core --lib` — 804 passed (4 new tests in `refresh::tests`:
      first-success, refresh-then-succeed, refresher-declines, non-recoverable-error)
- [x] `cargo test -p sf_core --test integration_tests put_get` — passes
- [x] No behavior change for existing call sites of either refresh path

## Follow-ups

This PR doesn't touch the lower-level `refresh_token(last_error)` API or
its 5 external call sites (heartbeat, statement, result_set,
connection). They predate `execute_with_refresh` and have their own
loops that need the imperative form. Could be migrated in a future PR
once the generic helper grows a public lower-level entry point if
needed.

### DIFF
--- a/sf_core/src/file_manager/s3_transfer.rs
+++ b/sf_core/src/file_manager/s3_transfer.rs
@@ -40,8 +40,8 @@ const REQUEST_TIMEOUT_SECS: u64 = 300;
 /// via the shared cache — no return-value plumbing required.
 ///
 /// When `refresher` is `Some`, the retry loop is driven by
-/// [`crate::refresh::execute_with_refresh`] — see that module for the shared
-/// retry-shape used by both the session-token and STS refresh paths.
+/// [`crate::refresh::execute_with_refresh`] via the [`S3StsRefresher`]
+/// implementation in this module.
 pub async fn upload_to_s3_or_skip(
     prepared: PreparedUpload,
     stage_info: &StageInfo,
@@ -74,23 +74,17 @@ pub async fn upload_to_s3_or_skip(
         }
     };
 
-    let outcome = match refresher.as_deref_mut() {
-        Some(r) => {
-            let mut adapter = StageCredsAdapter::new(r, &stage_info.creds, |e| {
-                upload_file_error::StageCredsRefreshSnafu.into_error(e)
-            });
-            execute_with_refresh(&mut adapter, attempt).await
-        }
-        None => attempt(stage_info.creds.clone()).await,
-    };
-
-    outcome.map_err(|e| match e {
-        S3AttemptError::Other(err) => err,
-        // The refresher declined to rotate (or there was none). Surface the
-        // original AWS error as a normal upload failure — same shape as any
-        // other S3 PUT error.
-        S3AttemptError::StsExpired(aws_err) => upload_file_error::S3UploadSnafu.into_error(aws_err),
-    })
+    run_s3_with_sts_refresh(
+        refresher,
+        &stage_info.creds,
+        |e| upload_file_error::StageCredsRefreshSnafu.into_error(e),
+        // Refresher declined to rotate (or there was none). Surface the
+        // original AWS error as a normal upload failure — same shape as
+        // any other S3 PUT error.
+        |aws_err| upload_file_error::S3UploadSnafu.into_error(aws_err),
+        attempt,
+    )
+    .await
 }
 
 /// Internal error type for one attempt of an S3 operation. The `StsExpired`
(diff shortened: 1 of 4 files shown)


sfc-gh-pczajka:
* **#1204** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1204?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1204?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1166** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1166?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1137** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1137?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sf_core/src/refresh.rs
```diff
@@ -0,0 +1,212 @@
+//! Generic refresh-and-retry pattern shared by the session-token and S3 STS
```
sfc-gh-pczajka: It's going to be shared by other providers as well, let's make the comment more generic

sf_core/src/file_manager/s3_transfer.rs
```diff
@@ -44,75 +50,149 @@ pub async fn upload_to_s3_or_skip(
...
+/// as a `Refresher<CloudCredentials, S3AttemptError<E>>` so the generic
+/// helper drives the retry loop. `wrap_refresh_err` translates a
+/// `StageCredsRefreshError` from the refresher into the operation's `E` —
+/// this avoids a blanket `From<StageCredsRefreshError>` impl on
+/// `UploadFileError`/`DownloadFileError`, which would skip snafu's
+/// source-location stamping. Tracks the last AWS key id handed out so a
+/// refresh that doesn't actually rotate (refresher inside its coalescing
+/// window) reports `Ok(false)` and the helper propagates the original
+/// error rather than spinning.
+struct StageCredsAdapter<'a, E, W> {
```
sfc-gh-pczajka: This code looks S3-specific. Could you give it better name?

sfc-gh-boler (approved):
lgtm

PR #1205, opened 2026-05-15
SNOW-3406388, gap 9: honor useS3RegionalUrl / useRegionalUrl from stage info

sfc-gh-pczajka:
## Summary

Plumbs the GS \`useS3RegionalUrl\` and \`useRegionalUrl\` flags into the universal driver's S3 client builder so PrivateLink-to-S3 accounts and Snowpipe Streaming work — they previously silently failed because the universal driver was building an aws-sdk-s3 client pointed at the global endpoint while the stage required the regional one.

- Adds \`use_s3_regional_url: bool\` to \`file_manager::StageInfo\`.
- \`stage_info_from_response\` populates it as \`useS3RegionalUrl OR useRegionalUrl\` — mirroring Python (\`s3_storage_client.py:85-91\`), JDBC (\`StorageClientFactory.java:55-58\`), and libsnowflakeclient (\`SnowflakeS3Client.cpp:106-113\`).
- \`create_s3_client\` overrides the SDK endpoint to \`https://s3.<region>.amazonaws.com[.cn]\` when the flag is set and GS hasn't supplied an explicit \`endPoint\` (FIPS / VPCE / custom endpoint takes precedence).
- The existing \`use_regional_url\` field stays GCS-conceptual (it carries the \`me-central2\` special case which is irrelevant for S3).

The \`cn-*\` suffix logic matches \`getDomainSuffixForRegionalUrl\` in \`snowflake-jdbc\` (\`SnowflakeS3Client.java:248-250\`).

## Tests

- 4 unit tests in \`s3_transfer.rs::tests\` covering \`regional_s3_endpoint\` for default / china / case-insensitive / GovCloud regions.
- 4 round-trip tests in \`query_response.rs::tests\` covering OR semantics (only \`useS3RegionalUrl\`, only \`useRegionalUrl\`, neither, both absent).
- Updated existing \`make_stage_info\` test fixtures in \`gcs_transfer.rs\` and \`azure_transfer.rs\`, and the wiremock-based integration tests in \`tests/integration/http/{azure,gcs}_retry.rs\`.

Not asserting the AWS SDK actually uses the resolved endpoint — aws-sdk-s3 doesn't expose that on its config in a stable way; reference drivers also test at the unit level by string-equality on the constructed URL.

## Test plan

- [x] \`cargo check -p sf_core\`
- [x] \`cargo clippy -p sf_core --all-targets -- -D warnings\`
- [x] \`cargo test -p sf_core --lib\` (815 passed)
- [x] Pre-commit hooks pass (formatting, clippy, cargo check)
- [ ] Validate on a PrivateLink-backed account that \`useS3RegionalUrl: true\` actually flips PUT/GET to the regional endpoint (requires server-side fixture; out of band)

## Out of scope

This addresses Gap 9 from the S3 feature-gap analysis. Other gaps (multipart upload, parallel transfer, streaming, SSE-S3, transfer acceleration, progress callbacks, GCM encryption, etc.) are left for follow-up PRs.

### DIFF
--- a/sf_core/src/file_manager/s3_transfer.rs
+++ b/sf_core/src/file_manager/s3_transfer.rs
@@ -266,9 +266,28 @@ async fn create_s3_client(
         };
         tracing::debug!("Using Snowflake-provided S3 endpoint: {endpoint_url}");
         s3_config = s3_config.endpoint_url(endpoint_url);
+    } else if stage_info.use_s3_regional_url {
+        // PrivateLink-to-S3 / Snowpipe Streaming: the global endpoint is
+        // unreachable, only `s3.<region>.amazonaws.com[.cn]` resolves.
+        let endpoint_url = regional_s3_endpoint(&stage_info.region);
+        tracing::debug!("Using S3 regional endpoint: {endpoint_url}");
+        s3_config = s3_config.endpoint_url(endpoint_url);
     }
 
     Ok(S3Client::from_conf(s3_config.build()))
+}
+
+/// Builds the S3 regional endpoint URL for a given region. China regions
+/// (`cn-*`) use the `amazonaws.com.cn` suffix; everything else uses
+/// `amazonaws.com`. Mirrors `getDomainSuffixForRegionalUrl` in
+/// snowflake-jdbc's `SnowflakeS3Client.java:248-250`.
+fn regional_s3_endpoint(region: &str) -> String {
+    let suffix = if region.to_ascii_lowercase().starts_with("cn-") {
+        "amazonaws.com.cn"
+    } else {
+        "amazonaws.com"
+    };
+    format!("https://s3.{region}.{suffix}")
 }
 
 /// Error returned when `create_s3_client` is called with non-S3 credentials.
@@ -411,5 +430,43 @@ mod tests {
         let cfg = to_aws_timeout_config(&policy);
         assert_eq!(cfg.operation_timeout(), Some(policy.max_elapsed));
         assert_eq!(cfg.operation_attempt_timeout(), policy.per_request_timeout);
+    }
+
+    // --- Regional endpoint construction ---
+
+    #[test]
+    fn regional_s3_endpoint_default_suffix() {
+        assert_eq!(
+            regional_s3_endpoint("us-east-1"),
+            "https://s3.us-east-1.amazonaws.com"
+        );
+    }
+
+    #[test]
+    fn regional_s3_endpoint_china_suffix() {
+        assert_eq!(
+            regional_s3_endpoint("cn-north-1"),
+            "https://s3.cn-north-1.amazonaws.com.cn"
+        );
+    }
+
+    #[test]
+    fn regional_s3_endpoint_china_match_is_case_insensitive() {
+        // GS could conceivably send the region in upper case; the suffix
+        // detection must not depend on case.
+        assert_eq!(
+            regional_s3_endpoint("CN-NORTH-1"),
+            "https://s3.CN-NORTH-1.amazonaws.com.cn"
+        );
+    }
+
+    #[test]
+    fn regional_s3_endpoint_govcloud_uses_default_suffix() {
+        // GovCloud regions are still under amazonaws.com (e.g.
+        // s3.us-gov-west-1.amazonaws.com); only `cn-*` gets the .cn TLD.
+        assert_eq!(
+            regional_s3_endpoint("us-gov-west-1"),
+            "https://s3.us-gov-west-1.amazonaws.com"
+        );
     }
 }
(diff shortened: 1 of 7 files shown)


sf_core/src/file_manager/s3_transfer.rs
```diff
@@ -266,11 +266,30 @@ async fn create_s3_client(
         };
         tracing::debug!("Using Snowflake-provided S3 endpoint: {endpoint_url}");
         s3_config = s3_config.endpoint_url(endpoint_url);
+    } else if stage_info.use_s3_regional_url {
+        // PrivateLink-to-S3 / Snowpipe Streaming: the global endpoint is
+        // unreachable, only `s3.<region>.amazonaws.com[.cn]` resolves.
+        let endpoint_url = regional_s3_endpoint(&stage_info.region);
+        tracing::debug!("Using S3 regional endpoint: {endpoint_url}");
+        s3_config = s3_config.endpoint_url(endpoint_url);
     }
```
sfc-gh-boler: <p class='graphite__hidden'><i>[Re: line +275]</i></p>
This should be extracted into a helper, so that we can test for all four cases: endpoint set (regardless of flag), only flag set, neither, and endpoint without scheme gets https:// prepended. Having a helper will work around the fact that s3 Config doesn't expose the URL.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1205?utm_source=unchanged-line-comment">Graphite</a>.</p>

sf_core/src/file_manager/s3_transfer.rs
```diff
@@ -266,11 +266,30 @@ async fn create_s3_client(
         };
```
sfc-gh-boler: <p class='graphite__hidden'><i>[Re: line +260]</i></p>
Sometimes it's `end_point` and sometimes `endpoint`. Let's be consistent (I vote for the latter format).
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1205?utm_source=unchanged-line-comment">Graphite</a>.</p>

sf_core/src/file_manager/s3_transfer.rs
```diff
@@ -266,11 +266,30 @@ async fn create_s3_client(
...
+        s3_config = s3_config.endpoint_url(endpoint_url);
     }
 
     Ok(S3Client::from_conf(s3_config.build()))
 }
 
+/// Builds the S3 regional endpoint URL for a given region. China regions
+/// (`cn-*`) use the `amazonaws.com.cn` suffix; everything else uses
+/// `amazonaws.com`. Mirrors `getDomainSuffixForRegionalUrl` in
+/// snowflake-jdbc's `SnowflakeS3Client.java:248-250`.
```
sfc-gh-boler: There isn't much point in referencing line numbers, because those change all the tiem.

sf_core/src/rest/snowflake/query_response.rs
```diff
@@ -1769,4 +1777,68 @@ mod tests {
...
+    fn s3_stage_info_json(use_s3_regional: Option<bool>, use_regional: Option<bool>) -> String {
+        let s3_regional = match use_s3_regional {
+            Some(b) => format!(", \"useS3RegionalUrl\": {b}"),
+            None => String::new(),
+        };
+        let regional = match use_regional {
+            Some(b) => format!(", \"useRegionalUrl\": {b}"),
+            None => String::new(),
+        };
+        format!(
```
sfc-gh-boler: Let's not construct JSON objects using string contact. Use `serde_json::json!`.

sfc-gh-boler (approved):
Looks good, just a few minor issues to address before merging.

sf_core/src/file_manager/s3_transfer.rs
sfc-gh-pczajka: Renamed `end_point` → `endpoint` across `StageInfo`, the transfer modules, query_response.rs, and the integration tests in 5ae63402.

sf_core/src/file_manager/s3_transfer.rs
sfc-gh-pczajka: Extracted `resolve_s3_endpoint(stage_info) -> Option<String>` and added 5 unit tests covering: explicit endpoint wins over the regional flag, only flag set, neither (returns `None` so the SDK can apply its own region resolution), scheme-less endpoint gets `https://` prepended, and `http://` is preserved. 5ae63402.

sf_core/src/file_manager/s3_transfer.rs
sfc-gh-pczajka: Dropped the `:248-250` line reference and pruned similar line numbers from the docstrings added in this PR (and the chained PR #1212). 5ae63402.

sf_core/src/rest/snowflake/query_response.rs
sfc-gh-pczajka: Switched `s3_stage_info_value` to `serde_json::json!` and `serde_json::from_value`. 5ae63402.

sfc-gh-boler (approved):
Nice!

Can you add a SNOW reference to the PR?

PR #1207, opened 2026-05-18
SNOW-3548006: Use ARM64EC for Windows

sfc-gh-pbulawa:
### TL;DR

Switch the Windows ARM64 ODBC driver build from native `aarch64-pc-windows-msvc` to the ARM64EC ABI (`arm64ec-pc-windows-msvc`).

### What changed?

- The Windows ARM build target is now `arm64ec-pc-windows-msvc` instead of `aarch64-pc-windows-msvc`. The artifact, cache key, vcpkg triplet, and msvc_arch fields are all updated to reflect `arm64ec` naming throughout CI configuration and test matrix mappings.
- `setup-msvc/action.yml` now translates an `arch` input of `arm64ec` to `arm64` before invoking `vcvarsall.bat`, since `vcvarsall.bat` has no `arm64ec` argument — ARM64EC is a compile/link-time ABI selection on top of the ARM64 toolchain.
- The `rustup target add` step in the MSI build workflow is now gated on `matrix.cargo_target && matrix.arch != 'x64'` rather than only triggering for x86, so any cross-target row (including ARM64EC) automatically installs its required Rust target.
- The vcpkg install step in the ODBC test workflow now accepts both `arm64` and `arm64ec` as msvc_arch values that require the ARM64 MSVC toolchain environment.
- The `package.ps1` example updated to reference the `arm64ec-pc-windows-msvc` target directory.

### How to test?

Run the ODBC build and test workflows targeting the `windows-11-arm` runner. Verify the produced MSI contains an ARM64EC driver DLL (`sfodbc.dll`) and that the build completes without `vcvarsall.bat` or Rust target errors. The existing matrix generation tests (`test_generate_matrix.py`) cover the updated mappings and can be run locally with `pytest`.

### Why make this change?

ARM64EC ("Emulation Compatible") is the correct ABI for shipping an ARM64 Windows driver that interoperates with x64 processes and the Windows ODBC Driver Manager. The native `aarch64-pc-windows-msvc` target produces a pure ARM64 binary that cannot be loaded by x64 host processes, whereas ARM64EC allows seamless interop. Since `vcvarsall.bat` does not accept `arm64ec` as an argument, the translation layer in `setup-msvc` keeps callers able to express intent as `arm64ec` end-to-end without breaking the toolchain setup step.


sfc-gh-pbulawa:
* **#1207** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1207?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1207?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #844, opened 2026-04-14
Add DATE and TIME to C_CHAR, C_BINARY, C_TIMESTAMP conversions and tests

sfc-gh-pbulawa:

PR #1211, opened 2026-05-18
SNOW-3098562: Preserve fractional seconds in TIME → SQL_C_TYPE_TIMESTAMP

sfc-gh-rkowalski:
## Summary
- TIME → `SQL_C_TYPE_TIMESTAMP` was hardcoding `fraction: 0` and emitting a `NumericValueTruncated` warning whenever the input had sub-second precision.
- `sql::Timestamp.fraction` is a `UInteger` that holds nanoseconds per the ODBC spec, so no truncation was actually needed — propagate `snowflake_value.nanosecond()` into the struct and drop the warning.
- Surfaced in dtm-server SXS as `[3706] DTM3103: Numeric value truncated. Result type not handled.` for queries like `select time '12:56:10.513'` (expected `12:56:10.513000`).

Introduced in #844; not previously covered by tests — `time_tests.rs` had no `TYPE_TIMESTAMP` write case. Sibling `timestamp.rs::to_sql_timestamp` already does the right thing.

## Test plan
- [x] `cargo test -p odbc --lib conversion::time_tests::` — 32 passed including new regression test
- [x] `cargo test -p odbc --lib conversion::` — 1065 passed, no regressions
- [ ] Re-run dtm-server `tdc-dtm-snowflake azure.sql` SXS to confirm the diff clears

### DIFF
--- a/odbc/src/conversion/time.rs
+++ b/odbc/src/conversion/time.rs
@@ -200,14 +200,10 @@ impl WriteODBCType for SnowflakeTime {
                     hour: snowflake_value.hour() as u16,
                     minute: snowflake_value.minute() as u16,
                     second: snowflake_value.second() as u16,
-                    fraction: 0,
+                    fraction: snowflake_value.nanosecond(),
                 };
                 binding.write_fixed(ts);
-                if snowflake_value.nanosecond() != 0 {
-                    Ok(vec![Warning::NumericValueTruncated])
-                } else {
-                    Ok(vec![])
-                }
+                Ok(vec![])
             }
             _ => UnsupportedOdbcTypeSnafu {
                 target_type: binding.target_type,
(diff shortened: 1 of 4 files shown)


sfc-gh-ddas (approved):

sfc-gh-rkowalski:
Closing in favor of fixing the dtm-server test reference instead.

BD#42 in odbc_tests/BehaviorDifferences.yaml explicitly documents that the new driver should follow the ODBC spec — `SQL_TYPE_TIME → SQL_C_TYPE_TIMESTAMP` zeroes the fraction field and emits SQLSTATE 01S07 — and that the old driver's fraction-preserving behavior is the bug. dtm-server should adapt its reference output to the spec-compliant behavior rather than have the universal-driver roll back to match the old driver.

PR #1212, opened 2026-05-18
feat(s3): support enable_stage_s3_privatelink session parameter

sfc-gh-pczajka:
## Summary

Follow-up to #1205. Adds the connection-level / session-parameter override to the regional-S3-URL feature. ORs into `StageInfo.use_s3_regional_url`, matching Python / JDBC / libsnowflakeclient.

## Changes

- sf_core: new `use_s3_regional_url` ParamDef with `ENABLE_STAGE_S3_PRIVATELINK_FOR_US_EAST_1` as alias; PUT/GET dispatch now reads the resolved session-param value and ORs it into stage info.
- Python: `enable_stage_s3_privatelink_for_us_east_1` deprecated alias for canonical `use_s3_regional_url`; BehaviorDifferences #27.

## Tests

- Rust unit: OR semantics, case-insensitive value, one-directional override.
- Python unit: deprecated alias maps + warns; canonical is silent.
- Python e2e: round-trip with kwarg set; deprecation fires on connect.

### DIFF
--- a/python/tests/e2e/put_get/test_put_get_s3_privatelink.py
+++ b/python/tests/e2e/put_get/test_put_get_s3_privatelink.py
@@ -1,0 +1,78 @@
+"""
+End-to-end coverage for the `use_s3_regional_url` kwarg and its
+deprecated alias `enable_stage_s3_privatelink_for_us_east_1`.
+
+We can't assert that the regional S3 endpoint is actually used — that
+would require a PrivateLink-enabled test account — but we can verify
+that:
+  * the kwarg is accepted by the public `connect()` API and a basic PUT
+    round-trips successfully (run against both drivers — the universal
+    driver receives the canonical name, the reference connector receives
+    the legacy name, since each is the only spelling that driver accepts);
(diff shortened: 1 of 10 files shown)


python/tests/e2e/put_get/test_put_get_s3_privatelink.py
```diff
@@ -0,0 +1,67 @@
...
+would require a PrivateLink-enabled test account — but we can verify
+that:
+  * the canonical kwarg is accepted by the public `connect()` API and
+    a basic PUT round-trips successfully;
+  * the legacy kwarg emits a `DeprecationWarning` at connection time.
+
+The OR-with-stage-info-flags logic is covered by Rust unit tests in
+`sf_core/src/rest/snowflake/query_response.rs::tests`, and the
+kwarg-rewrite logic by `tests/unit/test_connection_config.py`.
+"""
```
sfc-gh-pczajka, line 19: Could we add an int test which wiremocks this request?

sfc-gh-turbaszek (approved):

PR #1213, opened 2026-05-18
SNOW-2872505: make async query polling unbounded by default

sfc-gh-asolarski:
## Summary
- Decouple async query polling from the per-HTTP-call retry budget — `wait_for_completion` no longer enforces `policy.max_elapsed` as a total wall-clock cap
- Individual `poll_query_status` HTTP calls remain bounded by the retry policy (120s default)
- Introduce `STATEMENT_POLL_DEADLINE` constant (`None` = unbounded) as a named hook for a future caller-supplied statement timeout

## Context
Queries that take longer than the retry policy's `max_elapsed` (120s) were incorrectly failing with `DeadlineExceeded`. The retry budget should only apply to individual HTTP poll attempts, not the total query execution time. This separates the two concerns and leaves a clear extension point for a configurable statement-level timeout in the future.

## Test plan
- `cargo check -p sf_core` passes
- Pre-commit hooks (formatting, clippy, cargo check) pass
- With `STATEMENT_POLL_DEADLINE = None`, the deadline checks compile away — zero runtime cost and identical behavior to an infinite poll loop
- Existing retry behavior for individual HTTP calls unchanged (bounded by `policy.max_elapsed`)

### DIFF
--- a/sf_core/src/rest/snowflake/async_exec.rs
+++ b/sf_core/src/rest/snowflake/async_exec.rs
@@ -19,6 +19,13 @@ const INLINE_SHORT_POLL_DELAYS: &[Duration] = &[
 ];
 const QUERY_SEQUENCE_ID: u64 = 1;
 
+/// Maximum wall-clock time the driver will spend polling for a single
+/// statement's completion. `None` means unbounded — poll until Snowflake
+/// returns data or a terminal error.
+///
+/// TODO: plumb a caller-supplied `statement_timeout` (e.g. configuration) so users can cap long-running queries.
+const STATEMENT_POLL_DEADLINE: Option<Duration> = None;
+
 /// Metrics for async query execution phases, logged for monitoring and debugging.
 ///
 /// Async execution follows this flow:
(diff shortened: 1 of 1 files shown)


sfc-gh-asolarski:
* **#1213** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1213?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1213?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sf_core/src/rest/snowflake/async_exec.rs
```diff
@@ -19,6 +19,13 @@ const INLINE_SHORT_POLL_DELAYS: &[Duration] = &[
 ];
 const QUERY_SEQUENCE_ID: u64 = 1;
 
+/// Maximum wall-clock time the driver will spend polling for a single
+/// statement's completion. `None` means unbounded — poll until Snowflake
+/// returns data or a terminal error.
+///
+/// TODO: plumb a caller-supplied `statement_timeout` (e.g. configuration) so users can cap long-running queries.
+const STATEMENT_POLL_DEADLINE: Option<Duration> = None;
```
sfc-gh-jszczerbinski, line 27: So this is always None? We want this to be configurable in the future? We don't want to convert it into static?

sfc-gh-asolarski: my assumption is that we want to make it configurable in the next steps (as it's configurable at least in some of the old drivers, e.g. queryTimeout on statement in jdbc)

sfc-gh-jszczerbinski (approved):

PR #1226, opened 2026-05-18
SNOW-3487070: python - async-first DatabaseDriverClient with blocking wrapper

sfc-gh-asolarski:
## Summary

- Convert `ProtoTransport.handle_message` to `async def`; the existing sync ctypes FFI is bridged via `asyncio.to_thread` so the Rust core call still runs off the event-loop thread.
- Make the generated `DatabaseDriverClient` async-native — every RPC method is now `async def` and `await`s the transport.
- Generate a new `DatabaseDriverBlockingClient` sync wrapper. Each blocking method submits the corresponding async coroutine to a lazily-started, process-wide background event loop via `asyncio.run_coroutine_threadsafe(...).result()`. The loop runs on a daemon thread.
- Wire `CoreDriver._client` to a `DatabaseDriverBlockingClient` wrapping `DatabaseDriverClient(ProtoTransport(), error_handler=_proto_to_public_error)`. Every `CoreDriver` method body is unchanged — they keep dispatching synchronously through the blocking wrapper, so no caller (connection, cursor, telemetry, config_manager, …) needs any modification.

## Context

First subtask of [SNOW-3487055](https://snowflakecomputing.atlassian.net/browse/SNOW-3487055) (asyncio epic). The goal is to make the RPC layer async-first while keeping the entire existing sync codebase working as-is. Subsequent subtasks (`AsyncSnowflakeCursorBase`, `AsyncConnection`, callback-based async FFI) build on this foundation.

Why a background-thread loop instead of literal `loop.run_until_complete` (the ticket says "or equivalent"): `run_until_complete` deadlocks if the calling thread already has a running loop, fails under concurrent calls from multiple threads, and doesn't tick background tasks between calls. A persistent loop on a dedicated thread + `run_coroutine_threadsafe` is the stdlib-blessed pattern for cross-thread submission and is what asgiref / httpx / boto3 use under the hood.

## ⚠️ What this PR is NOT (read before reviewing)

**The transport is async in name only.** `asyncio.to_thread` does not make `sf_core_api_call_proto` non-blocking — it relocates the blocking call onto a worker thread so the event loop isn't blocked. The Rust call still pins a thread for its full duration. Concrete consequences callers should be aware of:

1. **Concurrency is bounded by the loop's default `ThreadPoolExecutor`** (`min(32, os.cpu_count() + 4)` workers). Beyond that, concurrent `await client.foo()` calls queue silently. We're also squatting on the *shared* default executor — anything else in the process using `asyncio.to_thread` competes for the same pool.
2. **Cancellation is a lie.** If the awaiting coroutine is cancelled (`asyncio.wait_for` timeout, `TaskGroup` shutdown), `asyncio.to_thread` does not interrupt the in-flight ctypes call — the worker thread keeps running until the C function returns. The `Future` is cancelled, the FFI work is not.
3. **GIL still serializes Python wrapper code.** Only the ctypes call itself releases the GIL. Buffer allocation, byte conversions, and proto parsing in `_handle_message_sync` are GIL-bound regardless of how many coroutines are "concurrent."
4. **Three-hop overhead from sync callers.** sync caller → `BlockingClient` → `run_coroutine_threadsafe` to background loop thread → `to_thread` to executor thread → ctypes. Tens of µs of pure overhead per RPC versus the previous direct call. No benefit until the FFI itself becomes async.
5. **No backpressure / no concurrency limits.** The previous sync code was implicitly throttled by its callers; the async surface exposes the FFI to unbounded fan-out (`gather`, `TaskGroup`) with no coordination with the Rust core's own concurrency model.
6. **Default executor + interpreter shutdown** can produce shutdown warnings if calls are in flight at process exit.

**The ticket explicitly accepts this**: *"Initial implementation uses `asyncio.to_thread` to bridge existing sync FFI."* This PR is bridge code — it lands the async surface so higher layers (`AsyncSnowflakeCursorBase`, `AsyncConnection`, `AsyncResultBatch`) can be built on top, then the transport gets replaced.

### Will SNOW-3487102 fix this?

[SNOW-3487102 (Callback-based async FFI boundary)](https://snowflakecomputing.atlassian.net/browse/SNOW-3487102) is the dedicated follow-up. Mapping the items above to its acceptance criteria:

| # | Issue                              | Fixed by SNOW-3487102?                                                                |
| - | ---------------------------------- | ------------------------------------------------------------------------------------- |
| 1 | Executor-bounded concurrency       | ✅ "Multiple concurrent async operations run without spawning threads"                |
| 2 | Cancellation propagation           | ⚠️ Not explicit in ACs — depends on whether the Rust callback design exposes cancellation tokens. Worth raising on that ticket. |
| 3 | GIL-serialized wrapper             | ✅ Indirect — no thread hops means less Python plumbing per call                     |
| 4 | Three-hop overhead                 | ✅ "Benchmark showing reduced overhead vs thread-based bridge"                       |
| 5 | No backpressure                    | ✅ Native dispatch lets us coordinate with Rust core's concurrency model            |
| 6 | Default-executor shutdown warnings | ✅ "asyncio.to_thread / run_in_executor removed from transport"                     |

So 5 of 6 are addressed by design; cancellation needs an explicit call-out on SNOW-3487102.

## Test plan

- [x] `cargo build -p sf_core`
- [x] `hatch build` (regenerates `database_driver_v1_services.py` cleanly)
- [x] `hatch run dev.py3.13:unit` → **1014 passed, 1 skipped, 0 failed** (zero test code changes)
- [x] `hatch run dev.py3.13:python -m pytest tests/test_exceptions.py` → 46 passed
- [x] Smoke-tested async (`await client.database_new(...)`) + blocking (`blocking.database_new(...)`) paths against a fake async transport; verified `core_driver.client` returns a `DatabaseDriverBlockingClient` whose underlying chain is `→ DatabaseDriverClient → ProtoTransport`.
- [ ] CI green

[SNOW-3487055]: https://snowflakecomputing.atlassian.net/browse/SNOW-3487055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/proto_generator/src/generators/python_generator.rs
+++ b/proto_generator/src/generators/python_generator.rs
@@ -340,6 +340,8 @@ class {service_name}BlockingClient:
         self._loop = loop
 
     def _run(self, coro):
+        if not self._loop.is_running():
+            raise RuntimeError("Background event loop is not running")
         return asyncio.run_coroutine_threadsafe(coro, self._loop).result()
 
 "#
--- a/proto_generator/src/generators/python_generator.rs
+++ b/proto_generator/src/generators/python_generator.rs
@@ -114,8 +114,13 @@ impl PythonGenerator {
             }
 
             // Generate client classes
-            for service in file.service {
+            for service in file.service.clone() {
                 content += &self.generate_client_class(&service, &package);
+            }
+
+            // Generate blocking client wrappers
+            for service in file.service {
+                content += &self.generate_blocking_client_class(&service, &package);
             }
 
             result.add_file(
@@ -128,7 +133,8 @@ impl PythonGenerator {
     }
 
     fn generate_common_imports(&self) -> String {
-        r#"from abc import ABC, abstractmethod
+        r#"import asyncio
+from abc import ABC, abstractmethod
 from typing import Optional
 from google.protobuf import message
 from .database_driver_v1_pb2 import *
@@ -291,8 +297,8 @@ class ProtoError(Exception):
             let error_type = method_error.or(service_error).unwrap_or(&empty_error);
 
             content += &format!(
-                r#"    def {name}(self, request: {input_type}) -> {output_type}:
-        (code, response_bytes) = self._transport.handle_message('{service_name}', '{name}', request.SerializeToString())
+                r#"    async def {name}(self, request: {input_type}) -> {output_type}:
+        (code, response_bytes) = await self._transport.handle_message('{service_name}', '{name}', request.SerializeToString())
         if code == 0:
             response = {output_type}()
             response.ParseFromString(response_bytes)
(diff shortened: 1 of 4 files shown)


sfc-gh-asolarski:
* **#1226** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1226?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1231](https://github.com/snowflakedb/universal-driver/pull/1231) <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1231?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1240](https://github.com/snowflakedb/universal-driver/pull/1240) <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1240?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1226?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1231, opened 2026-05-19
SNOW-3487102: python - callback-based async FFI boundary

sfc-gh-asolarski:
## Summary

- Add `sf_core_api_call_proto_async` in Rust (`sf_core/src/protobuf/c_api.rs`).
  Spawns the work on the existing tokio runtime and drives it with a real
  `.await` (no `block_on`, no thread-per-call). When the request finishes a C
  callback is invoked from a tokio worker thread with
  `(user_data, status, response_ptr, response_len)`. Panics in the underlying
  transport are caught via `futures::FutureExt::catch_unwind` and surfaced as
  status-2 transport errors instead of aborting the runtime.
- Add `sf_core_cancel_request(request_id: u64)` for cooperative cancellation:
  every async request gets a monotonic ID and registers an `AbortHandle` in a
  `Mutex<BTreeMap>` registry; cancel removes + aborts. The spawned task waits
  on a `tokio::sync::Notify` until the AbortHandle is registered, so a
  fast-completing task can never deregister itself before the caller registers
  it (avoids a stale-handle leak in the registry).
- Replace the `asyncio.to_thread` bridge in `ProtoTransport.handle_message`
  with a true callback-based dispatch: create an `asyncio.Future`, build a C
  callback that resolves the Future via `loop.call_soon_threadsafe`, submit
  via `sf_core_api_call_proto_async`, await the Future. On
  `asyncio.CancelledError`, call `sf_core_cancel_request(request_id)` before
  re-raising — cancellation now genuinely propagates to Rust.
- Lifetime safety: pin the `RESPONSE_CALLBACK` ctypes object to the Future
  via `future._proto_transport_callback_ref`. Without this, an awaiting
  coroutine that gets cancelled would drop the only Python-side reference
  while Rust still holds the function pointer (use-after-free / segfault).
  The closure also `future.done()`-guards `set_result`, so a callback that
  fires after the Future is cancelled is a safe no-op.
- Use `ctypes.string_at` for the response copy (single memcpy) instead of
  the O(n) Python `__getitem__` slice path.
- Empty-message safety: `sf_core_api_call_proto_async` skips
  `slice::from_raw_parts(ptr, 0)` when `request_len == 0`, since ctypes may
  pass a null pointer for a zero-length buffer and `from_raw_parts` requires
  non-null even at length 0. `DatabaseNewRequest()` serializes to `b""`, so
  this code path runs in production.

## Tests

`python/tests/unit/test_callback_ffi.py` — 13 unit tests that mock at the
FFI boundary so they run without the dylib loaded:

- `TestSingleCall` — status 0/1/2 paths, unknown status raises, request
  bytes / api / method are passed through unchanged.
- `TestCancellationSafety` — callback firing into a cancelled Future is a
  no-op; cancel-before-submit works.
- `TestCancellationPropagation` — cancelling the awaiting task calls
  `sf_core_cancel_request` exactly once; normal completion does not call
  cancel; a late callback racing past the abort point does not double-cancel.
- `TestLifetimeSafety` — directly asserts
  `future._proto_transport_callback_ref is callback` (mutation-tested:
  forcing the pin to `None` causes the test to fail).
- `TestConcurrency` — 25 in-flight requests, completed in reverse order,
  each resolves to its own response.
- `TestEmptyMessage` — regression for the `b""` case.

`python/tests/unit/conftest.py` (new) — hermetic `connection` fixture for
unit tests. Overrides `tests/conftest.py`'s module-scoped fixture that
attempts a real Snowflake connection (and hangs forever in unit-test mode
without `parameters.json`). This was the **actual** root cause of the
`test_connect_returns_connection` timeout from the previous PR; the
`@pytest.mark.skip` was a cover-up. The skip is removed and the test now
passes via the mocked Connection.

## Why callback-based instead of `asyncio.to_thread`

The `asyncio.to_thread` bridge from SNOW-3487070 had three real costs:

- Concurrency capped at the loop's default `ThreadPoolExecutor`
  (`min(32, os.cpu_count() + 4)` workers); above that, calls queue silently
  and contend with anything else in the process using `to_thread`.
- A pointless three-thread hop on every sync call: caller → loop thread →
  executor thread → ctypes.
- Hidden coupling to the default executor's shutdown lifecycle.

With this PR the transport is genuinely non-blocking up to whatever
concurrency the tokio runtime provides; no Python thread is spawned per call.

## What this PR is **not**

- Cancellation is now propagated to Rust, but it is still **best-effort**:
  tokio aborts the task at its next await point. If the task is past its
  last await (mid-callback dispatch) the callback can still fire — Python's
  `future.done()` guard makes that case a safe no-op.
- The tokio runtime still has `worker_threads(1)`. That's fine for true
  async I/O but limits concurrency for any sync work the transport might do
  internally; tunable separately.

## Test plan

- [x] `cargo build -p sf_core`
- [x] `hatch build` regenerates `database_driver_v1_services.py` cleanly
- [x] `pytest tests/unit/test_callback_ffi.py` → **13 passed**
- [x] `hatch run dev.py3.13:unit` → **1027 passed, 1 skipped, 0 errors, ~12 s**
      (was 1013 passed / 2 skipped / 1 timeout under the previous bridge)
- [x] Smoke test against real `libsf_core`: 1 single call + 50 concurrent
      + 1 cancellation — all green; `threading.active_count()` stays at 2
      throughout
- [x] Mutation test: forcing `future._proto_transport_callback_ref = None`
      makes `TestLifetimeSafety` fail with the expected message
- [x] `ruff format` and `ruff check` clean
- [ ] CI green

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/python/src/snowflake/connector/_internal/api_client/c_api.py
+++ b/python/src/snowflake/connector/_internal/api_client/c_api.py
@@ -63,6 +63,7 @@ LOGGER_CALLBACK = ctypes.CFUNCTYPE(
 core.sf_core_init.argtypes = [LOGGER_CALLBACK]
 core.sf_core_init.restype = ctypes.c_uint32
 
+# Sync proto API (blocks until complete)
 core.sf_core_api_call_proto.restype = ctypes.c_uint32
 core.sf_core_api_call_proto.argtypes = [
     ctypes.c_char_p,  # const char* api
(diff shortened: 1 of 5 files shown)


sfc-gh-asolarski:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1231?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#1231** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1231?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1231?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1226** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1226?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1240](https://github.com/snowflakedb/universal-driver/pull/1240) <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1240?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

PR #1107, opened 2026-05-11
SNOW-3520403: odbc: iODBC driver-manager support (CI matrix, test infra, SKIP_IODBC)

sfc-gh-pbulawa:
### TL;DR

Wire iODBC into the ODBC driver's CI by adding a `macos-arm-aws-iodbc` PR matrix cell, the C++ test-harness compatibility shims (`SKIP_IODBC`, ODBC 3.8 fallbacks, `ODBCINSTINI`-based driver registration), and ~141 `SKIP_IODBC` annotations on tests that fail on iODBC for reasons unrelated to the driver itself (ODBC 3.8 incompatibilities, SQLSTATE differences, UTF-16 hardcoding in the test scaffolding).

Top of a 3-PR stack:
- #1238: `SfOdbcIni` config refactor (base).
- #1239: UTF-32 `SQLWCHAR` ABI support — the change that makes iODBC functional at runtime.
- **#1107 (this PR)**: iODBC CI cell + test-harness infrastructure that exercises #1238 + #1239 end-to-end.

### What changed?

**CI matrix (`ci/test_matrix/models/odbc.py`, `ci/test_matrix/generate_matrix.py`)**
- Add a `DM` (driver-manager) axis with `unixodbc` (default) and `iodbc`. Constraint: `iodbc` is only valid on macOS.
- Add a `macos-arm-aws-iodbc` PR cell so iODBC is exercised on every PR.
- The matrix generator emits `driver_manager` in the row only when it differs from the default, and includes `iodbc` in the matrix-cell name so the CI run is unambiguous.

**Reusable workflow (`.github/workflows/_test-odbc.yml`)**
- `brew install libiodbc` alongside `unixodbc` so both DMs are available in the same job image; `ccache` and `test-deps` cache keys include `${matrix.driver_manager || 'unixodbc'}` to keep iODBC and unixODBC caches disjoint.
- Export `DRIVER_MANAGER` to the test runner so the build picks the right DM headers and so the harness writes the right `sf.odbc.ini`.

**Test runner (`scripts/odbc/run_tests_unix.sh`)**
- Branch on `$DRIVER_MANAGER`: select iODBC vs. unixODBC `odbc_config`, link against the matching library, and write a temporary `sf.odbc.ini` with `DriverManagerEncoding=UTF-32` when iODBC is selected.

**C++ harness compatibility (`odbc_tests/common/...`)**
[...]

PR #1238, opened 2026-05-19
SNOW-3520403: config: extract `SfOdbcIni` loader and refactor logging config wiring

sfc-gh-pbulawa:
### TL;DR

Centralise `sf.odbc.ini` parsing in a new `sf_core::config::sf_odbc_ini::SfOdbcIni` snapshot so the logging subsystem and the ODBC wide-string ABI read the file once, from one place, and see identical values. Document the file's location, security requirements, and the supported logging keys in the ODBC driver README.

This is the base of a 3-PR stack:
- **#1238 (this PR)**: config refactor — `SfOdbcIni` loader + README documentation.
- **#1239**: UTF-32 `SQLWCHAR` support consuming `SfOdbcIni::raw_value` for `DriverManagerEncoding`.
- **#1107**: iODBC driver-manager support — CI matrix cell, test infrastructure, `SKIP_IODBC` annotations.

### What changed?

**`sf_core::config::sf_odbc_ini::SfOdbcIni`** (new)
- Process-wide snapshot loaded lazily via `OnceLock` on first call to `SfOdbcIni::global()`.
- Locates the INI file in the documented order (`$SF_ODBC_INI` → `<config_dir>/snowflake/sf.odbc.ini` → `~/.snowflake/sf.odbc.ini`), enforces `chmod 600` on Unix, and falls back to defaults with an `eprintln!` warning on any I/O or parse failure (the `LogManager` is not yet wired up at this point).
- Routes logging keys to `LoggingConfig` and stores everything else in a case-insensitive `raw_values` map so foreign subsystems (e.g. the UTF-32 ABI follow-up) can look up keys without re-parsing the file.

**`sf_core::logging::ini_config`**
- Extract `apply_logging_key(key, value, &mut LoggingConfig) -> Result<bool, LogError>` as the canonical per-key primitive. Returns `Ok(true)` when the key is owned by logging, `Ok(false)` when it belongs to another subsystem.
- `apply_ini_section` (used by `parse_ini_*` helpers and tests) now delegates to `apply_logging_key`, eliminating the duplicate match between the standalone parser and the new snapshot.

**`sf_core::logging::log_manager`**
- `LogManager::for_odbc()` now reads its `LoggingConfig` from `SfOdbcIni::global().logging().clone()` instead of opening the INI file directly. This guarantees the logging subsystem and any future readers see the same snapshot.
- `LoggingConfig` derives `Clone` to support this.

**`odbc/README.md`**
- New `Configuration via sf.odbc.ini` section: file lookup order, Unix `chmod 600` requirement, and a table of the recognised logging keys (`LogLevel`, `LogPath`, `LogFile`, `LogMaxSize`, `LogMaxCount`, `LogRotation`, `LogEnabled`, `LogQueryText`, `LogQueryParameters`, `ErrorTraceEnabled`).
- The next PR in the stack appends a `DriverManagerEncoding` subsection.

### How to test?

- `cargo test -p sf_core --lib config::sf_odbc_ini::` — 9 unit tests covering the dispatch (logging vs. raw), case-insensitive lookups, missing-file/missing-perms fallback, and a real on-disk round-trip with `chmod 600`.
- `cargo test -p sf_core --lib logging::` — existing 50-test suite passes unchanged (the refactor is behaviour-preserving).
- `cargo test -p sf_core --test logging_for_odbc_tests` — end-to-end test that writes a temp INI, sets `$SF_ODBC_INI`, calls `LogManager::for_odbc()`, and verifies an event lands in the log file. Now also exercises the `SfOdbcIni` singleton's load path.

### Why make this change?

The ODBC driver needs `sf.odbc.ini` for two independent subsystems: logging (which has always read it) and the upcoming UTF-32 `SQLWCHAR` ABI (which needs `DriverManagerEncoding`). Reading the file in two places would either drift out of sync or require both subsystems to know about each other's keys. Funnelling everything through a single `SfOdbcIni` snapshot keeps the two consumers loosely coupled while ensuring they observe identical configuration.

The README change makes the file's location, security requirements, and supported keys discoverable without having to read the Rust source.

### DIFF
--- a/sf_core/src/config/sf_odbc_ini.rs
+++ b/sf_core/src/config/sf_odbc_ini.rs
@@ -1,0 +1,278 @@
+//! One-shot reader for `sf.odbc.ini`.
+//!
+//! The instance is created lazily by [`SfOdbcIni::global`] on first
+//! access. The driver's bootstrap calls it during environment allocation
+//! to seed the [`LogManager`]; the ODBC wrappers call it on the first
+//! wide-string operation. Both observers see the same snapshot.
+
+use std::collections::HashMap;
+use std::path::{Path, PathBuf};
+use std::sync::OnceLock;
+
+use crate::logging::LoggingConfig;
+use crate::logging::ini_config;
+
+/// Process-wide snapshot of `sf.odbc.ini`.
+///
+/// Construction is fallible-by-default: any I/O or parse error in the load
+/// path is surfaced via `eprintln!` (the [`LogManager`] is not yet wired up
+/// when the snapshot is first built) and the field defaults are used so
+/// the driver continues to function. The `path` field records which file,
+/// if any, was actually read so diagnostics elsewhere can reference it.
+///
+/// [`LogManager`]: crate::logging::LogManager
+pub struct SfOdbcIni {
+    /// Path that was actually read, if any. Kept so future diagnostics can
+    /// answer "where did this config come from?"; currently surfaced only
+    /// to tests via [`SfOdbcIni::path`].
+    #[allow(dead_code)]
+    path: Option<PathBuf>,
+    logging: LoggingConfig,
+    /// Lowercased keys that are not part of the logging namespace. Other
+    /// subsystems look these up via [`SfOdbcIni::raw_value`].
+    /// Keeping the values untyped lets `sf_core` host the singleton
+    /// without taking a dependency on every consumer's types.
+    raw_values: HashMap<String, String>,
+}
+
+impl SfOdbcIni {
+    /// Returns the process-global snapshot, loading and caching it on
+    /// first access. Subsequent calls return the same instance.
+    pub fn global() -> &'static SfOdbcIni {
+        static INSTANCE: OnceLock<SfOdbcIni> = OnceLock::new();
+        INSTANCE.get_or_init(Self::load)
+    }
+
+    /// Locate, permission-check, and parse `sf.odbc.ini`. Any failure
+    /// falls back to defaults and is reported on `stderr` (we run before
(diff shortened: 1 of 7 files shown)


sfc-gh-pbulawa:
* **#1238** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1238?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1238?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sf_core/src/config/sf_odbc_ini.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// Keeping the values untyped lets `sf_core` host the singleton
+    /// without taking a dependency on every consumer's types.
+    raw_values: HashMap<String, String>,
+}
+
+impl SfOdbcIni {
+    /// Returns the process-global snapshot, loading and caching it on
+    /// first access. Subsequent calls return the same instance.
+    pub fn global() -> &'static SfOdbcIni {
+        static INSTANCE: OnceLock<SfOdbcIni> = OnceLock::new();
```
sfc-gh-jszczerbinski, line 42: No statics please

sfc-gh-jszczerbinski (changes requested):
Please do not use statics: I think this refactor is based on static mainly - you need to rethink the approach. Also the refactor added around 372 lines and deleted 27 without changing the behavior...

sfc-gh-pbulawa:
This is the new PR: https://app.graphite.com/github/pr/snowflakedb/universal-driver/1260

PR #1250, opened 2026-05-19
NO-SNOW: add code-review design discipline rule and end-of-task self-review skill for Cursor + Claude Code

sfc-gh-boler:
Distill recurring review findings into two complementary artifacts for Cursor and Claude Code, with proper sync infrastructure to prevent drift.

## What changed

**Rule: code-review design discipline** (`alwaysApply`)

- `.claude/rules/code-review-design-discipline.md` — canonical source; full content lives here.
- `.cursor/rules/code-review-design-discipline.mdc` — derived copy with Cursor frontmatter; body is byte-identical to the Claude file.

Both files carry full content (not a pointer). `alwaysApply` rules are injected into the agent system prompt at session start. A pointer file would load the body via tool call, placing it in conversation history where context compaction can silently drop it mid-session. Full content in both files keeps the rule in context for the lifetime of any session.

The code example in section 1 was restructured per review: DI (threading from the call site) is now the primary `✅` example; cfg-based compile-time substitution is demoted to a clearly labeled `⚠️ Workaround` with an explanation of its limitations.

**Skill: end-of-task self-review** (on-demand)

- `.claude/skills/end-of-task-self-review/SKILL.md` — canonical source; full content lives here.
- `.cursor/skills/end-of-task-self-review/SKILL.md` — thin pointer that tells the agent to read the canonical file.

Skills fire on demand. The agent reads the canonical file fresh at invocation time, so a pointer is safe and avoids duplicating ~180 lines of content.

**Sync infrastructure** (`scripts/check-ai-rules-sync.sh` + pre-commit hook `ai-rules-sync`)

Two checks run automatically on every commit that touches `.claude/` or `.cursor/` AI files:

1. **Rule sync** — `.cursor/rules/*.mdc` body must be byte-identical to `.claude/rules/*.md`. Catches drift before it reaches the remote.
2. **Skill mirror** — every `.claude/skills/*/SKILL.md` must have a matching `.cursor/skills/*/SKILL.md` pointer. Catches the case where a dev adds a Claude skill but forgets the Cursor pointer, which would silently leave Cursor users unable to invoke it.

The hook fires on commit (early feedback, before push) via the `files:` pattern covering all four directories.

**Breadcrumbs for future maintainers**

- `CLAUDE.md` — documents both artifact types and their sync rules with rationale.
- `.claude/rules/*.md` — HTML `<!-- sync-target: ... -->` comment explaining the derive→sync workflow and why pointers are unsafe for `alwaysApply` rules.
- `.cursor/rules/*.mdc` — YAML `#` comments in frontmatter citing the canonical source and the compaction risk.
- `.claude/skills/*/SKILL.md` and `.cursor/skills/*/SKILL.md` — YAML `#` comments explaining the pointer-is-safe rationale.

## Principles covered (rule)

Test seams without global mutable state, no double parsing, error discriminability across layer boundaries, internally consistent visibility, newtypes over manual Debug/Clone impls, upholding documented contracts when adding match arms.

## Checklist covered (skill)

Commit-scope discipline, assertion strictness, refactor-stale comments, user-facing string changes called out, fixture self-sabotage, dead-code helpers, discarded return values, mock path anchoring, matches! context messages, ignored-test hygiene.

The rule fires while writing; the skill fires before declaring done. The two cross-reference each other at their footers.

### DIFF
--- a/.claude/rules/code-review-design-discipline.md
+++ b/.claude/rules/code-review-design-discipline.md
@@ -1,0 +1,152 @@
+# Code Review Design Discipline
+
+Six principles that came up repeatedly during cross-driver review.
+Apply them while writing the change, not after.
+
+## 1. Don't ship test seams as global mutable state
+
+When you need test-time behavior to differ from production, in priority
+order:
+
+1. **Dependency injection through configuration** — pass the overridable
+   behavior as a field on a struct that's already flowing through the
+   system. Zero global state.
+2. **Compile-time substitution** via `cfg(any(test, feature = "test-utils"))`
+   defaults. No runtime cost, no atomic loads, no unreachable production
+   branches.
+3. **Per-test scoping** (e.g. `tokio::task_local!`) when DI is awkward.
+4. **Process-global state**, only if cfg-gated so production can't see it.
+   Use `Relaxed` ordering for kill switches; `SeqCst` is overkill.
+
+If the codebase already has a DI seam (`Box<dyn …>`, `Arc<dyn …>`,
+function-pointer field), plumb it through instead of adding a static.
+
+```rust
+// ❌ Process-wide irreversible kill switch in production code
+pub(super) static BROWSER_LAUNCH_DISABLED: AtomicBool = AtomicBool::new(false);
+fn default_launch() -> Launcher {
+    Box::new(|url| async move {
+        if BROWSER_LAUNCH_DISABLED.load(SeqCst) { return; }  // ← read on every prod login
+        webbrowser::open(url)
+    })
+}
+
+// ✅ Config-injected, cfg-defaulted, no global state
+pub(crate) browser_launcher: Option<Arc<dyn Fn() -> Launcher + Send + Sync>>,
+// in from_settings():
+#[cfg(any(test, feature = "test-utils"))]
+let browser_launcher = Some(Arc::new(|| Box::new(|_| async {})));
+#[cfg(not(any(test, feature = "test-utils")))]
+let browser_launcher = None;
+```
+
+## 2. Don't parse the same thing twice
+
+Before adding a parser/builder for a config type, grep for existing
+functions that produce the same target from the same inputs. If you find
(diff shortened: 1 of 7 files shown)


.claude/rules/code-review-design-discipline.md
```diff
@@ -0,0 +1,140 @@
...
+pub(super) static BROWSER_LAUNCH_DISABLED: AtomicBool = AtomicBool::new(false);
+fn default_launch() -> Launcher {
+    Box::new(|url| async move {
+        if BROWSER_LAUNCH_DISABLED.load(SeqCst) { return; }  // ← read on every prod login
+        webbrowser::open(url)
+    })
+}
+
+// ✅ Config-injected, cfg-defaulted, no global state
+pub(crate) browser_launcher: Option<Arc<dyn Fn() -> Launcher + Send + Sync>>,
```
sfc-gh-jszczerbinski: So I strongly disagree with having this long-term. I understand that having compile time injection for tests is easier than threading a DI object from top-level.

Since we prioritize dev speed at least till GA, it's sensible to have this as a last resort option. Let's have DI as primary example here and this as "workaround" / "hack" secondary example

sfc-gh-jszczerbinski (changes requested):
Can we find a way not to have duplicated files for claude and cursor?

sfc-gh-boler:
> Can we find a way not to have duplicated files for claude and cursor?

Internal best practise is https://docs.google.com/document/d/19AzCwCdWovX6rKWQpXCW9jFTp5dPCQjobAdDDgQ_u0U/edit, but it depends on sf cli. I'll see I can extract the core stuff into a standalone script here.

sfc-gh-boler:
> Can we find a way not to have duplicated files for claude and cursor?

Unfortunately there is no better way to handle "rules" this today other than duplication. Skills can use pointers. 

Expanded the PR to do that and help keep both cursor and claude rules in sycn.

sfc-gh-jszczerbinski (approved):

PR #1257, opened 2026-05-20
Fix DTM replay tests for SHOW TABLES 26-column schema

sfc-gh-jszczerbinski:
## Summary
- Snowflake added 3 new columns (`search_optimization`, `search_optimization_progress`, `search_optimization_bytes`) after `change_tracking` in `SHOW TABLES` output
- Updated both failing DTM replay tests to expect 26 columns instead of 23, with correct column names, types, and positions
- Verified passing against reference ODBC driver v3.16.0

## Test plan
- [x] Run `run_reference.sh -R "exec_direct_getdata_(14col_26rows|23col_1rows)"` — both pass
- [ ] CI runs full reference test suite

### DIFF
--- a/odbc_tests/tests/replay/datometry/exec_direct_getdata_14col_26rows/test.cpp
+++ b/odbc_tests/tests/replay/datometry/exec_direct_getdata_14col_26rows/test.cpp
@@ -79,8 +79,7 @@ TEST_CASE("Replay: exec_direct_getdata_14col_26rows", "[dtm]") {
     SQLSMALLINT numCols = 0;
     SQLRETURN ret = SQLNumResultCols(stmt0, &numCols);
     CHECK_THAT(OdbcResult(ret, SQL_HANDLE_STMT, stmt0), OdbcMatchers::IsSuccess());
-    // Matches the recorded trace: SHOW TABLES returns 23 columns.
-    CHECK(numCols == 23);
+    CHECK(numCols == 26);
   }
 
   // SQLDescribeCol col 1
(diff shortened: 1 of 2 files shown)


odbc_tests/tests/replay/datometry/exec_direct_getdata_14col_26rows/test.cpp
```diff
@@ -79,8 +79,7 @@ TEST_CASE("Replay: exec_direct_getdata_14col_26rows", "[dtm]") {
     SQLSMALLINT numCols = 0;
     SQLRETURN ret = SQLNumResultCols(stmt0, &numCols);
     CHECK_THAT(OdbcResult(ret, SQL_HANDLE_STMT, stmt0), OdbcMatchers::IsSuccess());
-    // Matches the recorded trace: SHOW TABLES returns 23 columns.
```
sfc-gh-rkowalski, line 82: Maybe these assertions are too strict? It will break again if columns are added/reordered. Perhaps we should not assert on total number and exact column ordering?

sfc-gh-rkowalski (approved):

odbc_tests/tests/replay/datometry/exec_direct_getdata_14col_26rows/test.cpp
sfc-gh-jszczerbinski: True, BTW those tests are disabled for new driver, I have a follow up planned

sfc-gh-pbulawa (approved):

PR #1261, opened 2026-05-20
SNOW-2872352: jdbc - implement connection close with server-side logout

sfc-gh-asolarski:

PR #1262, opened 2026-05-20
NO-SNOW: jdbc - introduce CoreDriver facade to encapsulate protobuf request construction

sfc-gh-asolarski:
## Summary

- Introduces a `CoreDriver` class that encapsulates all protobuf request construction and provides centralized error conversion (`ServiceException`/`TransportException` → `SQLException`)
- Refactors `SnowflakeConnectionImpl` and `SnowflakeStatementImpl` to use `ProtobufApis.coreDriver` singleton instead of constructing protobuf requests inline
- Removes boilerplate try/catch blocks from every call site — error handling is now in one place

## Context

Mirrors the Python `CoreDriver` pattern (`python/src/snowflake/connector/_internal/api_client/client_api.py`) where all core driver calls go through a facade that hides proto request/response details. This gives us:

- A single place to change error mapping logic
- Callers that never import `*Request` types
- Cleaner diffs when adding new core driver methods
- Simpler test mocking (mock `CoreDriver` directly)

### DIFF
--- a/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/SnowflakeConnectionImpl.java
+++ b/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/SnowflakeConnectionImpl.java
@@ -33,23 +33,15 @@ import net.snowflake.client.internal.api.implementation.statement.SnowflakePrepa
 import net.snowflake.client.internal.api.implementation.statement.SnowflakeStatementImpl;
 import net.snowflake.client.internal.log.SFLogger;
 import net.snowflake.client.internal.log.SFLoggerFactory;
+import net.snowflake.client.internal.unicore.CoreDriverApi;
 import net.snowflake.client.internal.unicore.ProtobufApis;
-import net.snowflake.client.internal.unicore.TransportException;
-import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverService;
 import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.ConfigSetting;
-import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.ConnectionCloseRequest;
 import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.ConnectionHandle;
-import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.ConnectionInitRequest;
-import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.ConnectionNewRequest;
-import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.ConnectionReleaseRequest;
-import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.ConnectionSetOptionsRequest;
 import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.ConnectionSetOptionsResponse;
 import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.DatabaseHandle;
-import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.DatabaseInitRequest;
-import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.DatabaseNewRequest;
-import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.DatabaseReleaseRequest;
 import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.ValidationIssue;
 import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.WrapperIdentity;
+import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.WrapperIdentity.Builder;
 import net.snowflake.client.internal.util.NotImplementedException;
 
 public class SnowflakeConnectionImpl implements SnowflakeConnection, Connection {
@@ -57,6 +49,7 @@ public class SnowflakeConnectionImpl implements SnowflakeConnection, Connection 
   private static final SFLogger logger = SFLoggerFactory.getLogger(SnowflakeConnectionImpl.class);
   private final String url;
   private final Properties properties;
+  private final CoreDriverApi coreDriverApi;
   private boolean autoCommit = true;
   private final AtomicBoolean closed = new AtomicBoolean(false);
   private String catalog;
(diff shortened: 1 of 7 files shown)


sfc-gh-asolarski:
* **#1272** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1272?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1262** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1262?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1262?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1261** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1261?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/SnowflakeConnectionImpl.java
```diff
@@ -34,20 +34,10 @@
...
-import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverService;
 import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.ConfigSetting;
-import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.ConnectionCloseRequest;
 import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.ConnectionHandle;
-import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.ConnectionInitRequest;
-import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.ConnectionNewRequest;
-import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.ConnectionReleaseRequest;
-import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.ConnectionSetOptionsRequest;
 import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.ConnectionSetOptionsResponse;
 import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverV1.DatabaseHandle;
```
sfc-gh-dheyman, line 41: there is still a lot of proto leaking here. Can we do it right?

jdbc/src/main/java/net/snowflake/client/internal/unicore/CoreDriver.java
```diff
@@ -0,0 +1,395 @@
...
+
+/**
+ * Facade over {@link DatabaseDriverService} that encapsulates protobuf request construction and
+ * provides centralized error conversion from core driver exceptions to {@link SQLException}.
+ *
+ * <p>Callers interact with domain-level parameters (handles, strings, maps) and never need to
+ * import or construct {@code *Request} protobuf objects directly.
+ */
+@RequiredArgsConstructor
+public class CoreDriver {
```
sfc-gh-dheyman: I'd make an interface first. Perhaps CoreDriverApi would be a better name?

jdbc/src/main/java/net/snowflake/client/internal/unicore/CoreDriver.java
```diff
@@ -0,0 +1,395 @@
...
+  @FunctionalInterface
+  interface ServiceCall<T> {
+    T call() throws DatabaseDriverService.ServiceException, TransportException;
+  }
+
+  private <T> T invoke(ServiceCall<T> callable) throws SQLException {
+    try {
+      return callable.call();
+    } catch (DatabaseDriverService.ServiceException e) {
+      throw SnowflakeSQLException.fromServiceException(null, e);
```
sfc-gh-dheyman: I think instead of passing null here we should pass an operation name.

jdbc/src/test/java/net/snowflake/client/internal/api/implementation/connection/SnowflakeConnectionImplTest.java
```diff
@@ -82,48 +79,49 @@ public void toConfigSettingReturnsNullForUnsupportedValues() {
...
-          .thenReturn(ConnectionInitResponse.getDefaultInstance());
-
       Properties props = new Properties();
       props.setProperty("account", "test_account");
       props.setProperty("user", "test_user");
       props.setProperty("password", "test_password");
       return new SnowflakeConnectionImpl("jdbc:snowflake://test.snowflakecomputing.com", props);
     }
 
-    private void stubSuccessfulClose() {
```
sfc-gh-dheyman, line 120: do we have a test to verify whether on unhappy path handles are still released?

jdbc/src/main/java/net/snowflake/client/internal/unicore/ProtobufApis.java
```diff
@@ -1,9 +1,9 @@
 package net.snowflake.client.internal.unicore;
 
-import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverService;
 import net.snowflake.client.internal.unicore.protobuf_gen.DatabaseDriverServiceClient;
 
 public class ProtobufApis {
-  public static DatabaseDriverService databaseDriverV1 =
-      new DatabaseDriverServiceClient(new JNICoreTransport());
+
+  public static CoreDriver coreDriver =
```
sfc-gh-dheyman: perhaps we can inject this into `SnowflakeConnectionImpl` via constructor for better testability?

jdbc/src/main/java/net/snowflake/client/internal/unicore/CoreDriver.java
sfc-gh-dheyman: honestly it would be great to have this generated from `proto_generator/src/generators/java_generator.rs`.

sfc-gh-dheyman (changes requested):

jdbc/src/main/java/net/snowflake/client/internal/unicore/CoreDriver.java
sfc-gh-asolarski: extracted

sfc-gh-asolarski: Agreed, that would be the ideal end state. For now we have the interface (CoreDriverApi) + manual implementation. We can look into generating it from java_generator.rs as a follow-up once the facade API stabilizes.

jdbc/src/main/java/net/snowflake/client/internal/unicore/CoreDriver.java
sfc-gh-asolarski: It;s was just a fallback in case of empty driver exception inside (which should never happen.  
I decided to remove the `fallbackMessage` parameter entirely. It now uses `exception.getMessage()` in the fallback path.

jdbc/src/main/java/net/snowflake/client/internal/unicore/ProtobufApis.java
sfc-gh-asolarski: good idea indeed, done

jdbc/src/test/java/net/snowflake/client/internal/api/implementation/connection/SnowflakeConnectionImplTest.java
sfc-gh-asolarski: yes, `releasesHandlesEvenWhenConnectionCloseThrows`

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/SnowflakeConnectionImpl.java
sfc-gh-asolarski: That's true but this is the first step toward encapsulating protobuf behind the facade.  
We can address it incrementally: next steps would be having CoreDriverApi accept/return higher-level types (e.g. wrapper classes for handles, options as a Map\<String, Object>, etc.) so that proto doesn't leak into the connection impl. This PR establishes the design direction.

sfc-gh-dheyman (approved):

PR #1270, opened 2026-05-21
NO-SNOW: ODBC Behavioral Differences 11 - 37 Review

sfc-gh-pbulawa:


sfc-gh-ddas (approved):

odbc_tests/BehaviorDifferences.yaml
```diff
@@ -221,10 +230,12 @@ behavior_differences:
...
-    name: "NaN FLOAT value silently converts to 0 for integer and BIT targets instead of returning 22003"
-    status: unknown
-    type: bug
-    reviewed: false
+    name: "NaN FLOAT value returns SQL_ERROR (22003) instead of silently converting to 0 for integer and BIT targets"
+    status: allowed
+    type: bugfix
+    reviewed: true
+    status_rationale: |
+      This is a ODBC compliance fix. The silent truncation may cause data discrepancies.
```
sfc-gh-boler: Calling this truncation is maybe a bit of a stretch. But more generally, it's likely we had users who knowingly or not depended on this behaviour. I think this breaking change is still fine, because it returns a loud error. However, we should somehow annotate this bD, so that it is prominent in release notes.

odbc_tests/BehaviorDifferences.yaml
```diff
@@ -247,10 +260,12 @@ behavior_differences:
       Applications that parse the PUT source column for the full path on Windows will now receive only the filename.
 
   18:
-    name: "Interval leading field precision is not enforced and SQL_DESC_DATETIME_INTERVAL_PRECISION is not supported"
-    status: unknown
-    type: bug
+    name: "Interval leading field precision is enforced and SQL_DESC_DATETIME_INTERVAL_PRECISION is respected"
+    status: allowed
+    type: bugfix
     reviewed: false
```
sfc-gh-boler: reviewed=true?

odbc_tests/BehaviorDifferences.yaml
```diff
@@ -357,29 +384,35 @@ behavior_differences:
 
   26:
     name: "SQLBindParameter rejects negative DecimalDigits with HY104 instead of accepting silently"
-    status: unknown
+    status: allowed
     type: bug
-    reviewed: false
+    reviewed: true
```
sfc-gh-boler, line 389: bugfix?

odbc_tests/BehaviorDifferences.yaml
```diff
@@ -357,29 +384,35 @@ behavior_differences:
...
   27:
-    name: "Unbound parameter marker returns 42000 (server error) instead of 07002 (COUNT field incorrect)"
-    status: unknown
+    name: "Unbound parameter marker returns SQL_ERROR (42000) server error instead of SQL_ERROR (07002) COUNT field incorrect"
+    status: allowed
     type: enhancement
-    reviewed: false
+    reviewed: true
+    status_rationale: |
+      Old driver parsed the query locally instead of passing it to the server. We are returning backend error now, as this is the role of the server to process the SQL query with the parser, instead of making the local parser version that may shift from the server's implementation.
```
sfc-gh-boler: The rationale should justify why this requires breaking current API contract, i.e. why don't we translate 42000 to 07002 in the client?

odbc_tests/BehaviorDifferences.yaml
```diff
@@ -414,6 +451,8 @@ behavior_differences:
     status: unknown
     type: bug
     reviewed: false
+    status_rationale: |
+      To be reviewed with Product. This is a big deviation from the ODBC spec, but some customers may explicitly use this behavior in the products.
```
sfc-gh-boler, line 455: what would take to provide a user-controlled param to choose either behaviour?

sfc-gh-boler (changes requested):
Largely looks good, but there are two comments (the last two) I'd like to close before merging.

odbc_tests/BehaviorDifferences.yaml
sfc-gh-mhofman, line 8: do we need flags: 
- IsBreakingChange
- Severity
cc @sfc-gh-turbaszek

odbc_tests/BehaviorDifferences.yaml
sfc-gh-pbulawa: Agree. For the PuPr/GA we have to list all of the fixes/changes in the release notes and documentation to keep everyone aligned.

sfc-gh-pbulawa: Renamed to `silent conversion`.

odbc_tests/BehaviorDifferences.yaml
sfc-gh-pbulawa: Fixed

odbc_tests/BehaviorDifferences.yaml
sfc-gh-pbulawa: Fixed

odbc_tests/BehaviorDifferences.yaml
sfc-gh-pbulawa: Added explanation.

odbc_tests/BehaviorDifferences.yaml
sfc-gh-pbulawa: It may not be trivial, as this condition could potentially require alignment between multiple API functions.

sfc-gh-pbulawa (commented):
Comments

odbc_tests/BehaviorDifferences.yaml
sfc-gh-boler: I'm still seeing the old snapshot

odbc_tests/BehaviorDifferences.yaml
sfc-gh-boler: Still seeing an old snapshot here

odbc_tests/BehaviorDifferences.yaml
sfc-gh-boler: How would you define IsBreakingChange?

Severity: sgtm

odbc_tests/BehaviorDifferences.yaml
sfc-gh-boler: Because it's likely some customer code depends on the old behaviour, our default answer should be: please copy the bug. So it would be helpful to have some sort of complexity / effort estimate to maybe change the default answer.

sfc-gh-boler (changes requested):
Difference #31 has a discussion I'd like to see complete before we merge this.

odbc_tests/BehaviorDifferences.yaml
sfc-gh-boler: (Resolving: per offline discussion, no recommendation has been made yet but we are leaning towards preserving the old bevahiour unconditionally.)

sfc-gh-boler (approved):

PR #1279, opened 2026-05-22
NO-SNOW: Load native lib from JAR resource so it works without CORE_PATH

sfc-gh-rkowalski:
The JNI native lib (libjdbc_bridge.so/.dylib/.dll) was previously loadable only via the CORE_PATH env var (or jdbc.library.path system property), both of which require the user to know an absolute filesystem path to the lib. This breaks deployment to sandboxed JVMs - Snowflake UDFs, Exasol BucketFS UDFs, AWS Lambda, GraalVM Native Image runners, etc. - where host env vars aren't visible and the host filesystem layout isn't predictable.

Fix:

1. jdbc/build.gradle copyNativeLib task: glob the actual cargo output filenames (libjdbc_bridge.so/.dylib + jdbc_bridge.dll) instead of libjdbc.* which never matched anything. Place them under build/resources/main/net/snowflake/client/internal/unicore/native/ so they're addressable as a JAR resource at runtime.

2. JNICoreTransport static initializer: try CORE_PATH and jdbc.library.path first (back-compat), then fall back to extracting the bundled native lib from the JAR resource into a per-JVM temp dir and System.load()ing it.

This is the standard sqlite-jdbc / netty / snappy-java pattern. No behavioral change for existing CORE_PATH-based callers.

### DIFF
--- a/jdbc/src/main/java/net/snowflake/client/internal/unicore/JNICoreTransport.java
+++ b/jdbc/src/main/java/net/snowflake/client/internal/unicore/JNICoreTransport.java
@@ -6,7 +6,7 @@ import net.snowflake.client.internal.log.SFLoggerFactory;
 public class JNICoreTransport implements CoreTransport {
   private static final SFLogger logger = SFLoggerFactory.getLogger(JNICoreTransport.class);
 
-  static {
+  public JNICoreTransport() {
     NativeLibraryLoader.init();
   }
 
--- a/jdbc/src/main/java/net/snowflake/client/internal/unicore/JNICoreTransport.java
+++ b/jdbc/src/main/java/net/snowflake/client/internal/unicore/JNICoreTransport.java
@@ -1,6 +1,5 @@
 package net.snowflake.client.internal.unicore;
 
-import net.snowflake.client.api.driver.SnowflakeDriver;
 import net.snowflake.client.internal.log.SFLogger;
 import net.snowflake.client.internal.log.SFLoggerFactory;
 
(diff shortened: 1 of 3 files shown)


jdbc/src/main/java/net/snowflake/client/internal/unicore/JNICoreTransport.java
```diff
@@ -1,38 +1,107 @@
...
+   * <ol>
+   *   <li>{@code CORE_PATH} env var — explicit absolute path (back-compat for dev/CI setups).
+   *   <li>{@code jdbc.library.path} system property — explicit absolute path.
+   *   <li>JAR resource — extract the bundled native lib to a temp file and load from there.
+   * </ol>
+   *
+   * <p>The JAR-resource path is the default for end users: it works in sandboxed JVMs (e.g. UDFs,
+   * BucketFS, Lambda) where host env vars and host filesystem layouts aren't predictable.
+   */
+  private static void loadNativeLibrary() {
```
sfc-gh-asolarski: hmm, maybe this whole logic (with helpers) should be encapsulated as this is not only about handling messages (transport)

sfc-gh-asolarski (approved):

jdbc/src/main/java/net/snowflake/client/internal/unicore/JNICoreTransport.java
sfc-gh-rkowalski: 👍 Moved to separate class

PR #1280, opened 2026-05-22
NO-SNOW: consolidate JDBC driver version constants

sfc-gh-dheyman:
Make build.gradle's project.version the single source of truth for the
JDBC driver version via a new generateDriverVersion Gradle task that
writes a package-private DriverVersion.java at compile time (mirrors
the existing generateProtobuf pattern). Promote DRIVER_NAME, DRIVER_VERSION,
JDBC_SPEC_VERSION, MAJOR_VERSION, MINOR_VERSION, JDBC_SPEC_MAJOR,
JDBC_SPEC_MINOR to public constants on SnowflakeDriver and derive
major/minor from DRIVER_VERSION via a regex-based parseVersionComponent
helper so they cannot drift.

Wire SnowflakeDatabaseMetaDataImpl.getDriverName/Version/Major/Minor
(previously NotImplementedException) and the JDBC-spec methods through
the SnowflakeDriver constants. Align getDriverName() to "Snowflake JDBC
Driver" (was "Snowflake"). Delete SnowflakeConnectionImpl.determineClientAppVersion()
and read SnowflakeDriver.DRIVER_VERSION directly when populating
WrapperIdentity.

Latent-bug fix: previously Driver.getDriverVersion()/getMajorVersion()/
getMinorVersion() returned hardcoded constants while WrapperIdentity sent
a manifest-resolved value. They now agree by construction.

Adds SnowflakeDriverTest version cases on top of the existing URL/
registration tests, plus a new SnowflakeDatabaseMetaDataImplTest.

PR #1281, opened 2026-05-22
NO-SNOW: implement getDatabaseProductVersion via cached SELECT CURRENT_VERSION()

sfc-gh-dheyman:
Implement getDatabaseVersion/MajorVersion/MinorVersion on SnowflakeConnectionImpl
via a per-connection cache populated lazily on first call by running
SELECT CURRENT_VERSION() (mirrors the Python connector pattern). Wire
SnowflakeDatabaseMetaDataImpl.getDatabaseProductVersion() to the connection.

The build suffix returned by the server (e.g. "8.46.1 abcdef") is stripped
to match the legacy Snowflake JDBC driver and the Python connector.
Failures propagate as SQLException; the cache is not populated on failure
so the next call retries.

Cache uses double-checked locking with a volatile field. Widens
SnowflakeDriver.parseVersionComponent to public so the connection (in a
different package) can reuse it.

The protobuf ConnectionGetInfoResponse does not currently carry
vendor_version, so we follow the SELECT CURRENT_VERSION() approach.
Avoiding the round-trip via a protocol extension is a worthwhile follow-up
but out of scope here.

Adds 4 unit tests for stripVersionSuffix, 1 Mockito test verifying
metadata-to-connection delegation, and a DatabaseVersionTests e2e suite
covering anchored regex, parity with raw query, major/minor parsing, and
value-stability across calls.

### DIFF
--- a/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/SnowflakeConnectionImpl.java
+++ b/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/SnowflakeConnectionImpl.java
@@ -58,6 +58,9 @@ public class SnowflakeConnectionImpl implements SnowflakeConnection, Connection 
   private final DatabaseHandle databaseHandle;
   public final ConnectionHandle connectionHandle;
 
+  private volatile String cachedDatabaseVersion;
+  private final Object databaseVersionLock = new Object();
+
   public SnowflakeConnectionImpl(String url, Properties properties) throws SQLException {
     this(url, properties, ProtobufApis.coreDriverApi);
   }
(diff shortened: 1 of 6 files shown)


sfc-gh-dheyman:
* **#1281** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1281?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1281?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1280** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1280?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/SnowflakeConnectionImpl.java
```diff
@@ -539,16 +542,51 @@ public String[] getChildQueryIds(String queryID) throws SQLException {
...
   public int getDatabaseMinorVersion() throws SQLException {
-    throw new NotImplementedException();
+    return SnowflakeDriver.parseVersionComponent(getDatabaseVersion(), 1);
   }
 
+  /** Issues a single {@code SELECT CURRENT_VERSION()} on first call per connection; cached. */
   @Override
   public String getDatabaseVersion() throws SQLException {
-    throw new NotImplementedException();
+    checkClosed();
```
sfc-gh-asolarski, line 556: if these calls are consistent across the session then maybe this locking is not even necessary, as anyway those calls are pretty chip

sfc-gh-asolarski (approved):
if you have problems with mocking you can rebase on the top of [PR 1272](https://app.graphite.com/github/pr/snowflakedb/universal-driver/1272/SNOW-2872352-jdbc---implement-statement-release-and-connection-level-statement-tracking)
it's solved there

PR #1286, opened 2026-05-22
NO-SNOW: pr-review-bot: switch reviewer config to YAML with notify/remind opt-outs

sfc-gh-turbaszek:
## Summary
- Convert the reviewer pool from the bespoke `<label> @login1 @login2` text format at `.github/reviewers` to YAML at `.github/reviewers.yml` with two top-level sections: a `reviewers:` map of GitHub login -> per-user prefs and a `rules:` map of PR-label -> `[logins]`.
- Add per-reviewer Slack opt-out flags (both default `true`):
  - `notify`: post the channel announcement when this user is picked as the initial reviewer for a new PR. `false` skips just that one channel post; the GitHub-side review request and assignee call still happen, so the user remains fully assignable.
  - `remind`: appear in the recurring reminder digest for stale PRs waiting on this user. `false` hides them from those digest entries.
- Update `scripts/pr_review_bot.py` to parse YAML (PyYAML), honour the two opt-outs, and update docstrings. The candidate pool is intentionally unchanged: every declared reviewer is always assignable; the opt-outs only suppress Slack pings, never the GitHub assignment itself.
- Update `.github/workflows/pr-review-bot.yml` to install PyYAML (`>=6,<7`) before running the script in both jobs.
- Update `.github/CODEOWNERS` comment to reference the new file path.

## Context
The previous config (#1173 / #1159) used a tiny custom parser for a label-keyed text file. As the team grew, a few reviewers asked to keep being assignment-eligible without getting Slack pings (initial assignment OR recurring digest). YAML lets us encode that cleanly without adding new files or surfaces, and PyYAML is a well-known stable single dep.

## Test plan
- Smoke-tested ``parse_reviewers`` on the real ``.github/reviewers.yml`` and on a synthetic four-user matrix covering all four ``(notify, remind)`` combinations:
  - All four reviewers stay in the candidate pool regardless of flags (everyone is always assignable).
  - ``filter_by_pref(..., \"remind\")`` correctly drops only ``remind: false`` users.
  - ``get_pref`` lookups are case-insensitive and unknown logins fall back to defaults (``True / True``).
- ``python3 -c \"import yaml; yaml.safe_load(open('.github/reviewers.yml')); yaml.safe_load(open('.github/workflows/pr-review-bot.yml'))\"`` succeeds for both files.
- Linter check on the edited files reports no errors.
- Live workflow against GitHub/Slack APIs not exercised here; the next workflow run on this PR will exercise the assign path end-to-end against the real config.


.github/workflows/pr-review-bot.yml
```diff
@@ -1,12 +1,16 @@
...
+# .github/reviewers.yml, request review + add as assignee via the
+# GitHub REST API, and ping #drivers-review via
+# slackapi/slack-github-action. On the schedule below, post a digest
+# of open non-draft PRs that have no APPROVED / CHANGES_REQUESTED
+# review yet; the digest pass also re-assigns PRs whose every
+# requested reviewer is OOO (Slack status) to a non-OOO substitute
+# drawn from the same label pool.
+#
+# .github/reviewers.yml also encodes per-reviewer opt-out flags
+# (`assign`, `remind`); see that file for details.
```
sfc-gh-boler, line 13: Flag is `notify`, not `assign` — everywhere else (the YAML, `_DEFAULT_PREFS`, the post-pick check in `cmd_assign`) calls it `notify`.

.github/workflows/pr-review-bot.yml
```diff
@@ -1,12 +1,16 @@
...
 #
 # All GitHub I/O goes through the `gh` CLI (with GITHUB_TOKEN) and all
 # Slack writes go through slackapi/slack-github-action. The Python
```
sfc-gh-boler, line 16: Missed in the migration: the duplicate header block at lines 39–47 below still describes the old text format and the old `.github/reviewers` path. Either retarget it to the YAML schema or just delete it (largely duplicates the updated block at lines 1–13).

scripts/pr_review_bot.py
```diff
@@ -1182,6 +1367,22 @@ def cmd_assign(args: argparse.Namespace) -> int:
...
+    # review-request notification — only the Slack channel post is
+    # skipped.
+    if not get_pref(prefs, reviewer, "notify"):
+        log.info(
+            "Reviewer %s opted out of initial Slack notification "
+            "(notify: false); skipping channel post for PR #%d.",
+            reviewer,
+            pr_number,
+        )
+        set_gh_output("skip", "true")
```
sfc-gh-boler, line 1383: Could reuse `_skip_assign(reason)` here instead of inlining `set_gh_output("skip", "true"); return 0` — keeps all assign-skip paths uniform.

sfc-gh-boler (approved):

sfc-gh-pczajka (approved):

PR #1288, opened 2026-05-23
SNOW-3571048: odbc_trace_tool — parse Windows ODBC DM traces and emit Excel/Power Query replay tests

sfc-gh-ddas:
### TL;DR

Adds a Windows ODBC Driver Manager (WinODBC) trace parser and extends the C++ code generator to handle `SQLSetStmtAttr`, `SQLColAttribute`, and unsupported ODBC calls gracefully.

### What changed?

**New WinODBC parser (`parser/winodbc.rs`)**
- Parses Windows ODBC DM trace files, including those saved as UTF-16 LE (BOM-aware file reading via `encoding_rs`).
- Detects the WinODBC format by scanning for the characteristic `ENTER`/`EXIT` header pattern before falling back to iODBC/unixODBC detection.
- Normalizes Wide-string API variants (e.g. `SQLExecDirectW` → `SQLExecDirect`) during parsing and call construction.
- Pairs `ENTER`/`EXIT` blocks per thread to reconstruct `TracedCall` entries and builds the handle graph from `SQLAllocHandle` calls.

**New model types**
- `SetStmtAttr` and `ColAttribute` structs added to `model.rs`, with full `OdbcCall` dispatch (return code, handle resolution, handle address mapping, and raw builder functions).
- `SQL_NO_DATA_FOUND` recognized as an alias for `SQL_NO_DATA`.
- `TraceFormat::WinOdbc` variant added.

**C++ generator improvements**
- `emit_set_stmt_attr` and `emit_col_attribute` emitters added.
- `SQLColAttribute` calls with undocumented field identifiers (e.g. field 32 probed by Excel) are silently skipped with a tally comment rather than crashing.
- `SQL_DBMS_VER` and `SQL_DBMS_NAME` added to `DRIVER_SPECIFIC_INFO_TYPES` so version-sensitive strings are asserted for presence rather than exact value.
- `GetData` indicator assertions now check `ind > 0` for non-narrow target types instead of asserting the exact byte count, keeping tests portable across wide-char encodings.
- `generate()` now returns `Result<String, GenerateError>`. If unsupported ODBC calls are encountered, it returns `Err(GenerateError::Unsupported(...))` listing each call name and occurrence count. A new `--allow-unsupported` CLI flag emits `// TODO` comments instead of failing.

**IR split fixes**
[...]

PR #1289, opened 2026-05-24
SNOW-3571049: SQLGetInfo — implement DRIVER_NAME, DRIVER_VER, DBMS_VER info types

sfc-gh-ddas:
### TL;DR

Adds support for `SQLGetInfo(SQL_DRIVER_NAME)`, `SQL_DRIVER_VER`, `SQL_DBMS_VER`, and `SQLColAttribute(SQL_DESC_NAME / SQL_DESC_LENGTH)` in the ODBC driver, backed by a new lightweight `ConnectionGetServerVersion` RPC that reads the Snowflake server version cached from the login response.

### What changed?

- `InfoType` enum gains `DriverName` (6), `DriverVer` (7), and `DbmsVer` (18) variants, with corresponding `get_info` handlers that write the driver name, driver version, and Snowflake server version strings respectively.
- `DescField` enum gains `Length` (1003) and `Name` (1011) variants. `col_attribute` now handles `SQL_DESC_NAME` by writing the column name and `SQL_DESC_LENGTH` by computing the logical column size via `column_size_from_field`.
- `AuthResponseMain.server_version` is promoted from a dead `_server_version` field to a live `server_version` field. It is captured in `LoginResult` and written into a new `Connection.server_version: RwLock<Option<String>>` field during login completion.
- A new `ConnectionGetServerVersion` RPC is added to the `DatabaseDriver` protobuf service. It reads `Connection.server_version` directly, avoiding the heavier `ConnectionGetInfo` aggregation path. Returns `None` (rendered as an empty string) when called before login completes.
- The ODBC `get_info` handler for `DbmsVer` uses `connection_get_server_version` via the blocking runtime extension, falling back to an empty string when the connection is not yet established.

### How to test?

- New unit tests in `sf_core` cover three cases for `connection_get_server_version`: returns `None` before login, returns the cached value after login, and returns `InvalidArgument` for a released handle.
- New unit tests in `odbc_types` pin the numeric discriminants for all `InfoType` and `DescField` variants used by Excel/PowerQuery during `SQLDriverConnect`, guarding against accidental renumbering.
- Connect Excel or PowerQuery to a Snowflake DSN and verify that `SQLDriverConnect` succeeds and the driver name, driver version, and DBMS version are reported correctly in the connection dialog or driver info panel.
- Run the existing AS-bound trace replay tests to confirm no regressions in `SQLGetInfo` dispatch.

### Why make this change?

Excel and PowerQuery probe `SQL_DRIVER_NAME` (6), `SQL_DRIVER_VER` (7), `SQL_DBMS_VER` (18), `SQL_DESC_NAME`, and `SQL_DESC_LENGTH` during `SQLDriverConnect` and result-set preview rendering. Without handlers for these attributes the driver returned `UnknownInfoType` / `UnknownAttribute` errors, causing Excel to fail or silently skip the connection. The `server_version` field was already present in the Snowflake login response but was unused; surfacing it through a dedicated lightweight RPC avoids the overhead of the full `ConnectionGetInfo` aggregation on every per-statement probe.

### DIFF
--- a/odbc/src/api/connection.rs
+++ b/odbc/src/api/connection.rs
@@ -1189,7 +1189,7 @@ pub fn get_info<E: OdbcEncoding>(
 ) -> OdbcResult<()> {
     tracing::debug!("get_info: connection_handle={connection_handle:?}, info_type={info_type}");
 
-    let _conn = conn_from_handle(connection_handle)?;
+    let dbc = conn_from_handle(connection_handle)?;
 
     let info_type = InfoType::try_from(info_type)?;
     tracing::debug!("get_info: info_type={info_type:?}");
@@ -1209,9 +1209,68 @@ pub fn get_info<E: OdbcEncoding>(
             }
             Ok(())
         }
+        InfoType::DriverName => {
+            write_string_bytes::<E>(
+                ODBC_DRIVER_NAME,
+                info_value_ptr as *mut E::Char,
+                buffer_length,
+                string_length_ptr,
+                None,
+            );
+            Ok(())
+        }
+        InfoType::DriverVer => {
+            write_string_bytes::<E>(
+                ODBC_DRIVER_VERSION,
+                info_value_ptr as *mut E::Char,
+                buffer_length,
+                string_length_ptr,
+                None,
+            );
+            Ok(())
+        }
         InfoType::DbmsName => {
             write_string_bytes::<E>(
                 "Snowflake",
+                info_value_ptr as *mut E::Char,
+                buffer_length,
+                string_length_ptr,
+                None,
+            );
+            Ok(())
+        }
+        InfoType::DbmsVer => {
+            // Sourced from `serverVersion` in the login response (parsed in
+            // [`sf_core::rest::snowflake::auth::AuthResponseMain`]). Matches
+            // the legacy driver and avoids the extra `SELECT CURRENT_VERSION()`
+            // round trip that JDBC currently performs.
+            //
+            // Uses the dedicated `connection_get_server_version` getter
+            // rather than `connection_get_info` — Excel polls this attribute
+            // during `SQLDriverConnect` and the full info aggregation is
+            // wasteful when only the version is needed.
+            //
+            // Before the connection is established, sf_core has no
+            // `server_version` yet — return an empty string instead of
+            // surfacing an error so callers that probe this attribute during
+            // `SQLDriverConnect` (Excel does) still succeed.
+            let conn_handle = match dbc.connection.lock().state {
+                ConnectionState::Connected { conn_handle, .. } => Some(conn_handle),
+                ConnectionState::Disconnected => None,
+            };
+            let version = match conn_handle {
+                Some(handle) => global().context(OdbcRuntimeSnafu)?.block_on(async |c| {
+                    let resp = c
+                        .connection_get_server_version(ConnectionGetServerVersionRequest {
+                            conn_handle: Some(handle),
+                        })
+                        .await?;
+                    Ok::<Option<String>, crate::api::OdbcError>(resp.server_version)
+                })?,
+                None => None,
+            };
+            write_string_bytes::<E>(
+                version.as_deref().unwrap_or(""),
                 info_value_ptr as *mut E::Char,
                 buffer_length,
                 string_length_ptr,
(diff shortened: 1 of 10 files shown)


sfc-gh-ddas:
* **#1307** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1307?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1291** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1291?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1290** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1290?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1289** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1289?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1289?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1288** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1288?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sfc-gh-mhofman:
PR title/commit say "catalogue params" but no catalog APIs are touched.
PR actually adds SQLGetInfo info types, SQLColAttribute desc fields, and ConnectionGetServerVersion RPC.
Please update title/desc for future researchers/analysis to be accurate.

odbc/src/api/connection.rs
sfc-gh-mhofman, line 1959: don't we need to turn on the tests for driver name, driver version and dbms ver in odbc_tests? 
They're still tagged SKIP_NEW_DRIVER_NOT_IMPLEMENTED

Either lift the SKIP_NEW_DRIVER_NOT_IMPLEMENTED() guards on the three SQLGetInfo: SQL_DRIVER_NAME/SQL_DRIVER_VER/SQL_DBMS_VER test cases (or add new tests) so the new code paths are actually verified.

odbc/src/api/connection.rs
```diff
@@ -1209,6 +1209,26 @@ pub fn get_info<E: OdbcEncoding>(
             }
             Ok(())
         }
+        InfoType::DriverName => {
+            write_string_bytes::<E>(
+                ODBC_DRIVER_NAME,
```
sfc-gh-mhofman, line 1214: Spec defines that strictly as file name of the driver and here we just return ODBC:

> A character string with the file name of the driver used to access the data source.

https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlgetinfo-function?view=sql-server-ver17

What do we report for the current driver?

odbc/src/api/connection.rs
```diff
@@ -1209,6 +1209,26 @@ pub fn get_info<E: OdbcEncoding>(
             }
             Ok(())
         }
+        InfoType::DriverName => {
```
sfc-gh-mhofman, line 1212: SQL_DRIVER_NAME / SQL_DRIVER_VER are now returned without checking connection state. The match arms unconditionally call write_string_bytes without consulting dbc.connection.lock().state
This is more spec compliant than in current driver but it conflicts with existing tests: odbc_tests/tests/odbc_api/driver_info/get_info_tests.cpp - ln 3983+
It will fail if Driver Manager for new driver do not intercept the errors.

sfc-gh-mhofman (changes requested):

sfc-gh-mhofman (approved):
As discusssed all fine

odbc/src/api/connection.rs
sfc-gh-ddas: Removed the guard on the tests. Kept the guard on exact values of driver name and versiona s these are not final.

odbc/src/api/connection.rs
sfc-gh-ddas: We have not finalized the version and driver name for the UD yet and making this call is out of scope for this PR. After this PR, it should be easy to change when the decision is made.

odbc/src/api/connection.rs
sfc-gh-ddas: Modified the test accordingly and added a BD.

PR #1291, opened 2026-05-24
SNOW-3571057: ODBC test fixtures — add DATATYPETESTS schema (ALLDATATYPES + LARGELOBS) to setup_readonly_metadata_db.sql

sfc-gh-ddas:
### TL;DR

Adds a `DATATYPETESTS` schema with data-bearing fixture tables to the readonly metadata DB setup script for use in Excel/Power Query trace replay tests.

### What changed?

Two new tables are introduced under the `DATATYPETESTS` schema:

- **`ALLDATATYPES`** — covers the full range of Snowflake data types (numeric, string, binary, boolean, date/time, semi-structured, and geography) with four representative rows:
  - `NORMAL` — typical, representative values
  - `BOUNDARY` — edge-case values (max integers, `Infinity`, `NaN`, max dates, etc.)
  - `UNICODE` — multi-script and emoji strings to exercise W-API encoding
  - `NULLROW` — all nullable columns set to `NULL`

- **`LARGELOBS`** — contains a `FULL` row with a ~128 KB `VARCHAR` and ~256 KB `VARBINARY` payload to exercise chunked LOB streaming via `SQLGetData`, and a `NULLROW` to assert `SQL_NULL_DATA` behavior on both LOB columns.

### How to test?

Run the updated `setup_readonly_metadata_db.sql` script against a Snowflake environment and verify that the `DATATYPETESTS` schema is created with both tables populated. Trace replay tests targeting type conversion, W-API encoding, and LOB streaming can then reference the `NORMAL`, `BOUNDARY`, `UNICODE`, and `NULLROW` row kinds by name rather than relying on row order.

### Why make this change?

The existing `CATALOGTESTS` schema only contains DDL-level fixtures with no row data. These new fixtures provide stable, deterministic rows needed to replay and validate ODBC driver behavior around type conversion, wide-character API encoding, and large object streaming without depending on external or ephemeral data.

PR #1299, opened 2026-05-25
SNOW-3571146: harden protoc download against transient GitHub Releases 5xx

sfc-gh-boler:
## Why

Build sf_core (macOS) regularly fails on `main` when GitHub Releases returns a transient HTTP 5xx for the protoc download. The current installer panics on a single non-200 (curl exit 22), so any 504 turns the whole Python release build red. This is the second time we've triaged "GitHub Releases 504 → release build red", so this PR absorbs the failure mode rather than retriaging the next occurrence.

**Triggering run:** [actions/runs/26399594496](https://github.com/snowflakedb/universal-driver/actions/runs/26399594496) (`build_wheels (macosx_x86_64)` → step *Build sf_core (macOS)*)
**Jira:** [SNOW-3571146](https://snowflakecomputing.atlassian.net/browse/SNOW-3571146)
**Triage thread:** [Slack](https://snowflake.slack.com/archives/C0B544MNQTD/p1779711321738349)

## What

Two layers of defence, neither sufficient on its own:

1. **`proto_generator/src/protoc_installer.rs`** now passes curl `--retry 5 --retry-all-errors --retry-delay 2 --retry-max-time 120` plus connect/total timeouts, and wraps the curl invocation in an outer 5-attempt exponential-backoff loop (1s → 2s → 4s → 8s). `--retry-all-errors` is the load-bearing flag: curl's default `--retry` ignores HTTP 5xx, which is exactly the failure mode we see. The outer loop covers the case where curl itself exits non-zero between retries (partial download, process-level error). The panic message now points at the `PROTOC` env-var override for offline / locked-down environments.

2. **`.github/actions/cargo-cache/action.yml`** now persists `~/.cache/protoc` alongside the cargo registry cache. After the first warm cache per `(OS, lockhash)`, every subsequent job finds the already-extracted protoc binary at the path the build script checks first and never calls GitHub Releases at all. `CACHE_VERSION` bumped `v2 → v3` to force one clean rebuild that primes the new path layout; restore-keys bumped to match.

## Test plan

- `cargo check -p proto_generator` — clean locally.
- `cargo fmt -p proto_generator --check` — clean locally.
- CI: this PR's own run exercises both code paths. First run will be a cold cache (downloads protoc once via the new retry loop); subsequent runs on the same lockhash should show a cache hit and skip the protoc download entirely.
- Manual outage simulation (followup): block outbound DNS to `github.com` on a test runner and confirm the build still completes from the warmed `~/.cache/protoc`.

## Notes for reviewers

- Total worst-case stall ≈ 15s of outer backoff + curl's internal retry budget (~120s) — well under any CI job budget, and far cheaper than rerunning the whole release pipeline on a 504.
- Bumping `CACHE_VERSION` invalidates every existing Rust cache across the action's callers; expected one-time slow run.
- `~/.cache/protoc` path matches the Unix layout the build script writes to on every OS (Windows uses `home_dir()` → `USERPROFILE`; `actions/cache` normalises `~/` correctly across runners).

[SNOW-3571146]: https://snowflakecomputing.atlassian.net/browse/SNOW-3571146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/cargo-cache/action.yml
+++ b/.github/actions/cargo-cache/action.yml
@@ -87,11 +87,14 @@ runs:
     #   v1: initial cache action (Apr 17 2025)
     #   v2: aggressive slim + size cap + llvm-cov layout (Apr 29 2025) — busts
     #       the 13-15 GB tarpaulin caches that don't honour the 6 GB cap.
+    #   v3: include ~/.cache/protoc in the registry cache so the proto_generator
+    #       build script never has to hit GitHub Releases on a warm cache hit.
+    #       Bust v2 to force one clean rebuild that primes the new path layout.
     - name: Compute cache keys
       id: keys
       shell: bash
       run: |
-        CACHE_VERSION='v2'
+        CACHE_VERSION='v3'
         LOCK_HASH="${{ hashFiles(inputs.lockfiles) }}"
         echo "registry=${{ runner.os }}-cargo-registry-${CACHE_VERSION}-${LOCK_HASH}" >> $GITHUB_OUTPUT
         echo "target=${{ runner.os }}-rust-${{ inputs.shared-key }}-${CACHE_VERSION}-${{ steps.rustc.outputs.hash }}-${LOCK_HASH}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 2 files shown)


sfc-gh-akolodziejczyk (approved):
LGTM

sfc-gh-boler:
Force-pushed `boler/protoc-installer-retry` after the first CI run failed:

> manylinux_x86_64 job: `curl: option --retry-all-errors: is unknown` → exit 2 on every attempt → outer loop exhausted → panic.

Root cause: the manylinux build container ships an older curl (pre-7.71) that hard-errors on the `--retry-all-errors` flag I'd added. Dropped it. curl's default `--retry` already covers HTTP 408/429/500/502/503/**504** as documented transient classes, so the actual 504 failure mode is still absorbed inside a single curl invocation. The outer exponential-backoff loop continues to catch everything else (partial downloads, handshake errors, future classes). Net effect: same coverage for the 504 case, works on every runner (macOS-15, ubuntu-24, Windows-2022, manylinux_2_*) instead of just the new-curl ones.

Also documented a caveat for the cache pin: the manylinux job runs inside a Docker container that doesn't volume-mount the host's `~/.cache/protoc`, so it always falls back to the cold download path. Pre-baking protoc into the manylinux image is a sensible follow-up but out of scope here.

Description rewritten to match.

sfc-gh-boler:
**CI status note** — the ODBC test failures on this PR are not caused by these changes. They reproduce identically across reruns and are tracked separately as **[SNOW-3571259](https://snowflakecomputing.atlassian.net/browse/SNOW-3571259)**: the shared `CATALOGTESTS` schema on `sfctest0` is missing the `DESC*TABLE` fixtures defined in `scripts/odbc/setup_readonly_metadata_db.sql`, so every ODBC PR is red until that script is re-run against `sfctest0`.

Likely trigger: in-flight fixture work in #1291 (Devansh Das) ran the script with a WIP version that wiped + partially repopulated the schema. Not caused by, and not fixable from, this PR — its only Rust touch is `proto_generator/src/protoc_installer.rs` and one cache-key bump in `.github/actions/cargo-cache/action.yml`.

Assigned SNOW-3571259 to Devansh; the protoc retry-with-backoff change itself is reviewed-ready as soon as the ODBC schema is restored (or with admin override for the unrelated jobs).

[SNOW-3571259]: https://snowflakecomputing.atlassian.net/browse/SNOW-3571259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1271, opened 2026-05-21
SNOW-3406377, gap 13: expose put_get_max_attempts as a session-level retry knob

sfc-gh-pczajka:
Exposes `put_get_max_attempts` as a session-level connection parameter (default: 6, set to 1 to disable retries). Previously the retry limit was hard-coded.

The value is read from the connection seed at PUT/GET dispatch time, so `set_option` overrides take effect even after connection init. Out-of-range values (≤ 0 or > `u32::MAX`) fall back to the default.

The parameter is threaded through `perform_put_get_transfer` → `upload_files`/`download_files` → per-backend functions for S3, Azure, and GCS.

## Reference behavior across drivers

| Driver | Knob | Default | Semantics |
|---|---|---|---|
| JDBC | `putGetMaxRetries` | 25 | retries (loop runs ≤ N+1 times) |
| libsf / ODBC | `SF_CON_PUT_MAXRETRIES` / `PUT_MAXRETRIES` | 5 | retries, clamp `[0, 100]` |
| Python connector | hardcoded `max_retry: int = 5` (no knob) | 5 | attempts (loop runs < N times) |
| **UD core (this PR)** | `put_get_max_attempts` | 6 (S3, GCS, Azure) | attempts (`1 = no retry`) |

PR #1305, opened 2026-05-26
refactor(statement): dedupe post-execute pipeline into finalize_execute_result

sfc-gh-pczajka:
## Summary

Dedupes the post-execute pipeline in `statement.rs`. Stacked on PR #1271 (`feature/s3-max-retries-from-session`); once that lands, this can be retargeted to `main`.

`execute_query_internal` and `connection_get_query_result` both ran the same five-step post-response pipeline:

- `response_to_descriptor`
- `multistatement::try_into_multi_result` early-return
- `match data.command.as_deref()` — PUT/GET dispatch (lock + late-bound `put_get_max_attempts` / `use_s3_regional_url_session_param` read + `perform_put_get_transfer`) or `data.into_rowset_data()`
- `resolve_reader_ctx`
- `build_execute_result`

The only per-caller input was the optional `StageCredsRefreshContext` (`Some(ctx)` for the sync execute path, `None` for the async-fetch path). Extracted into `DatabaseDriverV1::finalize_execute_result`, so each call site shrinks to one line:

```rust
// execute_query_internal:
let stage_creds_refresh_context = StageCredsRefreshContext { sql, query_parameters, conn };
self.finalize_execute_result(&conn_arc, data, Some(stage_creds_refresh_context)).await

// connection_get_query_result body:
let data = fetch_query_response_data(&conn_ptr, &query_id).await?;
self.finalize_execute_result(&conn_ptr, data, None).await
```

The PUT/GET input read (`put_get_max_attempts`, `use_s3_regional_url_session_param`) stays inline inside `finalize_execute_result` — it's a single locked block with no independent reuse, so a separate helper would only exist as a test seam.

Net: −18 lines in `statement.rs` (+32 / −50). The post-execute pipeline now lives in one method; future tweaks (new dispatch shapes, additional late-bound inputs) only need to touch one site.

## Test plan

- [x] `cargo test -p sf_core --lib` — 1003 pass
- [x] `cargo clippy -p sf_core --tests --no-deps` clean
- [x] `cargo fmt -p sf_core --check` clean

### DIFF
--- a/sf_core/src/apis/database_driver_v1/statement.rs
+++ b/sf_core/src/apis/database_driver_v1/statement.rs
@@ -6,7 +6,7 @@ use super::connection::{Connection, RefreshContext, with_valid_session};
 use super::error::*;
 use super::global_state::DatabaseDriverV1;
 use super::multistatement;
-use super::query::perform_put_get_transfer;
+use super::query::{StageCredsRefreshContext, perform_put_get_transfer};
 use super::result_set::{
     ColumnMetadata, ExecuteQueryResult, fetch_query_response_data, resolve_reader_ctx,
     response_to_descriptor,
@@ -21,7 +21,8 @@ use crate::config::param_registry::{DEFAULT_PUT_GET_MAX_ATTEMPTS, param_names};
 use crate::config::settings::Setting;
 use crate::handle_manager::Handle;
 use crate::rest::snowflake::{
-    QueryExecutionMode, QueryInput, snowflake_abort_query, snowflake_query_with_client,
+    QueryExecutionMode, QueryInput, query_response, snowflake_abort_query,
+    snowflake_query_with_client,
 };
 
 use crate::config::rest_parameters::QueryParameters;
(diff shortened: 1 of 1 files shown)


sf_core/src/apis/database_driver_v1/statement.rs
```diff
@@ -344,6 +345,32 @@ impl DatabaseDriverV1 {
         drop(stmt);
 
         let data = response.data;
+        // Build refresh context for PUT/GET so the file manager can
+        // recover from STS `ExpiredToken` by re-issuing the original
+        // PUT/GET SQL to obtain fresh stage credentials. The refresher
+        // calls back into `RefreshContext::execute_with_refresh`, so a
+        // session-token expiry mid-batch is renewed transparently.
```
sfc-gh-pczajka: This is S3-specific comment, however the code is going to be shared between cloud providers. It should be generic

sfc-gh-pczajka:
work continued in mirror repo

PR #1306, opened 2026-05-26
SNOW-3166337: Implement batch parameter binding for Statement and PreparedStatement

sfc-gh-rkowalski:
Implements addBatch / executeBatch / clearBatch for Statement and PreparedStatement (previously SQLFeatureNotSupportedException stubs).

  - Statement-level: continue-on-error per JDBC §14; failed entries get EXECUTE_FAILED; any failure throws BatchUpdateException with the full-length array.
  - PreparedStatement-level: single array-bind RPC with SNOW-14034 expansion when the server count matches batchSize, else fan-out to SUCCESS_NO_INFO.

  Batch state and orchestration live in StatementBatch and PreparedBatch; the JDBC impls just delegate.

  getBatchQueryIDs() returns one entry per submitted row (Statement, null for failures) or one per array-bind batch (PS).

### DIFF
--- a/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedBatch.java
+++ b/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedBatch.java
@@ -1,0 +1,202 @@
+package net.snowflake.client.internal.api.implementation.statement;
+
+import java.sql.BatchUpdateException;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import net.snowflake.client.api.exception.ErrorCode;
+import net.snowflake.client.api.exception.SnowflakeSQLException;
+import net.snowflake.client.internal.api.implementation.statement.PreparedStatementBindingSerializer.ParameterValue;
+import net.snowflake.client.internal.log.SFLogger;
+import net.snowflake.client.internal.log.SFLoggerFactory;
+
+/**
+ * Column-major accumulation of parameter values across {@code addBatch()} calls. Each map entry
+ * holds a {@code List<String>} — one element per accumulated row. The entry's bind type is promoted
+ * from {@code "ANY"} to a real type when the first non-null value arrives in a column that has been
+ * all-nulls so far.
+ */
+final class PreparedBatch {
+  private static final SFLogger logger = SFLoggerFactory.getLogger(PreparedBatch.class);
+
+  private final Map<Integer, ParameterValue> columns = new HashMap<>();
+
+  /**
+   * Append a row built from the per-row {@code currentValues} map. Two-pass: every column is
+   * validated before any is mutated so a type-mismatch on a later column doesn't leave earlier
+   * columns at a different length.
+   */
+  void addRow(SqlPlaceholderMetadata meta, Map<Integer, ParameterValue> currentValues)
+      throws SQLException {
+    if (meta.hasMixedPlaceholderStyles()) {
+      throw new SnowflakeSQLException(
+          "Mixed positional and numeric placeholders are not supported");
+    }
+    for (int parameterIndex : meta.referencedParameterIndexes()) {
+      validate(parameterIndex, currentValues);
+    }
+    for (int parameterIndex : meta.referencedParameterIndexes()) {
+      commit(parameterIndex, currentValues);
+    }
+  }
+
+  void clear() {
+    columns.clear();
+  }
+
+  /** Number of accumulated rows; derived from any column's list size (every column same length). */
+  int size() {
+    if (columns.isEmpty()) {
+      return 0;
+    }
+    return ((List<?>) columns.values().iterator().next().value()).size();
+  }
+
+  boolean isEmpty() {
+    return size() == 0;
+  }
+
+  private Map<Integer, ParameterValue> snapshot() {
+    return new HashMap<>(columns);
+  }
+
+  /**
+   * Single-roundtrip array-bind execution. Returns one entry per row: per-row {@code 1} when the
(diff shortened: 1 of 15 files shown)


jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedBatch.java
```diff
@@ -0,0 +1,202 @@
...
+  void clear() {
+    columns.clear();
+  }
+
+  /** Number of accumulated rows; derived from any column's list size (every column same length). */
+  int size() {
+    if (columns.isEmpty()) {
+      return 0;
+    }
+    return ((List<?>) columns.values().iterator().next().value()).size();
```
sfc-gh-asolarski: maybe we should just track a rowCount to avoid this kind of checks and make each call cheeper?

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedBatch.java
```diff
@@ -0,0 +1,202 @@
...
+    return result;
+  }
+
+  private static int[] allFailed(int batchSize) {
+    int[] failed = new int[batchSize];
+    Arrays.fill(failed, Statement.EXECUTE_FAILED);
+    return failed;
+  }
+
+  private void validate(int parameterIndex, Map<Integer, ParameterValue> currentValues)
```
sfc-gh-asolarski: maybe this validation logic could be extracted into a separate class?

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
```diff
@@ -18,6 +19,14 @@ final class PreparedStatementBindingSerializer {
   private static final SFLogger logger =
       SFLoggerFactory.getLogger(PreparedStatementBindingSerializer.class);
 
+  /** Process-wide; a fresh allocator per execute is measurably expensive in batch scenarios. */
+  private static final RootAllocator SHARED_ALLOCATOR = new RootAllocator(Long.MAX_VALUE);
+
+  /** Test-only: outstanding bytes on the shared allocator (for {@code @AfterAll} leak checks). */
```
sfc-gh-asolarski: so maybe instead of exposing a test utility as a method, just make SHARED_ALLOCATOR package private? so it's accessible in test

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/SnowflakePreparedStatementImpl.java
```diff
@@ -472,8 +505,12 @@ private void setParameter(int parameterIndex, String bindType, Object value) thr
           placeholderMetadata.placeholderCount());
       return;
     }
+    // Boxed primitives passed through setObject must be stringified before they reach the
+    // serializer (which rejects non-String values).
+    String normalized = java.util.Objects.toString(value, null);
```
sfc-gh-asolarski: nit - just import Objects?

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/SnowflakeStatementImpl.java
```diff
@@ -81,18 +89,33 @@ protected int executeUpdateWithBindings(String sql, QueryBindings bindings) thro
     return getCurrentUpdateCountAsInt();
   }
 
-  protected boolean executeWithBindings(String sql, QueryBindings bindings) throws SQLException {
+  protected long executeLargeUpdateWithBindings(
+      String sql, PreparedStatementBindingSerializer.NativeBindings bindings) throws SQLException {
+    boolean producedResultSet = executeWithBindings(sql, bindings);
+    if (producedResultSet) {
```
sfc-gh-asolarski, line 105: is this a sensible check? we called `executeWithBindings` that already executed the statement, if it fails than it will be forwarded - why to check it here?

sfc-gh-asolarski (approved):
minor remarks

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/SnowflakeStatementImpl.java
sfc-gh-rkowalski: This is part of JDBC spec :/ `snowflake-jdbc` behaves same way and we should maintain parity
> [SQLException](https://docs.oracle.com/javase/8/docs/api/java/sql/SQLException.html) - if a database access error occurs, this method is called on a closed Statement, **the given SQL statement produces a ResultSet** object, the method is called on a PreparedStatement or CallableStatement

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/SnowflakePreparedStatementImpl.java
sfc-gh-rkowalski: fixed

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedStatementBindingSerializer.java
sfc-gh-rkowalski: fixed

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedBatch.java
sfc-gh-rkowalski: extracted

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/statement/PreparedBatch.java
sfc-gh-rkowalski: fixed

PR #1310, opened 2026-05-26
SNOW-2881687: Expose lastQueryId in API

sfc-gh-rkowalski:
## Summary
  - Implement the `getQueryID()` API on `SnowflakeStatement`, `SnowflakeResultSet`, and `SnowflakeResultSetMetaData`, mirroring the contract from the legacy snowflake-jdbc driver.
  - Surface the server-side query ID on `SnowflakeSQLException` (sourced from `DriverException.query_id` in the proto layer) so callers can correlate failed queries with the Snowflake query history.
  - Preserve the failed query's ID on the parent `Statement` so `getQueryID()` returns it after an exception, matching the legacy `setQueryIdWhenValidOrNull` semantics.

  ## Behavior
  - **Fresh statement** — `getQueryID()` returns `null`.
  - **Successful execute** — overwrites with the new query ID; same ID is reachable via the producing `Statement`, `ResultSet`, and `ResultSetMetaData`.
  - **Failed execute with server-side ID** — `Statement.getQueryID()` returns the failed query's ID; equal to `((SnowflakeSQLException) ex).getQueryId()`.
  - **Failed execute without server-side ID** (e.g. transport error) — any prior stale ID is wiped; `getQueryID()` returns `null`.
  - **Multi-statement** — `Statement.getQueryID()` returns the parent batch ID; each child `ResultSet` exposes its own per-child ID.
  - **Batch** — `getBatchQueryIDs()` is positionally aligned with `executeBatch()` update counts; failed slots can carry a server-side ID when one was returned.

  ## Files
  **Production**
  - `SnowflakeSQLException` — new `queryId` field + `getQueryId()` getter; populated from the proto.
  - `SnowflakeStatementImpl` — `executeStatement` catches `SQLException`, calls `captureQueryIdFromException` to preserve the server-side ID before rethrowing; result-set construction now passes the query ID at
  construction (final field, captured eagerly so a later execute on the same statement doesn't contaminate prior result sets).
  - `SnowflakeResultSetImpl` — final `queryId` field, surfaced by `getQueryID()`, propagated into `getMetaData()`.
  - `SnowflakeResultSetMetaDataImpl` — final `queryId` field, surfaced by `getQueryID()`.

### DIFF
--- a/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/resultset/SnowflakeResultSetImpl.java
+++ b/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/resultset/SnowflakeResultSetImpl.java
@@ -1146,7 +1146,6 @@ public class SnowflakeResultSetImpl implements ResultSet, SnowflakeResultSet {
 
   @Override
   public String getQueryID() throws SQLException {
-    checkClosed();
     return queryId;
   }
 
--- a/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/resultset/SnowflakeResultSetImpl.java
+++ b/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/resultset/SnowflakeResultSetImpl.java
@@ -46,6 +46,7 @@ import org.apache.arrow.vector.VectorSchemaRoot;
 public class SnowflakeResultSetImpl implements ResultSet, SnowflakeResultSet {
 
   private final SnowflakeStatementImpl statement;
+  private final String queryId;
   private final CursorState cursor = new CursorState();
   private final SchemaState schema;
   private final ArrowResources resources;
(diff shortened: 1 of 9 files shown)


jdbc/src/main/java/net/snowflake/client/internal/api/implementation/resultset/SnowflakeResultSetImpl.java
```diff
@@ -1142,7 +1146,8 @@ private void checkColumnIndex(int columnIndex) throws SQLException {
 
   @Override
   public String getQueryID() throws SQLException {
-    throw new NotImplementedException();
+    checkClosed();
```
sfc-gh-asolarski: can't we return query id even for closed result set? we have this information
Is it a valid case that someone would like to free the resources, but keep this metadata for some reason?

jdbc/src/test/java/net/snowflake/client/api/statement/LastQueryIdTests.java
sfc-gh-asolarski, line 1: hmm, we keep it under jdbc/e2e/query, but it's not backed with gherkin files, maybe it;s fine but maybe we should keep it inside (jdbc/src/test/java/net/snowflake/client/api/statement, where we have other integ (but really e2e) tests?)

I think we should define clearly this test structure

sfc-gh-asolarski (approved):

sfc-gh-asolarski:
nothing major

jdbc/src/main/java/net/snowflake/client/internal/api/implementation/resultset/SnowflakeResultSetImpl.java
sfc-gh-rkowalski: Good catch. Fixed.

jdbc/src/test/java/net/snowflake/client/api/statement/LastQueryIdTests.java
sfc-gh-rkowalski: I've moved this to `api` package since it does not belong to *.feature tests

PR #1312, opened 2026-05-26
SNOW-3406377, gap 4, part 1: streaming S3 PUT/GET (internal refactor)

sfc-gh-pczajka:
## Summary

Internal refactor of the S3 PUT/GET file transfer path to lay the foundation
for gap 4 (streaming I/O) and gap 16 (whole-file `Vec<u8>` in memory).
No public API change for existing callers — `upload_files`, `upload_single_file`,
`download_files`, and `download_single_file` have unchanged signatures.

### What changed

- **`ByteSource` enum** (`Path(PathBuf)` | `Bytes(Vec<u8>)`) in
  `file_manager/types.rs`. `SingleUploadData` now carries `source: ByteSource`
  + `source_path_str: String` instead of a bare `file_path: String`.
  `PreparedUpload.data` changed from `Vec<u8>` to `ByteSource`.

- **Streaming encryption** (`file_manager/encryption.rs`): `encrypt_file_data`
  streams plaintext through `openssl::symm::Crypter` in 64 KiB chunks; SHA-256
  digest is computed over the ciphertext in the same pass. The old
  `decrypt_file_data(&[u8]) -> Result<Vec<u8>>` is replaced by
  `decrypt_ciphertext_to_writer<R: Read, W: Write>` which writes decrypted
  plaintext directly to an output `Write` in 64 KiB chunks — the plaintext is
  never buffered as a full `Vec<u8>`.

- **S3 PUT streaming** (`file_manager/s3_transfer.rs`): `ByteSource::Path`
  sources are uploaded via `ByteStream::read_from().path()` (no intermediate
  `Vec` copy); `ByteSource::Bytes` sources use `ByteStream::from(bytes)`.

- **Compression auto-detect** (`file_manager/mod.rs`): reads only a 16-byte
  prefix from the source (was the full file).

- **`compress_reader<R: Read>()` ** (`compression.rs`): streaming gzip wrapper
  with `mtime=0`, bit-identical to `compress_data()`.

- **GCS/Azure unchanged** for this PR: both call `ByteSource::into_bytes()` and
  continue to work with an in-memory `Vec<u8>`. Their streaming refactor is PR-2.

### Behavioral change (documented)

**Digest verification timing** — the old `decrypt_file_data` verified the
SHA-256 digest *eagerly* over the full in-memory ciphertext **before**
decryption began. Under the new streaming path, the digest is computed over
ciphertext as it flows through and verified **at finalize time** (after all
bytes have been decrypted and written to the output writer). If
`EncryptionError::DigestMismatch` is returned, partial plaintext may already
have been written to the destination file — callers are expected to discard
the partial output (e.g. delete the file).

This change was unavoidable: pre-verification requires buffering the full
ciphertext, which contradicts the goal of eliminating the full-file buffer.
The corruption detection guarantee is preserved; only the *timing* differs.

### Public types for PR-3 / PR-4

Wrapper layers that supply in-memory buffers (JDBC `uploadStream`, Python
`file_stream`) should construct:

```rust
SingleUploadData {
    source: ByteSource::Bytes(data),
    source_path_str: "<display name>".to_string(),
    // ... other fields unchanged
}
```

The encrypt/decrypt functions are exposed at
`sf_core::file_manager::internal::{encrypt_file_data, decrypt_ciphertext_to_writer}`
under `#[cfg(any(test, feature = "test-utils"))]`. For production call sites
use `upload_single_file` / `download_single_file` as before.

## Test plan

- [x] `cargo clippy -p sf_core` — no warnings or errors
- [x] `cargo test -p sf_core --test byte_source_roundtrip` — 3 tests pass:
  - `bytes_source_encrypt_decrypt_roundtrip` (small payload, full round-trip)
  - `bytes_source_encrypt_decrypt_large_payload` (256 KiB, exercises multiple
    64 KiB Crypter chunks)
  - `bytes_source_decrypt_detects_tampered_digest` (tampered digest yields
    `DigestMismatch`)
- [x] `cargo test -p sf_core --lib` — all existing unit tests pass
- [x] `cargo build --workspace` — clean build

### DIFF
--- a/sf_core/Cargo.toml
+++ b/sf_core/Cargo.toml
@@ -161,6 +161,10 @@ name = "logging_query_tests"
 path = "tests/logging_query_tests.rs"
 
 [[test]]
+name = "byte_source_roundtrip"
+path = "tests/byte_source_roundtrip.rs"
+
+[[test]]
 name = "e2e_tests"
 path = "tests/e2e/mod.rs"
 
--- a/sf_core/src/compression.rs
+++ b/sf_core/src/compression.rs
@@ -1,6 +1,6 @@
 use flate2::{Compression, GzBuilder, bufread::GzDecoder};
 use snafu::{Location, ResultExt, Snafu};
-use std::io::{Read, Write};
+use std::io::{BufReader, Read, Write};
 
 // PUT/GET compression
 pub fn compress_data(input_data: Vec<u8>) -> Result<Vec<u8>, CompressionError> {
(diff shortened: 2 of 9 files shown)


sfc-gh-pczajka:
work continued in mirror repo

PR #1315, opened 2026-05-26
SNOW-3406377, gap 4, part 3: jdbc uploadStream/downloadStream API

sfc-gh-pczajka:
## Summary

- **RPC shape**: Option A — two new connection-level RPCs (`ConnectionUploadStream` / `ConnectionDownloadStream`) that accept raw bytes across the JNI boundary, synthesize a `PUT`/`GET` SQL to obtain stage credentials and encryption material from GS, then delegate to the existing `upload_single_file` / `download_single_file` helpers with `ByteSource::Bytes(data)` (added by PR-1). No new stage-credential plumbing was needed; the existing `SnowflakeStageCredsRefresherPub` (renamed from the private `SnowflakeStageCredsRefresher` in `query.rs`) covers credential refresh during long transfers.

- Replaces the two `NotImplementedException` stubs in `SnowflakeConnectionImpl.uploadStream` / `downloadStream` with working implementations that read the caller's `InputStream` into a `byte[]`, forward to the core RPC, and wrap the returned bytes in a `ByteArrayInputStream`.

- Four commits: (1) proto schema + `Cargo.toml` dependency, (2) Rust core `file_transfer.rs`, (3) Java API layer (`CoreDriverApi` / `CoreDriverApiImpl`), (4) JDBC wiring + unit tests.

## Behavioral changes vs. reference snowflake-jdbc

| Area | Reference JDBC | This implementation |
|------|---------------|---------------------|
| `compressData` default | `true` (gzip + `.gz` suffix) | Same — when `config == null` or `config.isCompressData()` is unset, defaults to `true` |
| `destPrefix` handling | `destPrefix + "/" + destFileName` when non-empty | Same — `build_dest_filename` mirrors this exactly |
| Overwrite semantics | Always overwrites for stream uploads | Same — `OVERWRITE = TRUE` in synthesized `PUT` |
| `decompress` default | Caller-controlled | `false` when `config == null`; caller must opt in |
| Stream ownership | Caller closes the `InputStream` | Same — we only read; never close |
| `downloadStream` return type | `InputStream` (may be streaming for large files) | `ByteArrayInputStream` wrapping fully materialised bytes |

The only observable difference from reference JDBC is that the `InputStream` returned by `downloadStream` is a `ByteArrayInputStream` (fully materialised in memory). For the gap 4 / gap 16 use-cases this is the defined contract.

## Test plan

- [x] `cargo test -p sf_core --lib` — 1004 tests pass (includes `build_dest_filename` / `build_stage_path` unit tests in `file_transfer.rs`)
- [x] `UploadDownloadStreamTest` (8 tests) — default config forwards `compressData=true`; `destPrefix` forwarded; `compressData=false` forwarded; all InputStream bytes read; download returns correct bytes; `decompress=true` forwarded; `IOException` wrapped as `SQLException`; null config handled
- [x] `SnowflakeConnectionImplTest` (19 tests) — no regressions
- [ ] Real end-to-end integration test against a Snowflake account — deferred to separate test pass per gap-4 spec

### DIFF
--- a/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/SnowflakeConnectionImpl.java
+++ b/jdbc/src/main/java/net/snowflake/client/internal/api/implementation/connection/SnowflakeConnectionImpl.java
@@ -523,14 +523,31 @@ public class SnowflakeConnectionImpl implements InternalSnowflakeConnection {
   @Override
   public void uploadStream(String stageName, String destFileName, InputStream inputStream)
       throws SQLException {
-    throw new NotImplementedException();
+    uploadStream(stageName, destFileName, inputStream, UploadStreamConfig.builder().build());
   }
 
   @Override
   public void uploadStream(
       String stageName, String destFileName, InputStream inputStream, UploadStreamConfig config)
       throws SQLException {
-    throw new NotImplementedException();
+    checkClosed();
+    byte[] data;
+    try {
+      java.io.ByteArrayOutputStream baos = new java.io.ByteArrayOutputStream();
+      byte[] buf = new byte[8192];
+      int n;
+      while ((n = inputStream.read(buf)) != -1) {
+        baos.write(buf, 0, n);
+      }
+      data = baos.toByteArray();
+    } catch (java.io.IOException e) {
+      throw new net.snowflake.client.api.exception.SnowflakeSQLException(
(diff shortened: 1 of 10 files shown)


sfc-gh-pczajka:
work continued in mirror repo

PR #1303, opened 2026-05-26
fix protoc cache to include arch

sfc-gh-akolodziejczyk:
fixes https://snowflake.slack.com/archives/C0B544MNQTD/p1779781705673369?thread_ts=1779781418.927019&cid=C0B544MNQTD

PR #1317, opened 2026-05-26
SNOW-3573633: fix cargo-cache arch-poisoning on macOS

sfc-gh-boler:
Tracking: [SNOW-3573633](https://snowflakecomputing.atlassian.net/browse/SNOW-3573633)  ·  Regression from: [SNOW-3571146](https://snowflakecomputing.atlassian.net/browse/SNOW-3571146) / #1299  ·  Triage: [slack thread](https://snowflake.slack.com/archives/C0B544MNQTD/p1779812099001669)

## Why

`main` is red on Python Release Build (commit `c6e487f6`) with:

```
thread 'main' panicked at sf_core/build.rs:37:10:
Failed to generate protobuf Rust code: ... failed to invoke protoc
(path: /Users/runner/.cache/protoc/v32.1/bin/protoc):
Bad CPU type in executable (os error 86)
```

#1299 added `~/.cache/protoc` to the cargo registry cache to eliminate the GitHub Releases dependency on warm cache hits. The cache key was keyed on `runner.os`, which GitHub Actions reports as `'macOS'` for **both** the `macos-*-x86_64` and `macos-*-arm64` runners. Before #1299 this was harmless (registry contents are arch-agnostic source). After #1299 the cache also contained the arch-specific protoc binary — whichever arch saved first poisoned the cache for the other. The x86_64 release_build runner restored an arm64 protoc, exec'd it, errno 86.

Failing run: https://github.com/snowflakedb/universal-driver/actions/runs/26459928982

## What

One-file change in `.github/actions/cargo-cache/action.yml`:

1. Add `${{ runner.arch }}` to both the registry and target cache keys.
2. Update the two `restore-keys` blocks to match (still hard-coded literals, per the file's existing convention).
3. Bump `CACHE_VERSION` `v3 → v4` to evict the poisoned `v3` caches in one shot. History note added explaining the bug so a future bump can avoid the same rake.

Verified no other workflow file references the old key shape (grep across `.github/`).

Target cache was technically already safe (`rustc.outputs.hash` encodes the host triple), but adding `runner.arch` is consistent, defensive against future cross-compile scenarios, and free.

## Alternative considered

Drop `~/.cache/protoc` from the cache and rely on the curl retry hardening that was the actual goal of #1299. Rejected because the cache eliminates the GitHub Releases dependency entirely on warm cache hits, which is real value — only the key shape was wrong.

## Test plan

- [ ] CI green on this PR across all runners (one-time cache-miss penalty as `v3` is evicted).
- [ ] Post-merge: rerun `main` Python Release Build to confirm `build_wheels (macosx_x86_64)` no longer hits errno 86.
- [ ] Spot-check a subsequent run to confirm warm `v4` cache restores on both `macos-x86_64` and `macos-arm64` independently.

[SNOW-3573633]: https://snowflakecomputing.atlassian.net/browse/SNOW-3573633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SNOW-3571146]: https://snowflakecomputing.atlassian.net/browse/SNOW-3571146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/cargo-cache/action.yml
+++ b/.github/actions/cargo-cache/action.yml
@@ -90,14 +90,28 @@ runs:
     #   v3: include ~/.cache/protoc in the registry cache so the proto_generator
     #       build script never has to hit GitHub Releases on a warm cache hit.
     #       Bust v2 to force one clean rebuild that primes the new path layout.
+    #   v4: add runner.arch to both cache keys. v3 became an active footgun on
+    #       macOS: GitHub Actions reports runner.os='macOS' for both the x86_64
+    #       and arm64 runners, so they shared a single registry cache key. When
+    #       v3 also began caching the arch-specific protoc binary at
+    #       ~/.cache/protoc/<ver>/bin/protoc, whichever arch saved first
+    #       poisoned the cache for the other arch — the x86_64 runner restored
+    #       the arm64 binary and exec'd it, producing `Bad CPU type in
+    #       executable (os error 86)` and a red main build (see SNOW-3571146
+    #       resolution comment). Adding runner.arch (Linux: X64/ARM64, macOS:
+    #       X64/ARM64, Windows: X64/ARM64) segregates the caches per HOST
+    #       arch. (Cross-target builds on the same host — e.g. building
+    #       aarch64-linux on an x86_64 runner — still need a distinct
+    #       shared-key from the caller to keep their target/ caches apart.)
+    #       Bump v3 → v4 to evict the poisoned v3 caches in one shot.
     - name: Compute cache keys
       id: keys
       shell: bash
       run: |
-        CACHE_VERSION='v3'
+        CACHE_VERSION='v4'
         LOCK_HASH="${{ hashFiles(inputs.lockfiles) }}"
-        echo "registry=${{ runner.os }}-cargo-registry-${CACHE_VERSION}-${LOCK_HASH}" >> $GITHUB_OUTPUT
-        echo "target=${{ runner.os }}-rust-${{ inputs.shared-key }}-${CACHE_VERSION}-${{ steps.rustc.outputs.hash }}-${LOCK_HASH}" >> $GITHUB_OUTPUT
+        echo "registry=${{ runner.os }}-${{ runner.arch }}-cargo-registry-${CACHE_VERSION}-${LOCK_HASH}" >> $GITHUB_OUTPUT
+        echo "target=${{ runner.os }}-${{ runner.arch }}-rust-${{ inputs.shared-key }}-${CACHE_VERSION}-${{ steps.rustc.outputs.hash }}-${LOCK_HASH}" >> $GITHUB_OUTPUT
 
     # ── Restore mode ──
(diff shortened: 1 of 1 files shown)


sfc-gh-boler:
## Independent review summary

Ran this through a staff-engineer reviewer. Verdict: 🟡 ready after addressing two minor items; no blockers.

**Inline edits applied** (and force-pushed):

- Tightened the v4 history-note in `cargo-cache/action.yml` and the commit body to clarify that `runner.arch` segregates by **host** arch only — cross-target builds on the same host still need a distinct `shared-key` from the caller. The earlier "defensive against future cross-compile scenarios" phrasing was half-true.

**Follow-up filed** as **[SNOW-3574311](https://snowflakecomputing.atlassian.net/browse/SNOW-3574311)** (Task, Medium):

- The reviewer surfaced **4 other workflow files with the same `runner.os`-only cache-key shape** (`test-python.yml` ×2 keys, `_build-python-wheels.yml` ×2 keys, `test-rust-core.yml` cargo-llvm-cov cache). They're safe today only because each runs a single arch per OS — the moment any of them grows ARM-Linux coverage, the same errno-86 failure reproduces. Tracked as a separate mechanical-fix ticket so this PR stays a tight one-file regression fix.

**Architectural follow-ups noted in SNOW-3574311** (not blocking, not in scope here):
- Factor the cache-key prefix into composite-action outputs so `restore-keys` doesn't duplicate the literal (the file currently documents the duplication as intentional — that design choice is worth reopening).
- Give `~/.cache/protoc` its own cache step keyed on the protoc version constant rather than `Cargo.lock` — would eliminate the bug class structurally instead of just preventing the latest instance.

The commit message + PR body got an "above-average" rating; remaining nits (subject coined-term, optional one-line warning in `protoc_installer.rs::cache_dir`) intentionally deferred.

[SNOW-3574311]: https://snowflakecomputing.atlassian.net/browse/SNOW-3574311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

sfc-gh-rsavenok (approved):

sfc-gh-boler:
Closing — duplicate of #1303, which @sfc-gh-akolodziejczyk opened ~9 hours before this PR. I didn't see his open PR when I filed (didn't search for prior art on `cargo-cache/action.yml` — process gotcha for next time).

Both fixes are functionally equivalent for the actual bug. His is more surgical (only the registry cache key, where `~/.cache/protoc` actually lives). My extra `runner.arch` on the target cache key was redundant since `rustc.outputs.hash` already varies per host.

I've moved the SNOW-3573633 link and the SNOW-3574311 audit follow-up over to #1303 via comment.

PR #1324, opened 2026-05-27
SNOW-3406389: disable reqwest auto-gunzip on GCS storage client

sfc-gh-mhofman:
## Summary

- Add `.no_gzip()` to `create_gcs_client` in `sf_core/src/file_manager/gcs_transfer.rs` so reqwest no longer auto-decompresses GCS GET responses that carry `Content-Encoding: gzip`. This eliminates silent corruption of CSE-encrypted downloads and of any stage object written by external tooling (`gsutil cp -Z`, BigQuery exports, customer ETL) that tags the object with gzip while the body is the natural-format payload.
- One-line behavioral change, scoped to the GCS storage HTTP client. No public API, protobuf, binding (Python / ODBC / JDBC), or `Cargo.toml` change. The shared TLS client (`tls/client.rs`), the control-plane / telemetry / OAuth clients, the Azure SAS client (`azure_transfer.rs`), and the CRL fetcher (`crl/cache.rs`) are intentionally untouched — gzip on control-plane JSON paths is still desirable.
- The upload-side `content-encoding: ""` strip in `upload_to_gcs` is unchanged and remains the PUT-side complement; driver-only round-trips are unaffected.

## Context

Design contract: [`--gcp--/2.6-response_gzip_workaround.md`](../blob/main/--gcp--/2.6-response_gzip_workaround.md). Master plan: [`--gcp--/IMPLEMENTATION_PLAN.md`](../blob/main/--gcp--/IMPLEMENTATION_PLAN.md) — this is Step 3 of the GCS PUT/GET gap-closure stack under `SNOW-3406389`.

**Why this must land before steps 4 and 5.** With auto-gunzip on, `Content-Length` (gap 2.5) compares the compressed wire size to the decompressed body length and fires spuriously, and the SHA-256 digest in `x-goog-meta-sfc-digest` (gap 2.3) is computed over stored bytes that no longer match what `download_from_gcs` sees. This step removes both blast-radius interactions before the dependent gaps land.

**Parity with reference drivers.**
- JDBC: `HttpUtil.disableContentCompression()` (`HttpUtil.java:420`), wired in via `SnowflakeGCSClient.java:237,:432` — direct analog.
- libsfclient: default libcurl behavior is no auto-decompression; the `CURLOPT_ACCEPT_ENCODING` opt-in at `http_perform.c:485` is gated to the chunk downloader, never to GCS file transfers.
- Python: `remove_content_encoding` urllib3 hook (`storage_client.py:54-59`) — the intent we mirror, though Python's connector installs the hook only on S3. sf_core now has the JDBC-strength defense.

## Test plan

```bash
cargo fmt --all -- --check                            # clean
cargo clippy -p sf_core --all-targets -- -D warnings  # clean
cargo test  -p sf_core --lib                          # 998 passed, 1 ignored
cargo test  -p sf_core --lib file_manager::gcs_transfer   # 36 passed
cargo test  -p sf_core --test integration_tests -- http:: # 34 passed
cargo test  -p sf_core --test integration_tests -- http::gcs   # 10 passed (6 existing + 4 new)
cargo test  -p sf_core --lib rest::snowflake::telemetry        # 3 passed
```

Four new wiremock-backed integration tests in `sf_core/tests/integration/http/gcs_retry.rs`:

| Test | Asserts |
|---|---|
| `gcs_download_content_encoding_gzip_with_non_gzip_body_is_returned_verbatim` | Response with `Content-Encoding: gzip` header and a non-gzip body succeeds and returns the wire bytes byte-for-byte (would error with "Invalid gzip header" before the fix). |
| `gcs_download_content_encoding_gzip_with_gzip_body_is_not_decoded` | Response with a valid gzip body and `Content-Encoding: gzip` returns the *compressed* wire bytes — positive byte-equality that proves the auto-decoder is off (would return the decoded payload before the fix). |
| `gcs_download_without_content_encoding_header_is_unchanged` | Happy-path regression guard — body returned verbatim when no `Content-Encoding` is present. |
| `gcs_download_does_not_advertise_gzip_accept_encoding` | The outbound `Accept-Encoding` header does not contain `gzip` (would contain `gzip` before the fix; `.no_gzip()` also suppresses reqwest's auto-injection). Guards against someone replacing `.no_gzip()` with `.gzip(true)` without also updating the body-side assertions. |

**Negative-case isolation check (per Step 3 prompt §8.5):** temporarily removed `.no_gzip()` from `create_gcs_client` and confirmed all three new no-gzip tests fail in the predicted way — one with a `reqwest::Error { kind: Decode, source: "Invalid gzip header" }`, one with a byte-mismatch between gzipped wire bytes and decoded payload, one with `accept-encoding: "gzip"` advertised on the wire. Restored the fix; full suite green again.

**Existing-test regression sweep:** `http::gcs_retry::*` (6 tests covering 401/403/400/404/503 retry paths) and `rest::snowflake::telemetry::*` (the `accept-encoding: gzip, deflate` assertion at `telemetry.rs:153` — proves the gzip feature stays globally enabled and only the GCS client opts out) all still pass.

## Drift report

Confirmed against the spec at the time of this PR:

- `sf_core/Cargo.toml:32` still declares `reqwest = { version = "0.12.23", features = ["json", "rustls-tls", "gzip"] }`. The gzip feature is on — `telemetry.rs:153` asserts `accept-encoding: gzip, deflate` is emitted live — so this is an **active** gap, not theoretical. The condition under which the prompt allowed a downgrade to a defensive `debug_assert!` does **not** apply; the full fix is shipped.
- `sf_core/src/file_manager/gcs_transfer.rs` `create_gcs_client` is at lines 284-289 as cited.
- `sf_core/src/file_manager/gcs_transfer.rs` upload-side `.header("content-encoding", "")` is at line 193 as cited; left untouched.
- Storage-client audit: `create_gcs_client` is the only reqwest builder in `sf_core/src/**` that touches GCS storage URLs. `connection.rs:2493` is a test helper for the control-plane client; `tls/client.rs` builds the shared TLS client (not used by GCS today); `crl/cache.rs:462` is the CRL fetcher; `azure_transfer.rs:254` is Azure (out of scope per spec §6 and §8 — tracked separately). No shared-client cross-bleed risk.
- Option taken: **Option A** (`.no_gzip()` on the storage `ClientBuilder`) per spec §5.4 recommendation. The reqwest 0.12.23 `ClientBuilder::no_gzip` symbol is present; the fallback to `.gzip(false) + Accept-Encoding: identity` (option B) was not needed.
- No `bindings/` changes (verified by inspecting the call chain `download_from_gcs` → `cursor.execute("GET …")`).

## Review deferrals

- **Arctic Owl rule-based review (`sf ai review run`):** clean — 0 comments, no eligible findings on the 2 changed files.
- **Arctic Owl agentic / Claude reviewer:** could not run from the agent's non-interactive shell because the local Snowhouse OAuth token is expired and `sfid` refresh requires a TTY for Okta. The change is one line of behavioral diff (`.no_gzip()`) plus a docstring and four wiremock tests; the rule-based pass, the manual self-review (scope, comment quality, negative-case isolation, parity with JDBC / libsfclient / Python) and the green regression sweep cover the residual risk. **No High/Medium findings deferred.**
- **ODBC-test reviewer:** N/A — no `bindings/odbc/**` changes (verified with `git diff main -- bindings/`, empty).

### DIFF
--- a/sf_core/src/file_manager/gcs_transfer.rs
+++ b/sf_core/src/file_manager/gcs_transfer.rs
@@ -284,6 +284,18 @@ fn map_http_error(e: HttpError) -> GcsRequestError {
 fn create_gcs_client() -> Result<reqwest::Client, GcsRequestError> {
     reqwest::Client::builder()
         .timeout(Duration::from_secs(REQUEST_TIMEOUT_SECS))
+        // Disable reqwest's auto-gzip path so a GCS response carrying
+        // `Content-Encoding: gzip` (typically set by external loaders such
+        // as `gsutil cp -Z` or BigQuery exports) is handed to the caller
+        // verbatim. The driver is moving opaque, possibly CSE-encrypted
+        // bytes, and downstream SHA-256 digest / Content-Length checks
+        // assume wire bytes == body bytes. Mirrors JDBC's
+        // `HttpUtil.disableContentCompression()`
+        // (`SnowflakeGCSClient.java:237,:432` via `HttpUtil.java:420`) and
+        // the intent of Python's `remove_content_encoding` urllib3 hook
+        // (`storage_client.py:54-59`); the upload-side `content-encoding`
+        // strip in `upload_to_gcs` is the matching PUT-side defense.
+        .no_gzip()
         .build()
         .map_err(|source| GcsRequestError::Http { source })
 }
--- a/sf_core/tests/integration/http/gcs_retry.rs
+++ b/sf_core/tests/integration/http/gcs_retry.rs
@@ -1,5 +1,8 @@
+use flate2::Compression;
+use flate2::write::GzEncoder;
 use sf_core::file_manager::{CloudCredentials, LocationType, StageInfo};
 use sf_core::sensitive::SensitiveString;
+use std::io::Write;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicU32, Ordering};
 use wiremock::matchers::{method, path};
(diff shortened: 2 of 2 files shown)


sfc-gh-mhofman:
* **#1324** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1324?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1324?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1326, opened 2026-06-08
feat(sf_core): add public RecordBatchReader API

zeroshade:
## Summary

Exposes a query result's Arrow data as a native `RecordBatchReader` from `sf_core`
without forcing an FFI round-trip through the core API. `result_set_get_stream`
previously returned an Arrow C-stream (`Box<FFI_ArrowArrayStream>`); it now returns
`Box<dyn RecordBatchReader + Send>` directly, and the `FFI_ArrowArrayStream` wrapping is
done in the protobuf layer at the C boundary.

## What changed

- `DatabaseDriverV1::result_set_get_stream(handle)` now returns
  `Box<dyn RecordBatchReader + Send>`, built lazily from the result set's stored
  `RowsetData` (so it can be requested multiple times).
- The Arrow C-stream conversion moved into the protobuf layer. A single
  `reader_to_arrow_stream_ptr` helper in the `database_driver_v1` converter wraps the
  reader in an `FFI_ArrowArrayStream`, shared by the `result_set_get_stream` and
  `statement_prepare` proto handlers. The proto wire contract
  (`ResultSetGetStreamResponse` and the generated client methods) is unchanged, so ODBC
  and the perf-test app are unaffected.
- `PrepareResult.stream` now carries the native reader; the FFI pointer is produced in
  the `statement_prepare` proto handler.

Rust consumers can obtain the reader without any FFI:

```rust
let exec = driver.statement_execute_query(stmt_handle, None, None).await?;
if let ExecuteQueryResult::Single(rs) = exec {
    let reader = driver.result_set_get_stream(rs.handle).await?; // Box<dyn RecordBatchReader + Send>
    // ... drain (see contract below)
}
```

## Contract

Awaiting `result_set_get_stream` only builds the reader and never blocks. **Iterating**
it, however, must happen in a synchronous context: for chunked result sets the reader
pulls chunks lazily via a blocking channel receiver, so draining from within an async
runtime (a `tokio` task or `block_on`) would call `blocking_recv` and panic. Drain it
after returning from `block_on` (keeping the runtime alive), on a dedicated
`std::thread`, or via `tokio::task::spawn_blocking`.

## Tests

In-crate tests (no network):
- `result_set_get_stream_returns_reader_without_ffi` — builds a result set from an
  inline JSON rowset and asserts the reader's schema and rows.
- `result_set_get_stream_reads_arrow_rowset` — drives an inline Arrow IPC rowset
  (`ArrowMultiChunk`) through the `PrefetchChunkReader` `blocking_recv` path, covering
  the Arrow format alongside JSON.

## Dependency change

Raises the `arrow` / `arrow-ipc` / `arrow-array` floor to `>=54.0.0, <59` in `sf_core`,
`odbc`, and the core perf-test app (`53.x` fails to build against `chrono 0.4.43`).
Intentional trade-off: `sf_core` exposes Arrow types publicly, so consumers using them
must keep their `arrow` major aligned with the version `sf_core` resolves.

## Notes

- Merged latest `main`, adapting to the upstream refactor that split the query path into
  a `result_set` module (the reader API is a method on `DatabaseDriverV1` rather than a
  free `process_query_response`).


zeroshade:
Addressing the Copilot review feedback. Note the branch was merged with latest `main` after some of these were filed, and upstream's refactor moved the query path into a `result_set` module — so a couple of comments now point at code that no longer exists.

### Resolved by the merge / redesign
- **`QueryResult` marked `#[non_exhaustive]` (`query.rs:41`)** — `QueryResult` and `process_query_response` no longer exist; upstream replaced them with the `result_set` module. The public reader API is now the method `DatabaseDriverV1::result_set_get_reader`, so there is no `#[non_exhaustive]` public struct and no struct-literal break.
- **Schema-only test draining inside `#[tokio::test]` (`query.rs:560`)** — that test was removed. Its replacement, `result_set_get_reader_returns_reader_without_ffi` (in `result_set.rs`), drains from a **synchronous** context: it obtains the reader inside `block_on`, then drains after returning while keeping the runtime alive — exactly the pattern the comment recommended.

### `result_set_get_reader` (`result_set.rs`)
This method is the non-FFI twin of the existing `result_set_get_stream` immediately above it and mirrors it deliberately:
- **Mutex guard held across `.await` (`:403`)** — identical to `result_set_get_stream`. The guard is per-result-set-handle and the awaited `build_reader_from_rowset_data` never re-acquires it, so there's no deadlock — only serialization of concurrent builds on the *same* handle. Kept consistent with the existing method; happy to refactor both (snapshot fields, drop the guard, then await) in a separate change if preferred.
- **`InvalidArgumentSnafu { argument: "ResultSet handle not found" }` (`:393`)** — also identical to `result_set_get_stream`. The `ApiError::InvalidArgument` variant only has an `argument: String` field (no separate message field), and the crate uses it for the human-readable detail throughout; setting it to `"result_handle"` would drop that text. Left consistent with the existing convention.
- **Sync reader can panic if iterated in an async runtime (`:384`)** — this is inherent to sf_core's existing chunked `PrefetchChunkReader` (the very reader `result_set_get_stream` wraps into the FFI stream), not new behavior, and it's documented on the method. If an async-friendly path is desired, I'm happy to add a follow-up `result_set_collect_batches` that drains via `spawn_blocking`.

### Arrow version range (`Cargo.toml`)
The broad `>=54.0.0, <59` is intentional. The committed lockfile pins a single coherent version per build (currently `56.2.0`), so an individual build is never multi-version; the multi-major-in-public-API caveat is acknowledged as a deliberate trade-off. If preferred, I can centralize Arrow via `[workspace.dependencies]` (`workspace = true`) so every `arrow-*` crate is guaranteed to resolve to one version while keeping the breadth.

sf_core/src/apis/database_driver_v1/result_set.rs
```diff
@@ -371,6 +372,37 @@ impl DatabaseDriverV1 {
...
+        let rs = rs_ptr.lock().await;
+
+        build_reader_from_rowset_data(
+            &rs.data,
+            rs.reader_ctx.http_client.clone(),
+            &rs.reader_ctx.prefetch_config,
+            &self.wrapper_presets,
+        )
+        .await
+        .context(QueryResponseProcessingSnafu)
```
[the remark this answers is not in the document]
zeroshade: Fixed in b06792da. Added a `snapshot_reader_inputs` helper that clones the needed inputs (`RowsetData`, the HTTP client, and `PrefetchConfig`) under the lock and drops the guard, so the per-result-set lock is no longer held across `build_reader_from_rowset_data(...).await` (which can do a network fetch for multi-chunk results). Applied to both `result_set_get_reader` and the existing `result_set_get_stream` for consistency, and derived `Clone` on `RowsetData` to enable the snapshot.

The clone is cheap: bulk result data lives in lazily-fetched remote chunks, not in `RowsetData` (which holds only small inline data + chunk metadata). `build_reader_from_rowset_data` never re-acquires the result-set lock, so there was no deadlock — this removes the liveness/lock-held-across-await concern.

sf_core/src/apis/database_driver_v1/result_set.rs
```diff
@@ -371,6 +372,37 @@ impl DatabaseDriverV1 {
...
+    pub async fn result_set_get_reader(
+        &self,
+        result_handle: Handle,
+    ) -> Result<Box<dyn RecordBatchReader + Send>, ApiError> {
+        let rs_ptr = self.results.get_obj(result_handle).ok_or_else(|| {
+            InvalidArgumentSnafu {
+                argument: "ResultSet handle not found".to_string(),
+            }
+            .build()
+        })?;
```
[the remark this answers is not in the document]
zeroshade: Fixed in 01e088db. The `ApiError::InvalidArgument` variant carries a single `argument: String` field (display: `"Invalid argument: {argument}"`) with no separate message field, so rather than drop the human-readable detail I prefixed it with the parameter name: `"result_handle: ResultSet handle not found"`. Applied to both `result_set_get_reader` and the existing `result_set_get_stream` for consistency.

sf_core/src/apis/database_driver_v1/result_set.rs
```diff
@@ -371,6 +372,37 @@ impl DatabaseDriverV1 {
...
+    /// Builds and returns a fresh Arrow [`RecordBatchReader`] for this result
+    /// set -- the Rust-native counterpart of [`Self::result_set_get_stream`]
+    /// that skips the `FFI_ArrowArrayStream` wrapping. Built lazily from the
+    /// stored `RowsetData`, so it can be requested multiple times.
+    ///
+    /// For chunked result sets the reader downloads/parses chunks lazily via a
+    /// blocking channel receiver, so it must be drained from a synchronous
+    /// context -- never from within an async runtime (a `tokio` task or
+    /// `block_on`), which would panic.
+    pub async fn result_set_get_reader(
```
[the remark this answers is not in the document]
zeroshade: The synchronous-iteration constraint is inherent to sf_core's existing chunked `PrefetchChunkReader` — the same reader that `result_set_get_stream` (immediately above) already wraps into the FFI stream — so it isn't new behavior introduced by this method.

I've expanded the doc comment in 01e088db to make the sharp edge explicit and harder to misuse: building the reader (the `async fn` itself) is always safe to call/await, including inside a `tokio` task or `block_on`; only **iterating** the returned reader must happen off the async runtime, because the chunked path uses `blocking_recv`. The recommended patterns (drain after returning from `block_on` while keeping the runtime alive, on a dedicated `std::thread`, or via `tokio::task::spawn_blocking`) are now called out inline.

I kept the method as the low-level, non-FFI twin of `result_set_get_stream` deliberately. If an async-friendly surface is desired, I'm happy to add a follow-up `result_set_collect_batches` that drains via `spawn_blocking` and hands back the batches directly.

sf_core/Cargo.toml
```diff
@@ -39,8 +39,8 @@ tokio-stream = "0.1"
 toml = "0.8"
 uuid = { version = "1.18", features = ["v4"] }
 base64 = "0.22.1"
-arrow = { version = "56.0.0", features = ["ffi"] }
-arrow-ipc = "56.0.0"
+arrow = { version = ">=54.0.0, <59", features = ["ffi"] }
+arrow-ipc = ">=54.0.0, <59"
```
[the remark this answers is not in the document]
zeroshade: Intentional. The committed `Cargo.lock` pins a single coherent Arrow version per build (currently 56.x), so any individual build resolves exactly one Arrow major — it's never multi-version. The broad `>=54.0.0, <59` range only widens what downstream consumers can unify on; the multi-major-in-public-API caveat is an acknowledged, deliberate trade-off.

If you'd prefer a hard guarantee that every `arrow-*` crate across the workspace resolves to the same version, I can centralize them via `[workspace.dependencies]` (`workspace = true`) while keeping the breadth. Happy to do that as a follow-up if preferred.

sf_core/src/apis/database_driver_v1/result_set.rs
```diff
@@ -8,6 +8,7 @@ use crate::chunks::{ChunkDownloadData, ChunkFormatKind, PrefetchConfig};
 use crate::handle_manager::Handle;
 use crate::rest::snowflake::query_response::{Data, RowsetData, Stats};
 use crate::rest::snowflake::snowflake_get_query_result;
+use arrow::array::RecordBatchReader;
```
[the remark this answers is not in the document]
zeroshade: `arrow::array::RecordBatchReader` is a valid path: the `arrow` crate re-exports the `arrow-array` items under `arrow::array`, so this resolves to the same trait as `arrow_array::record_batch::RecordBatchReader`, and the return type `Box<dyn RecordBatchReader + Send>` uses that same trait.

Verified on the pinned Arrow 56.x: `cargo clippy --package sf_core --all-targets -- -D warnings` compiles cleanly with this import (exit 0, no warnings). No change needed here.

sf_core/Cargo.toml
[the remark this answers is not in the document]
zeroshade: Same as the related Arrow comment on this file: the broad range is intentional, and the committed `Cargo.lock` pins a single Arrow major per build (56.x today), so resolver outcome is deterministic for any given build and won't silently shift majors. The breadth exists to let downstream consumers unify Arrow on their chosen version.

If a stronger same-version guarantee across all `arrow-*` crates is preferred, I can add `[workspace.dependencies]` unification (`workspace = true`) as a follow-up while keeping the compatibility range.

sf_core/src/apis/database_driver_v1/result_set.rs
```diff
@@ -371,6 +386,32 @@ impl DatabaseDriverV1 {
...
 
+    /// Builds and returns a fresh Arrow [`RecordBatchReader`] for this result
+    /// set -- the Rust-native counterpart of [`Self::result_set_get_stream`]
+    /// that skips the `FFI_ArrowArrayStream` wrapping. Built lazily from the
+    /// stored `RowsetData`, so it can be requested multiple times.
+    ///
+    /// For chunked result sets the reader downloads/parses chunks lazily via a
+    /// blocking channel receiver, so it must be drained from a synchronous
+    /// context -- never from within an async runtime (a `tokio` task or
+    /// `block_on`), which would panic.
```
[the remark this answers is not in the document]
zeroshade: Fixed in 01e088db. Reworded the doc comment to remove the ambiguity: it now states explicitly that calling/awaiting the method is async-safe (it only builds the reader and never blocks), and that only **iterating** the returned reader must happen in a synchronous context. The recommended patterns are spelled out: drain after returning from `block_on` while keeping the runtime alive, on a dedicated `std::thread`, or via `tokio::task::spawn_blocking` — to avoid the `blocking_recv` panic.

sf_core/src/apis/database_driver_v1/result_set.rs
```diff
@@ -461,3 +507,70 @@ impl DatabaseDriverV1 {
...
+            .expect("fixture must deserialize into query_response::Data");
+        let descriptor = response_to_descriptor(&data, &WrapperPresets::default());
+        let reader_ctx = ReaderContext {
+            http_client: reqwest::Client::new(),
+            prefetch_config: PrefetchConfig::default(),
+        };
+        let handle = driver.create_result_set(descriptor, data.into_rowset_data(), reader_ctx);
+
+        let runtime = tokio::runtime::Runtime::new().expect("failed to build tokio runtime");
+        let reader: Box<dyn RecordBatchReader + Send + 'static> = runtime
```
sfc-gh-mhofman: The signature says Box<dyn RecordBatchReader + Send> but the test annotates + 'static. Can we remove to avoid confusion on the API?

sf_core/src/apis/database_driver_v1/result_set.rs
```diff
@@ -461,3 +507,70 @@ impl DatabaseDriverV1 {
...
+        "queryResultFormat": "json",
+        "rowset": [["1", "alice"], ["2", "bob"]],
+        "rowtype": [
+            {"name": "ID", "type": "FIXED", "nullable": false, "precision": 38, "scale": 0},
+            {"name": "NAME", "type": "TEXT", "nullable": true, "length": 100, "byteLength": 400}
+        ]
+    }"#;
+
+    #[test]
+    fn result_set_get_reader_returns_reader_without_ffi() {
```
sfc-gh-mhofman: the problem here is that it doesn't run on arrow and thus do not verify prefetch.rs#L181   which would panic when using arrow

sf_core/src/apis/database_driver_v1/result_set.rs
```diff
@@ -371,6 +386,37 @@ impl DatabaseDriverV1 {
...
+    /// Calling this method is async-safe: awaiting it (including inside a
+    /// `tokio` task or `block_on`) only builds the reader and never blocks.
+    ///
+    /// **Iterating** the returned reader is what must happen in a synchronous
+    /// context. For chunked result sets the reader downloads/parses chunks
+    /// lazily via a blocking channel receiver, so draining it from within an
+    /// async runtime would call `blocking_recv` and panic. Drain it after
+    /// returning from `block_on` (while keeping the runtime alive), on a
+    /// dedicated `std::thread`, or via `tokio::task::spawn_blocking`.
+    pub async fn result_set_get_reader(
```
sfc-gh-jszczerbinski: Creating new Rust API function seems redundant - can just pushout the conversion to protobuf layer?

sf_core/src/apis/database_driver_v1/result_set.rs
```diff
@@ -353,16 +368,16 @@ impl DatabaseDriverV1 {
     ) -> Result<Box<FFI_ArrowArrayStream>, ApiError> {
```
sfc-gh-jszczerbinski: Can we simply adjust this function instead?

sfc-gh-jszczerbinski (changes requested):

sf_core/src/apis/database_driver_v1/result_set.rs
zeroshade: I was trying to avoid changing the existing functions, but if we're okay with that approach, I'll update to do that

sf_core/src/apis/database_driver_v1/result_set.rs
zeroshade: Fixed

sf_core/src/apis/database_driver_v1/result_set.rs
zeroshade: Fixed, and added a test that does use arrow so both cases are tested and run

sf_core/src/apis/database_driver_v1/result_set.rs
zeroshade: Updated as requested remove the redundant function and move the FFI handling to the proto layer.

sfc-gh-jszczerbinski (approved):

sfc-gh-turbaszek:
Merged via 565d9a7ce7c8b009c84030b807ade0132f322a7a

PR #1327, opened 2026-06-09
feat(sf_core): add insecure_skip_tls_verify to disable TLS verification in one call

zeroshade:
## Summary

Adds a single connection parameter to skip TLS verification, instead of requiring
`verify_certificates` and `verify_hostname` to be set separately.

## What's new

New `Bool` connection param **`tls_skip_verify`** (default `false`). When enabled it
forces **both** `verify_hostname` and `verify_certificates` to `false` in one call, and
takes precedence over the individual flags.

## Implementation

- Registered in `config/param_registry.rs` (`Bool`, default `false`, `Connection` scope,
  `used_at_connect`).
- Honored in **both** TLS-config builders, so every connection path respects it:
  - `connection_config::build_tls_config` (ParamStore-backed)
  - `TlsConfig::from_settings` (used via `rest_parameters`)
- `from_settings` resolves keys through the `param_names::*` constants and the registry's
  alias table (rather than hardcoded string literals), so it honors the same wrapper keys
  as the ParamStore path.
- Boolean coercion (`Bool` / `"true"`/`"1"`/`"on"` / `"false"`/`"0"`/`"off"` / non-zero
  `Int`) is shared via `Setting::coerce_bool`, so both builders provably agree.
- Emits a `tracing::warn!` on connect when verification is disabled.
- Downstream rustls / CRL verification already keys off `TlsConfig.verify_hostname` /
  `verify_certificates`; with `verify_certificates == false`, `create_tls_client_with_config`
  early-returns an insecure client without installing the CRL verifier, so CRL revocation
  checks are bypassed as well.

## Behavior

| `tls_skip_verify` | resulting `verify_hostname` / `verify_certificates` |
| --- | --- |
| `true` | both forced `false` (overrides the individual flags; CRL also bypassed) |
| `false` / unset | existing behavior — individual flags, default `true` |

Backward compatible: unset/`false` leaves current behavior unchanged.

## Tests

- `tls_skip_verify_disables_both_checks`
- `tls_skip_verify_overrides_individual_verify_flags`
- `tls_skip_verify_canonicalizes_from_any_case_and_disables_verification` (end-to-end via
  `ConnectionConfig::build`)
- `tls_skip_verify_bypasses_crl_even_when_crl_check_mode_enabled` (locks the registry's
  CRL claim against drift)
- `from_settings_honors_registered_aliases` (raw alias resolution in `from_settings`)
- `from_settings_skip_disables_both_for_bool_and_string`
- existing `build_tls_booleans_*` continue to pass

## Note

Named `tls_skip_verify` (per review). Insecure; intended for testing/development only.
The auto-generated `python/.../connection_config.py` binding is regenerated to include the
new field.

### DIFF
--- a/sf_core/src/config/connection_config.rs
+++ b/sf_core/src/config/connection_config.rs
@@ -329,7 +329,12 @@ fn build_tls_config(settings: &ParamStore) -> TlsConfig {
     let custom_root_store_path = settings
         .get_string(CUSTOM_ROOT_STORE_PATH)
         .map(PathBuf::from);
-    let skip_tls_verify = settings.get_bool(INSECURE_SKIP_TLS_VERIFY).unwrap_or(false);
+    let skip_tls_verify = settings.get_bool(TLS_SKIP_VERIFY).unwrap_or(false);
+    if skip_tls_verify {
+        tracing::warn!(
+            "TLS verification disabled via tls_skip_verify: certificate, hostname, and CRL revocation checks are all bypassed. Do not use in production."
+        );
+    }
     let verify_hostname = !skip_tls_verify && settings.get_bool(VERIFY_HOSTNAME).unwrap_or(true);
     let verify_certificates =
         !skip_tls_verify && settings.get_bool(VERIFY_CERTIFICATES).unwrap_or(true);
@@ -1174,9 +1179,9 @@ mod tests {
     }
 
     #[test]
-    fn insecure_skip_tls_verify_disables_both_checks() {
+    fn tls_skip_verify_disables_both_checks() {
         let mut settings = minimal_password_settings();
-        settings.insert("insecure_skip_tls_verify".into(), Setting::Bool(true));
+        settings.insert("tls_skip_verify".into(), Setting::Bool(true));
 
         let config = ConnectionConfig::build(&settings).unwrap();
         assert!(!config.tls.verify_hostname);
@@ -1184,15 +1189,60 @@ mod tests {
     }
 
     #[test]
-    fn insecure_skip_tls_verify_overrides_individual_verify_flags() {
+    fn tls_skip_verify_overrides_individual_verify_flags() {
         let mut settings = minimal_password_settings();
-        settings.insert("insecure_skip_tls_verify".into(), Setting::Bool(true));
+        settings.insert("tls_skip_verify".into(), Setting::Bool(true));
         settings.insert("verify_hostname".into(), Setting::Bool(true));
         settings.insert("verify_certificates".into(), Setting::Bool(true));
 
         let config = ConnectionConfig::build(&settings).unwrap();
         assert!(!config.tls.verify_hostname);
         assert!(!config.tls.verify_certificates);
+    }
+
+    #[test]
+    fn tls_skip_verify_canonicalizes_from_any_case_and_disables_verification() {
+        use crate::config::param_registry::registry;
+
+        for raw_key in ["tls_skip_verify", "TLS_SKIP_VERIFY", "Tls_Skip_Verify"] {
+            let canonical = registry()
+                .resolve(raw_key)
+                .unwrap_or_else(|| panic!("{raw_key} should resolve"))
+                .canonical_name;
+            assert_eq!(canonical, "tls_skip_verify");
+
+            let mut settings = minimal_password_settings();
+            settings.insert(canonical.to_string(), Setting::Bool(true));
+
+            let config = ConnectionConfig::build(&settings).unwrap();
+            assert!(
+                !config.tls.verify_hostname,
+                "{raw_key} should disable hostname check"
+            );
+            assert!(
+                !config.tls.verify_certificates,
+                "{raw_key} should disable cert check"
+            );
+        }
+    }
+
+    #[test]
+    fn tls_skip_verify_bypasses_crl_even_when_crl_check_mode_enabled() {
+        // Locks the registry description's CRL claim against drift: tls_skip_verify forces
+        // verify_certificates=false, and create_tls_client_with_config (tls/client.rs) then
+        // early-returns an insecure client without installing the CRL verifier — so CRL is
+        // bypassed even at crl_check_mode=ENABLED. The mode itself is left intact, just unused.
+        let mut settings = minimal_password_settings();
+        settings.insert("tls_skip_verify".into(), Setting::Bool(true));
+        settings.insert("crl_check_mode".into(), Setting::String("ENABLED".into()));
+
+        let config = ConnectionConfig::build(&settings).unwrap();
+        assert!(!config.tls.verify_certificates);
+        assert!(!config.tls.verify_hostname);
+        assert!(matches!(
+            config.tls.crl_config.check_mode,
+            CertRevocationCheckMode::Enabled
+        ));
     }
 
     #[test]
--- a/sf_core/src/config/connection_config.rs
+++ b/sf_core/src/config/connection_config.rs
@@ -329,8 +329,10 @@ fn build_tls_config(settings: &ParamStore) -> TlsConfig {
     let custom_root_store_path = settings
         .get_string(CUSTOM_ROOT_STORE_PATH)
         .map(PathBuf::from);
-    let verify_hostname = settings.get_bool(VERIFY_HOSTNAME).unwrap_or(true);
-    let verify_certificates = settings.get_bool(VERIFY_CERTIFICATES).unwrap_or(true);
+    let skip_tls_verify = settings.get_bool(INSECURE_SKIP_TLS_VERIFY).unwrap_or(false);
+    let verify_hostname = !skip_tls_verify && settings.get_bool(VERIFY_HOSTNAME).unwrap_or(true);
+    let verify_certificates =
+        !skip_tls_verify && settings.get_bool(VERIFY_CERTIFICATES).unwrap_or(true);
 
     TlsConfig {
         crl_config,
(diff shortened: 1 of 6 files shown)


sfc-gh-turbaszek (approved):

sf_core/src/config/connection_config.rs
```diff
@@ -329,8 +329,10 @@ fn build_tls_config(settings: &ParamStore) -> TlsConfig {
     let custom_root_store_path = settings
         .get_string(CUSTOM_ROOT_STORE_PATH)
         .map(PathBuf::from);
-    let verify_hostname = settings.get_bool(VERIFY_HOSTNAME).unwrap_or(true);
-    let verify_certificates = settings.get_bool(VERIFY_CERTIFICATES).unwrap_or(true);
+    let skip_tls_verify = settings.get_bool(INSECURE_SKIP_TLS_VERIFY).unwrap_or(false);
```
sfc-gh-mhofman: can we name it less scary? @sfc-gh-boler

sf_core/src/tls/config.rs
```diff
@@ -26,14 +26,22 @@ impl TlsConfig {
...
+            settings
+                .get_bool(key)
+                .or_else(|| {
+                    settings
+                        .get_string(key)
+                        .map(|s| s.eq_ignore_ascii_case("true"))
+                })
+                .unwrap_or(default)
+        };
+        let skip_tls_verify = bool_setting("insecure_skip_tls_verify", false);
```
sfc-gh-mhofman: Correctness: the alias `TLS_SKIP_VERIFY` registered in `param_registry.rs` is honored only by `build_tls_config`, which goes through `ParamStore` and resolves aliases. This `from_settings` path looks the canonical name up directly in the underlying settings map with no alias resolution, so a user who passes `TLS_SKIP_VERIFY` through whatever feeds this builder will silently keep TLS verification on. Either canonicalize the keys before this lookup, or also check `bool_setting("tls_skip_verify", …)`. The registry alias-resolution test only exercises canonicalization and does not catch this.

sf_core/src/tls/config.rs
```diff
@@ -26,14 +26,22 @@ impl TlsConfig {
...
+        let verify_hostname = !skip_tls_verify && bool_setting("verify_hostname", true);
```
sfc-gh-mhofman: Maintainability: the canonical names are hardcoded as string literals here (`"insecure_skip_tls_verify"` on the line above, plus `"verify_hostname"` here and `"verify_certificates"` below), while `build_tls_config` uses the `param_names::*` constants. If a canonical name is renamed, this path goes silently dead with no compile-time tie. Prefer `param_names::INSECURE_SKIP_TLS_VERIFY.as_str()` etc. The two `verify_*` reads already had this issue — would be good to fix in the same pass.

sf_core/src/config/connection_config.rs
sfc-gh-mhofman: Worth a `tracing::warn!` when `skip_tls_verify == true` (e.g. `"TLS verification disabled via insecure_skip_tls_verify — do not use in production"`). Other Snowflake drivers and most TLS libraries emit a one-line warn on connect when verification is off; this is exactly the kind of switch that ends up silently shipped to prod otherwise.

sf_core/src/config/connection_config.rs
```diff
@@ -1171,6 +1173,28 @@ mod tests {
...
 
+    #[test]
+    fn insecure_skip_tls_verify_disables_both_checks() {
+        let mut settings = minimal_password_settings();
+        settings.insert("insecure_skip_tls_verify".into(), Setting::Bool(true));
+
+        let config = ConnectionConfig::build(&settings).unwrap();
+        assert!(!config.tls.verify_hostname);
+        assert!(!config.tls.verify_certificates);
+    }
```
sfc-gh-mhofman, line 1189: Both new tests insert the canonical key directly into the settings map. Worth one more test that goes through the alias `TLS_SKIP_VERIFY` and `ConnectionConfig::build` (registry-backed canonicalization) to lock in that the alias actually disables verification end-to-end. The `param_registry.rs` alias test only proves the registry maps the alias — it does not prove this code path honors it.

sf_core/src/config/connection_config.rs
```diff
@@ -1171,6 +1173,28 @@ mod tests {
...
+    fn insecure_skip_tls_verify_overrides_individual_verify_flags() {
+        let mut settings = minimal_password_settings();
+        settings.insert("insecure_skip_tls_verify".into(), Setting::Bool(true));
+        settings.insert("verify_hostname".into(), Setting::Bool(true));
+        settings.insert("verify_certificates".into(), Setting::Bool(true));
+
+        let config = ConnectionConfig::build(&settings).unwrap();
+        assert!(!config.tls.verify_hostname);
+        assert!(!config.tls.verify_certificates);
+    }
```
sfc-gh-mhofman, line 1201: The registry description for this flag claims "CRL revocation checks are bypassed too" when it is on. That behavioral claim isn't asserted anywhere. A test that sets `insecure_skip_tls_verify=true` together with `crl_check_mode=enabled` and asserts the resulting effective CRL/verification state would lock it down — otherwise the doc and the code can drift apart silently.

sfc-gh-mhofman (commented):
Not blocking, but flagging an architectural concern: this PR wires the new flag in two places — `connection_config::build_tls_config` (ParamStore-backed, alias-aware) and `TlsConfig::from_settings` (raw settings map, used via `rest_parameters`). The PR's own inline comment in `from_settings` spells out that the two paths must agree on coercion. Every future TLS flag is now a known two-place change with no compile-time tie between them, and as the inline comments below show, the divergence is already producing real issues (alias silently dropped, hardcoded canonical names). Worth a follow-up to unify on a single TLS-config builder so that wiring a flag once is enough.

sfc-gh-mhofman:
PTAL at the comments @zeroshade

sf_core/src/config/connection_config.rs
zeroshade: removed `INSECURE` from the name, so it's just `TLS_SKIP_VERIFY`

zeroshade: Renamed the canonical param to `tls_skip_verify` and dropped both the `insecure_skip_tls_verify` name and the redundant `TLS_SKIP_VERIFY` alias — `tls_skip_verify` is now the single name everywhere (registry, both TLS builders, tests, and the regenerated Python binding). Pushed in 579d33c7.

sf_core/src/tls/config.rs
zeroshade: Fixed. `from_settings` now resolves through a registry-backed helper (`lookup_setting`): it checks the canonical `ParamKey` first, then falls back to the registry's registered aliases for that param — so the raw-bag path honors the same wrapper keys as the ParamStore-backed `build_tls_config`. Added `from_settings_honors_registered_aliases` (passes a raw `TLS_VERIFY_HOSTNAME` and asserts it takes effect) to lock this in. (`tls_skip_verify` itself no longer has an alias after the rename, but the mechanism now covers `verify_hostname`/`verify_certificates`, which do.)

sf_core/src/tls/config.rs
zeroshade: Done. `from_settings` now uses the `param_names::*` constants (`TLS_SKIP_VERIFY`, `VERIFY_HOSTNAME`, `VERIFY_CERTIFICATES`, `CUSTOM_ROOT_STORE_PATH`) instead of string literals, so a canonical-name rename is a compile-time break here too.

sf_core/src/config/connection_config.rs
zeroshade: Added. `build_tls_config` now emits `tracing::warn!("TLS verification disabled via tls_skip_verify: certificate, hostname, and CRL revocation checks are all bypassed. Do not use in production.")` when the flag is on.

sf_core/src/config/connection_config.rs
zeroshade: Added `tls_skip_verify_canonicalizes_from_any_case_and_disables_verification`: for `tls_skip_verify` / `TLS_SKIP_VERIFY` / `Tls_Skip_Verify` it resolves the key through the registry, asserts the canonical is `tls_skip_verify`, then runs the value through `ConnectionConfig::build` and asserts both checks end up disabled — i.e. the registry-backed canonicalization path actually honors it end-to-end.

sf_core/src/config/connection_config.rs
zeroshade: Added `tls_skip_verify_bypasses_crl_even_when_crl_check_mode_enabled`: sets `tls_skip_verify=true` together with `crl_check_mode=ENABLED` and asserts `verify_certificates==false` / `verify_hostname==false` while `crl_config.check_mode` stays `Enabled`. A comment ties this to `tls/client.rs`, where `create_tls_client_with_config` early-returns an insecure client (no CRL verifier installed) once `verify_certificates==false` — so the description's "CRL revocation checks are bypassed too" claim is now locked against drift.

zeroshade:
@sfc-gh-mhofman thanks for the review — addressed all six inline comments in 579d33c7. Summary:

1. **Naming** — renamed canonical to `tls_skip_verify`; dropped the `insecure_skip_tls_verify` name *and* the redundant `TLS_SKIP_VERIFY` alias. Single name everywhere now.
2. **Alias dropped in `from_settings`** — `from_settings` now resolves via the `param_names::*` constants + the registry alias table (`lookup_setting`), matching the ParamStore path; new test passes a raw `TLS_VERIFY_HOSTNAME` to prove it.
3. **Hardcoded literals** — replaced with `param_names::*` constants.
4. **Warn** — `tracing::warn!` on connect when verification is disabled.
5. **Alias e2e test** — canonicalization-through-`ConnectionConfig::build` test added.
6. **CRL claim** — test pins `tls_skip_verify=true` + `crl_check_mode=ENABLED` ⇒ `verify_certificates=false`, with a comment tying it to the `client.rs` early-return that skips the CRL verifier.

Also regenerated the auto-generated `python/.../connection_config.py` binding (the base PR had left it stale) and fixed a double-period in the generated docstring.

On the **architectural note** (unifying `build_tls_config` and `TlsConfig::from_settings`): agreed it'd be cleaner to have one source of truth. I kept this PR to the minimal hardening you flagged as non-blocking — the two paths now share `Setting::coerce_bool` (so coercion can't drift) and both resolve through `param_names`/registry, which removes the two concrete bugs (silently-dropped alias, hardcoded names). Full collapse into a single builder (which would also fold in the parallel `build_crl_config` / `CrlConfig::from_settings` duplication) I'd like to do as a dedicated follow-up rather than expand the scope here — happy to file it.

sfc-gh-jszczerbinski:
Merged in commit: https://github.com/snowflakedb/universal-driver/commit/6065c6c8db78. Thanks for contribution

PR #1329, opened 2026-06-11
fix(sf_core): DDL rows_affected returns None, not a misleading 1; share query-type taxonomy

zeroshade:
## Summary

DDL statements previously surfaced a misleading **`rows_affected = 1`**. The native
`calculate_rows_affected` blindly passed through Snowflake's `data.total` for any
non-DML statement, and Snowflake populates `total: 1` as a generic *success* marker
for DDL — so `CREATE`/`DROP`/`ALTER`/`ALTER SESSION` all reported "1 row affected".

The backend already tells us the statement kind via `statementTypeId`, and the ODBC
layer already combined that with the row count correctly (DDL → `SQLRowCount == -1`).
This PR lifts that classification into `sf_core` as a **single source of truth** so the
native result path and the ODBC `SQLRowCount` path can no longer drift.

## What changed

**New shared module** — `sf_core::query_types::statement_type`
- Moves the `statementTypeId` taxonomy (`QueryType` + `ResultKind`: `Cursor` /
  `UpdateCount` / `NoResult`) out of the ODBC crate into `sf_core`, semantics unchanged
  (reference-driver `SFResults.cpp` precedence, the `MANAGE_PATS` / `COPY` exceptions,
  bit-mask family matching). The taxonomy tests moved with it.
- `sf_core/src/query_types.rs` became `query_types/` (`row_type.rs` unchanged, history
  preserved via `git mv`; new `statement_type.rs`; `mod.rs` re-exports both). All existing
  `crate::query_types::RowType` paths are unaffected.

**The fix** — `calculate_rows_affected` keys off the shared classifier:
| ResultKind | rows_affected |
| --- | --- |
| `UpdateCount` (DML) | sum of affected-row columns from the rowset (unchanged) |
| `Cursor` (SELECT / SHOW / PUT / GET / ...) | `data.total` |
| `NoResult` (DDL / TCL / unknown) | **`None`** (was the misleading `data.total`) |

A single `effective_statement_type_id()` derives the type — server `statementTypeId`,
falling back to `command = UPLOAD/DOWNLOAD` for file transfers that omit it — and feeds
**both** descriptor classification and rows-affected, so they cannot diverge.

**ODBC repointed** — `odbc/src/api/query_type.rs` is now a thin
`pub use sf_core::query_types::statement_type::{QueryType, ResultKind};`. Every consumer
(`statement.rs`, `utils.rs`, `data.rs`) keeps its import path; both layers are now the
same code.

## Tests

- Shared classifier: full `ResultKind` matrix (DML/SELECT/SHOW/DDL/TCL/MULTI_STMT/unknown).
- `calculate_rows_affected`: DML sum, null cells, `SELECT` uses total, **DDL → None**,
  **unknown → None**, and **file-transfer fallback uses total** (regression for the
  PUT/GET omitted-id case).
- Updated the `ALTER SESSION SET` integration assertions from `Some(1)` → `None`.
- `cargo fmt` clean; `sf_core` + `odbc` compile clean; `cargo clippy` clean; all `sf_core`
  module tests and the full **1382 `odbc` lib tests** pass.

## Compatibility

Native consumers reading `rows_affected` for DDL now get `None` instead of `Some(1)`.
This matches what the ODBC layer already reports (`SQLRowCount == -1` for DDL) and removes
a misleading success count. DML and SELECT behavior is unchanged.


sfc-gh-jszczerbinski (approved):
Looks good to me, please rebase

zeroshade:
@sfc-gh-jszczerbinski i've rebased this! Thanks!

sfc-gh-jszczerbinski:
Merged in: https://github.com/snowflakedb/universal-driver/commit/c557468154f3b80adb1d779e5a3e226c6da1ac72, thanks for contributing

PR #1330, opened 2026-06-11
feat(core): support in-band per-statement QUERY_TAG

zeroshade:
## Why

Setting a per-statement `QUERY_TAG` currently requires a separate
`ALTER SESSION SET QUERY_TAG = '...'` statement — an extra statement
handle and a network round-trip. Snowflake's `/queries/v1/query-request`
endpoint already accepts per-request session parameters in the
`parameters` object (the same path used today for `MULTI_STATEMENT_COUNT`
and `STATEMENT_TIMEOUT_IN_SECONDS`), so the tag can be sent in-band with
the statement it applies to.

This is consistent with the SQL API reference (`query_tag` is in the
per-request parameter allowlist) and with how gosnowflake
(`WithQueryTag`), snowflake-connector-python (`_statement_params`), and
snowflake-jdbc (`setParameter("QUERY_TAG", ...)`) all behave.

## What changed

- Added a statement-scoped `statement_query_tag` parameter to the param
  registry, mapped to the server-side `QUERY_TAG` key in the
  query-request `parameters` object via `QUERY_PARAMETER_NAMES`.
- Kept it deliberately distinct from the existing connection-level
  `query_tag` passthrough (an unregistered key forwarded as a session
  parameter at login). Because registry resolution is case-insensitive,
  registering `query_tag` itself would divert session-wide login
  forwarding and cause connection-string `QUERY_TAG` to be rejected as
  statement-scoped. A separate canonical key avoids that while still
  emitting the wire-level `QUERY_TAG`.
- Consumers set it per statement via `statement_set_option(s)`; it then
  serializes into the query body for both the sync and async execute
  paths through the existing `build_query_parameters` flow. No wrapper or
  protobuf changes — core layer only.

## Usage (core Rust consumer)

```rust
ds.statement_set_option(
    stmt,
    param_names::STATEMENT_QUERY_TAG.as_str().to_string(),
    Setting::String("etl_job_42".to_string()),
).await?;
// execute -> request body includes "parameters": { "QUERY_TAG": "etl_job_42" }
```

## Tests

- `build_query_parameters_maps_statement_query_tag` — maps the canonical
  key to `QUERY_TAG`.
- `statement_query_tag_serializes_into_request_parameters` — asserts the
  serialized request body contains `parameters.QUERY_TAG`.
- `statement_accepts_statement_query_tag_option` — accepted as a
  statement-scoped option.
- `collect_unknown_settings_returns_only_unregistered_keys` — existing
  connection-level `query_tag` login forwarding is unchanged (regression
  guard).

## Notes / scope

- A tag set via `statement_set_option` persists on the statement handle
  across re-executes until changed; strictly per-call scoping would be a
  follow-up.
- No language-wrapper surface added yet; this exposes the capability at
  the core layer only.

### DIFF
--- a/sf_core/src/apis/database_driver_v1/statement.rs
+++ b/sf_core/src/apis/database_driver_v1/statement.rs
@@ -714,8 +714,10 @@ fn setting_to_json_value(setting: &Setting) -> serde_json::Value {
 
 /// Server-side parameter names forwarded in the query request.
 /// Each entry maps a local `ParamKey` to the Snowflake server-side parameter name.
-const QUERY_PARAMETER_NAMES: &[(ParamKey, &str)] =
-    &[(param_names::MULTI_STATEMENT_COUNT, "MULTI_STATEMENT_COUNT")];
+const QUERY_PARAMETER_NAMES: &[(ParamKey, &str)] = &[
+    (param_names::MULTI_STATEMENT_COUNT, "MULTI_STATEMENT_COUNT"),
+    (param_names::STATEMENT_QUERY_TAG, "QUERY_TAG"),
+];
 
 fn build_query_parameters(settings: &ParamStore) -> Option<HashMap<String, serde_json::Value>> {
     let mut params = HashMap::new();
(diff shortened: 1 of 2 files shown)


sfc-gh-jszczerbinski (commented):
Already implemented

PR #1334, opened 2026-07-16
SNOW-3784403 Align CRL implementation with the old driver

sfc-gh-pfus:
### TL;DR

Hardens CRL caching with a configurable download size cap, a corrected validity-time unit (days → seconds), a new on-disk cache removal delay, and a periodic background cleanup of stale in-memory and on-disk CRL entries.

### What changed?

**CRL download size cap**
- Added `crl_max_download_size` (default: 20 MB) to `CrlConfig` and the param registry. CRL downloads now stream the response body and abort with a `DownloadSizeExceeded` error if either the `Content-Length` header or the accumulated body exceeds the limit.
- Introduced `execute_bytes_with_retry_capped` in `http/retry.rs` to enforce the cap during streaming. `ResponseTooLarge` is surfaced immediately without retries; transport/status failures still go through the normal retry path.

**CRL validity time unit correction**
- The Rust core's `crl_validity_time` param is now in **seconds** (default: 86400 s / 24 h), aligning with gosnowflake. The Python connector's public API continues to accept the value in **days** and converts to seconds in `to_options`. The Python field is hand-written in `connection_config_mixin.py` and excluded from code generation via `PYTHON_HANDWRITTEN_FIELDS` in `param_defs_export.rs` to prevent the seconds-based generated field from shadowing it.

**On-disk cache removal delay**
- Added `crl_on_disk_cache_removal_delay` (default: 25200 s / 7 h) to `CrlConfig` and the param registry. Expired on-disk CRL files are retained for this duration past their `nextUpdate` before the background cleaner removes them, aiding debuggability.

**Periodic background cache cleanup**
- The CRL background task now runs a cleanup pass every hour. In-memory entries past their `nextUpdate` or older than `validity_time` are evicted. On-disk files whose `nextUpdate` plus the removal delay has passed are deleted. Unparseable files are never touched.

**Cache freshness check**
- Both in-memory and on-disk cache reads now enforce a dual condition: the CRL's `nextUpdate` must be in the future **and** the entry's age must not exceed `validity_time`. On-disk entries use the file's mtime as the download time, matching gosnowflake's `stat.ModTime()` approach.

**Default timeout reduction**
- `crl_http_timeout` default reduced from 30 s to 10 s.

### How to test?

- Unit tests in `crl/config.rs` verify that defaults match gosnowflake values and that settings are read in the correct units (seconds for time, MB converted to bytes).
- `cleanup_on_disk_cache_is_safe` test in `crl/cache.rs` confirms that unparseable files and foreign files are never removed, that a missing directory does not panic, and that cleanup is a no-op when disk caching is disabled.
- Integration tests in `crl/integration_test.rs` cover the updated default values.
- End-to-end: configure `crl_max_download_size` to a very small value and point the connector at a real CRL endpoint to confirm the download is aborted with a `DownloadSizeExceeded` error.

### Why make this change?

- **Memory safety**: an unbounded CRL download could exhaust process memory; the size cap prevents this.
- **Unit alignment**: the previous Python API accepted `crl_validity_time` in days while the Rust core was also treating it as days internally. Aligning the core to seconds (matching gosnowflake) and converting at the Python boundary preserves backward compatibility while making the internal representation unambiguous.
- **Cache hygiene**: without periodic cleanup, stale CRL entries could accumulate indefinitely in memory and on disk. The hourly background pass and the configurable removal delay give operators control over retention while keeping the cache fresh.

### DIFF
--- a/python/src/snowflake/connector/_internal/connection_config_mixin.py
+++ b/python/src/snowflake/connector/_internal/connection_config_mixin.py
@@ -87,6 +87,19 @@ class ConnectionConfigMixin:
     with snowflake-connector-python; never forwarded to the Rust core.  When
     ``None`` the wrapper treats it as ``True`` (legacy default)."""
 
+    crl_validity_time: int | None = None
+    """Maximum age of a cached CRL before it is re-fetched, in **days**.
+
+    When unset (``None``) the Rust core's own default applies (24h); the value
+    is not forwarded. When set, it is interpreted as days and converted to
+    seconds in :meth:`to_options` before being forwarded, because the core
+    param ``crl_validity_time`` is seconds-based.
+
+    Kept in days for backward compatibility with the Python connector's public
+    API. Hand-written here — and excluded from the generated dataclass via
+    ``PYTHON_HANDWRITTEN_FIELDS`` in ``param_defs_export.rs`` — so the
+    seconds-based core param does not shadow this days-based field."""
+
     _extra: dict[str, Any] = field(default_factory=dict, repr=False)
     """Unknown kwargs forwarded to the Rust core for validation.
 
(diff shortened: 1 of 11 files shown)


sfc-gh-pfus:
* **#1334** <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1334?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/universal-driver/1334?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1337, opened 2026-07-20
SNOW-3675196: thread async cancellation through the driver stack

zeroshade:
## Summary

Threads async cancellation (`tokio_util::sync::CancellationToken`) end-to-end through the driver stack so a caller-initiated cancel — including ODBC `SQLCancel` — reliably interrupts in-flight work and surfaces as `CANCELLED` / `HY008` instead of a generic error.

## What changed

**Core threading**
- Generated `Transport` / `DatabaseDriverClient` methods take a `CancellationToken`, threaded through the `DatabaseDriverV1` core ops, REST/chunk retry helpers, and the multi-chunk fetch path (`fetch_chunks_reader` → prefetch/initial-chunk readers → `get_chunk_data`) down to the leaf awaits.
- `execute_with_retry` and the async-query poll loop race the token via `tokio::select!`, returning `HttpError::Cancelled` / `SfError::Cancelled`.
- `ApiError::is_cancelled` propagates cancellation to `StatusCode::CANCELLED`; ODBC maps core `CANCELLED` to `OperationCanceled` (`HY008`).

**Coverage**
- SHOW / reader-drain paths surface cancellation (including successful drains).
- HTTP retry `on_response` + body reads, REST response-body reads, and Azure/GCS/S3 file-transfer body reads race the token.
- `logout` preserves cancellation (bypasses BestEffort suppression).
- Native Okta and external-browser login (including the callback wait) are cancellable; result-set handles are released on stream-fetch error.
- ODBC: the `SQLCancel` armed token is threaded through catalog RPCs, the SHOW path, `SQLMoreResults`, execute finalization, and the numeric-settings refresh; the async-spawn path publishes the token before spawning to close a TOCTOU race.

## Known follow-ups (tracked separately)

- OAuth Authorization Code / OAuth Client Credentials / Workload Identity login flows are not yet threaded.
- GCS credential-refresh cancellation (shared `StageInfoRefresher` / `StageInfoRefreshError`) is not yet cancellation-aware.

## Testing

- Added cancellation unit tests (HTTP retry, reader drain, ODBC).
- `cargo check --package sf_core --package odbc --all-targets` is clean.
- Cancellation-focused test suites pass (sf_core + odbc).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2692,6 +2692,7 @@ dependencies = [
  "proto_utils",
  "sf_core",
  "tokio",
+ "tokio-util",
  "tracing",
  "tracing-subscriber",
 ]
@@ -4117,6 +4118,9 @@ dependencies = [
 [[package]]
 name = "proto_utils"
 version = "0.1.0"
+dependencies = [
+ "tokio-util",
+]
 
 [[package]]
 name = "ptr_meta"
--- a/jdbc_bridge/Cargo.toml
+++ b/jdbc_bridge/Cargo.toml
@@ -17,6 +17,7 @@ proto_utils = { path = "../proto_utils" }
 tracing-subscriber = { version = "0.3.19", features = ["fmt", "env-filter"] }
 tracing = "0.1.41"
 tokio = { version = "1.50.0", features = ["rt"] }
+tokio-util = "0.7"
 
 [dependencies.jni]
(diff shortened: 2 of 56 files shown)


sfc-gh-jszczerbinski (commented):
This being worked on internally with a slightly different design.

PR #1347, opened 2026-09-18
NO-SNOW: FIPS F2: drop OpenSSL from shipped artifacts

zeroshade:
> **Stacked on #1346** (which is itself stacked on #1345). The diff against
> `main` therefore includes those PRs' 14 commits — **review only the last
> commit here**, `NO-SNOW: FIPS F2: drop OpenSSL from shipped artifacts`.
> GitHub cannot use a cross-repo base branch, so this targets `main` until the
> PRs below it merge.

## What

Closes **F2**, the last Blocker-severity finding in the FIPS plan: shipped
artifacts could statically link an unvalidated OpenSSL (`openssl-src
300.6.0+3.6.2`).

`openssl` moves to `[dev-dependencies]` and the `vendored-openssl` feature is
deleted from `sf_core`, `odbc`, `jdbc_bridge` and `python_bridge`. Because
`cargo build` does not compile dev-dependencies, **no shipped artifact links
OpenSSL at all** any more. This is only unblocked now because #1346 finished
porting the four production call sites — before that, deleting the feature
would have made artifacts depend on a *host* OpenSSL, which is strictly worse.

Evidence rather than assertion:

- `openssl-src` is gone from `Cargo.lock` **and** `python/Cargo.lock.sdist`
- `cargo tree -e normal` finds no OpenSSL in `sf_core`, `odbc`, `jdbc_bridge`,
  `python_bridge` or `nodejs_bridge`
- `ldd target/debug/libsfodbc.so` shows no `libssl`/`libcrypto`

## Three things worth a reviewer's attention

**1. The blast radius was larger than the manifests.** `vendored-openssl` had
six consumers outside `.github/workflows/`, each of which would have failed CI
on an unknown-feature error: `ci/build_wif_artifacts.sh`, the three
`ci/test_matrix/mappings/*.py`, that generator's test expectations, and
`python/hatch_native_arrow.py`.

**2. The WIF e2e binary changes linkage.** It used to vendor OpenSSL
specifically so the public runtime container could stay minimal. It is a *test*
binary, so it still needs OpenSSL and now links `libssl.so.3` dynamically
(confirmed with `ldd`, not assumed). `ci/wif/run_in_container.sh` therefore
installs `openssl-libs`; build and runtime images share the `rockylinux:8` base,
so the ABI lines up.

**3. `nodejs_bridge` never had a `vendored-openssl` passthrough.** Its artifacts
have been dynamically linking the *host* libssl while the other three bindings
linked a vendored copy. This change erases that asymmetry as a side effect
rather than papering over it.

## Why OpenSSL stays unvendored for tests

`cargo test` *does* compile dev-dependencies — the X.509/CRL fixtures in
`crl/certificate_parser.rs`, `tls/x509_utils.rs` and `tls/test_helpers.rs` use
OpenSSL, and porting them would put the same library on both sides of those
tests and stop them proving anything.

A cargo feature cannot reference a dev-dependency, so `vendored` would have to
be unconditional — which would make every Windows test build compile OpenSSL
through Perl and nmake, exactly what the vcpkg lanes exist to avoid. The test
lanes keep supplying OpenSSL instead (vcpkg on Windows, distro packages
elsewhere).

## Deliberately out of scope

- **Perl provisioning is left in place.** It is now dead on these lanes —
  `aws-lc-sys` sets `DISABLE_PERL=ON`/`DISABLE_GO=ON` when using pregenerated
  sources (`builder/cmake_builder.rs`), so the Strawberry-Perl PATH steps and
  the `perl-IPC-Cmd`/`perl-Time-Piece` installs no longer do anything. Pruning
  them wants a CI run to confirm, and Phase 6's FIPS lanes will need Perl and Go
  back regardless, since `aws-lc-fips-sys` does not take that branch.
- **No rename of `fips-tls` to `fips`.** Coverage is still incomplete: F7 (SigV4
  HMAC via RustCrypto, plus ring linked through `aws-sdk-s3`'s default
  `sigv4a`) and F8 (`oauth2` draws CSRF/PKCE randomness from `rand`) are open.
  The feature comment in `sf_core/Cargo.toml` is updated to name those two as
  the remaining gaps instead of OpenSSL.

## Verification

- 2199 tests pass by default, 2200 with `--features fips-tls`, on a real
  `aws-lc-fips-sys` build
- 114 matrix-generator tests pass; the `legacy` dict in
  `test_generate_matrix.py` is updated as that test's own comment instructs
- `e2e_tests` target compiles; `cargo check` clean for all four bindings
- `python/Cargo.lock.sdist` refreshed with `scripts/python-sdist-cargo-lock.sh
  refresh` and re-verified under `--locked`
- fmt clean; clippy unchanged from baseline


sf_core/Cargo.toml
```diff
@@ -147,16 +222,52 @@ brotli = "8.0.2"
...
+# both sides of those tests and stop them proving anything.
+#
+# It is a dev-dependency rather than a normal one because `cargo build` does not
+# compile dev-dependencies: shipped artifacts now link no OpenSSL at all, which
+# is what retired the `vendored-openssl` feature (F2). Left unvendored on
+# purpose -- a cargo feature cannot reference a dev-dependency, so `vendored`
+# would have to be unconditional, and that would make every Windows test build
+# compile OpenSSL from source through Perl and nmake. The test lanes supply it
+# instead (vcpkg on Windows, distro packages elsewhere).
+openssl = "0.10.73"
```
sfc-gh-pfus: Correct place for it. Test fixtures can keep generating keys with OpenSSL; production artifacts must not. Leaving it unvendored as a dev-dep is the right call — a cargo feature cannot point at a dev-dependency, which is why `vendored-openssl` had to go rather than be redirected.

ci/wif/run_in_container.sh
```diff
@@ -17,10 +17,12 @@ set -o pipefail
...
-# The runtime image is minimal; the vendored-openssl binary only needs glibc +
-# libstdc++, and rustls-native-certs needs the system CA bundle. Install the
-# few runtime bits defensively (no-op if already present).
-dnf install -y --setopt=install_weak_deps=False libstdc++ ca-certificates >/dev/null 2>&1 || true
+# The runtime image is minimal. The e2e binary needs glibc + libstdc++, plus
+# openssl-libs: sf_core keeps `openssl` as a dev-dependency for test fixtures
+# and links it dynamically, even though the driver itself no longer uses
+# OpenSSL. rustls-native-certs needs the system CA bundle. Install the few
+# runtime bits defensively (no-op if already present).
+dnf install -y --setopt=install_weak_deps=False libstdc++ ca-certificates openssl-libs >/dev/null 2>&1 || true
```
sfc-gh-boler: `openssl-libs` is now load-bearing: the e2e binary dynamically links libssl. `|| true` (and the stdout/stderr redirect) will hide a failed install, and the process then dies later with a missing-library error that looks like a WIF failure.

Drop `|| true` on this line, or fail closed if `ldd sf_core_e2e` still shows unresolved `libssl`/`libcrypto`. The previous "defensive no-op" was fine when the binary was fully vendored; it is not fine now.

python/hatch_native_arrow.py
```diff
@@ -124,7 +124,7 @@ def _cargo_overlay(target_dir: str) -> None:
             "--target-dir",
             target_dir,
             "--features",
-            "vendored-openssl,native-arrow",
+            "native-arrow",
```
sfc-gh-boler: Good catch on this path. `python/hatch_build.py` still copies every DLL from `$OPENSSL_DIR/bin` into the wheel on Windows, with a comment about `OPENSSL_STATIC=1` / co-located libssl. After F2 that cdylib should not need those DLLs. If CI still sets `OPENSSL_DIR`, we will ship unused OpenSSL DLLs next to `sf_core_python`.

Not in this diff — flagging here because it is the same class of leftover as `vendored-openssl` on this file. Strip the bundling, or gate it on something that still actually links OpenSSL.

tests/performance/drivers/core/app/Cargo.lock
```diff
@@ -4221,10 +4195,10 @@ dependencies = [
...
  "http-body-util",
  "infer",
- "jwt",
  "keyring",
  "libc",
  "lru 0.12.5",
+ "md-5 0.10.6",
  "memchr",
  "num-traits",
  "oauth2",
```
sfc-gh-boler: This nested lock still records `openssl` as a normal `sf_core` dependency (two lines below). The workspace and sdist locks were refreshed; this one was not.

Unlike the workspace `Cargo.lock`, this lock does not flatten test-only crates onto `sf_core` — no `criterion` / `ctor` / `wiremock`, but `openssl` is still here. That is the graph from when OpenSSL lived in `[dependencies]`. `cargo build --locked` in this tree will either fail the lock check or keep linking OpenSSL into the performance binary.

F2's own list of secondary lockfiles that pin `sf_core` includes this path. Regenerate it the same way as `python/Cargo.lock.sdist`.

.github/workflows/_build-odbc-packages.yml
```diff
@@ -200,7 +200,7 @@ jobs:
             $env:PATH = "C:\Strawberry\perl\bin;$env:PATH"
           }
           Write-Host "perl: $((Get-Command perl -ErrorAction SilentlyContinue).Source)"
-          cargo build --locked ${{ matrix.cargo_profile }} --package odbc --features vendored-openssl ${{ matrix.cargo_extra }} --target ${{ matrix.cargo_target }}
+          cargo build --locked ${{ matrix.cargo_profile }} --package odbc ${{ matrix.cargo_extra }} --target ${{ matrix.cargo_target }}
```
sfc-gh-boler: The feature is gone, but the comment two lines up still says this job needs PowerShell so `openssl-src` Configure can see Strawberry Perl. That is no longer true, and the same stale note is still in `_test-odbc.yml` (not in this diff).

Either delete the Perl PATH prepend on this job, or rewrite the comment to say it is leftover and why it stays. As written it will send the next person into a Configure/MSYS debug that no longer exists.

sf_core/Cargo.toml
zeroshade: Thanks -- no change needed here, recording the confirmation. The "a cargo feature cannot point at a dev-dependency" constraint is the whole reason `vendored-openssl` had to be retired rather than redirected, and that is now noted next to the dependency so it does not get re-proposed.

ci/wif/run_in_container.sh
zeroshade: Fixed in 94f555a8a. Dropped both the `|| true` and the output redirect, so dnf's own diagnostics reach the log and `-e` stops the script.

Also took the `ldd` check you offered as the alternative, because it verifies a different thing: a clean dnf exit does not prove *this* binary resolves. One subtlety worth flagging since it would have silently defeated it -- the output is captured into a variable before being matched rather than piped into `grep`, because inside an `if` condition `-e` is suspended, so a failing `ldd` would otherwise read as "nothing unresolved" and pass.

python/hatch_native_arrow.py
zeroshade: Fixed in 94f555a8a -- the bundling is gone.

Confirmed your read before removing it: `cargo tree -p python_bridge -e normal` lists no openssl crate, since `openssl` is now only an `sf_core` dev-dependency and a release cdylib does not build those. And the exposure was real, not theoretical -- both `_build-python-wheels.yml` and `build-python-release.yml` still run `setup-windows-openssl` on the ARM64 lanes, so `OPENSSL_DIR` is set and we would have shipped DLLs nothing loads.

I left the action in place: it is what lets the `cryptography` dependency build from source on that runner, which is unrelated to our linkage. The comment that replaces the copy says what would have to become true to bring bundling back, and that it should be gated on actual linkage rather than on `OPENSSL_DIR` being set.

tests/performance/drivers/core/app/Cargo.lock
zeroshade: Fixed in 94f555a8a. Regenerated via `cargo metadata` against that manifest, the same path `scripts/python-sdist-cargo-lock.sh` uses. `openssl` is gone from `sf_core`'s dependency list and the `openssl`/`openssl-sys` stanzas are gone entirely; `cargo check --locked` on the perf app now passes.

Your point about this lock not flattening test-only crates is what made it worth chasing rather than dismissing as lock noise -- no `criterion`/`wiremock`, so `openssl` sitting there really did mean a normal dependency edge.

It also generalises past F2, which I would have missed otherwise: F7 changes the AWS SDK feature set, so the same lock went stale again there. Fixed in that commit too (`fc694025f`), and all three locks are now verified `--locked` on each of the four branches in the stack rather than only at the tip.

.github/workflows/_build-odbc-packages.yml
zeroshade: Fixed in 94f555a8a -- removed the prepend rather than rewriting the comment, in all three places.

Checked first that nothing still needs Perl: `vendored-openssl` is retired and no step compiles OpenSSL from source any more, so the `openssl-src` Configure rationale has no remaining referent. Removed from `_build-odbc-packages.yml`, `_test-odbc.yml` (which you noted was outside the diff), and `_build-python-wheels.yml`, which had the same block in bash form.

The replacement comments state why no Perl is needed rather than staying silent, specifically so the next person does not reinstate it. `shell: pwsh` stays on the ODBC jobs for the explicit `$LASTEXITCODE` propagation, which the comments now say instead of the stale Perl reason.

sfc-gh-boler (commented):
LGTM

@sfc-gh-pfus what do you want to do next here?

zeroshade:
Rebased onto current `main` (`ca238468f`, 29 commits on from the previous base). One real conflict, in `.github/workflows/test-jdbc.yml` against `d73004bf6` "Speed up JDBC CI native builds" — flagging how I resolved it since it went against what my branch previously did.

That commit pinned the Windows lane to `key-suffix: debug-vendored-openssl` with "Stable suffix: warm main caches must still hit". My branch had renamed it to `debug-v2`. I resolved to **`debug-default`**, matching the other three lanes.

Reasoning, in case you disagree:

- After F2 this lane runs plain `cargo build --locked --package jdbc_bridge` — byte-identical invocation to the `debug-default` lanes. The suffix is documented as a "build-configuration discriminator", and the configuration is now the default one, so `debug-vendored-openssl` names a feature the lane no longer passes and `debug-v2` names nothing at all.
- Sharing the name with Linux/macOS cannot collide: `jdbc-bridge-cache` keys on `${{ runner.os }}-${{ runner.arch }}-…` before the suffix.
- It is not the gratuitous bump your env note warns against. That note says do not bump *for the debug-info change alone*, which is right because that does not change artifact validity. This does — and in fact the bump is not what protects us: the action already hashes `Cargo.lock`, `sf_core/Cargo.toml` and `jdbc_bridge/**`, all of which F2 touches, so warm entries invalidate on their own. No extra cold build either way.

I also dropped `debug-vendored-openssl` from the env comment at the top of the file, since no lane uses that key after this change.

Happy to switch to keeping `debug-vendored-openssl` verbatim if you would rather the diff not touch the key at all.

sfc-gh-boler (approved):

PR #1348, opened 2026-09-18
NO-SNOW: FIPS F7: drop the AWS SDK's ring-backed TLS stack

zeroshade:
> **Stacked on #1347** (which sits on #1346, which sits on #1345). The diff
> against `main` therefore carries those PRs' 15 commits — **review only the
> last commit here**, `NO-SNOW: FIPS F7: drop the AWS SDK's ring-backed TLS
> stack`. GitHub cannot use a cross-repo base branch, so this targets `main`
> until the PRs below it merge.

## What

Partially closes **F7**: ring is linked into the FIPS artifact. Measured on a
`--features fips-tls` lib-test binary built on this tree:

| | before | after |
| --- | --- | --- |
| ring symbols | 221 | **175** |
| rustls versions | 0.21 **and** 0.23 | **0.23 only** |
| hyper 0.14 | linked | **gone** |

aws-lc-fips symbols 1574, non-FIPS aws-lc still 0.

## The trap, for anyone auditing this later

The **`rustls` feature** on `aws-sdk-s3`/`aws-sdk-sts` is not the rustls the
rest of the driver uses. It expands to `aws-smithy-runtime/tls-rustls` →
`aws-smithy-http-client/legacy-rustls-ring`, linking hyper 0.14 and a *second*,
older rustls 0.21 built on ring.

`default-https-client` is the opposite — it maps to `rustls-aws-lc`, is
aws-lc-backed, and is **deliberately kept**, because it still supplies a
connector to paths that build an `SdkConfig` without an explicit client
(including one S3 test that does exactly that). The names invite precisely the
wrong conclusion.

Nothing ever used the legacy connector: every `aws_config::defaults(...)` call
site hands the SDK its own reqwest-backed `HttpClient` via
`tls::aws_http_client`. It was linked but dead.

`sigv4a` goes with the defaults — it pulls `aws-runtime/sigv4a` →
`aws-sigv4/sigv4a` → `dep:ring` and only signs S3 multi-region access point
ARNs. Snowflake stages are ordinary region-scoped buckets, so nothing here can
reach that signer.

## The residual 175 symbols are a decision, not an oversight

Please read the comment on the `aws-config` line before suggesting
`default-features = false` there. Ring's last source is `aws-config/sso`, which
hard-requires `dep:ring`. Dropping it **does** reach zero — verified, and
`aws-sdk-sso`/`ssooidc` fall out too — but it also removes AWS IAM Identity
Center credentials from the SDK's default credential chain, and two of the three
call sites resolve through that chain on purpose:

- `workload_identity/aws.rs` — credentials are `Option`; ambient discovery is
  the entire point of the WIF path
- `telemetry/platform_detection/aws.rs` — passes no `credentials_provider` at all

Only `file_manager/s3_transfer.rs` supplies explicit Snowflake-issued
credentials and would be unaffected.

So the trade is: ring in the FIPS artifact, or no AWS SSO credential support in
*any* artifact. Cargo features are additive and cannot be negated, so it cannot
be split per-build without the named-profile machinery the design doc rejects.
It also lands hardest on WORKLOAD_IDENTITY, the login path most likely to be
used in the deployments that want FIPS in the first place.

**Decision taken: keep SSO, record ring as a known deviation.**

## Also not fixable from this repo

`aws-sigv4` computes plain SigV4's HMAC-SHA256 through RustCrypto
**unconditionally** (`aws-sigv4/src/sign/v4.rs`), so the scope is every S3 stage
transfer, not just WIF. That needs an upstream ask or a written deviation.

## Verification

- 2199 tests pass by default, 2200 with `--features fips-tls`
- fmt clean; clippy unchanged from baseline; all four bindings `cargo check`
- root `Cargo.lock` and `python/Cargo.lock.sdist` both refreshed and verified
  under `--locked`


sf_core/Cargo.toml
```diff
@@ -74,8 +111,47 @@ prost = { version = "0.14.1", optional = true }
...
+# `default-https-client` is deliberately kept. It maps to
+# `aws-smithy-http-client/rustls-aws-lc`, so it is aws-lc-backed rather than
+# ring-backed, and it still supplies a default connector to the paths that build
+# an `SdkConfig` without an explicit client.
+#
+# `sigv4a` is dropped with the defaults: it pulls `aws-runtime/sigv4a` ->
+# `aws-sigv4/sigv4a` -> `dep:ring`, and it only signs S3 multi-region access
+# point ARNs. Snowflake stages are ordinary region-scoped buckets, so nothing
+# here can reach that signer.
+aws-sdk-s3 = { version = "1.106.0", default-features = false, features = ["rt-tokio", "default-https-client"] }
```
sfc-gh-pfus: `default-https-client` reintroduces a second TLS stack (smithy’s aws-lc connector) for any `SdkConfig` built without `tls::aws_http_client`. That path ignores the driver’s `TlsConfig` — CRL, custom roots, proxy, and the FIPS gate.

Main already compiles the SDK’s default HTTP client out, so a missing transport is a compile error rather than a silent second stack. On rebase, please keep that rule and do not bring `default-https-client` back. Being aws-lc-backed is not enough; it still is not *this* TLS policy.

SSO/ring residual and RustCrypto SigV4 can stay as recorded deviations for a later PR — this comment is only about the default connector.

sf_core/Cargo.toml
```diff
@@ -105,9 +111,43 @@ prost = { version = "0.14.1", optional = true }
...
+#
+# Decision: keep SSO support and record ring as a known deviation. Cargo
+# features are additive and cannot be negated, so this cannot be split
+# per-build without the named-profile machinery the design doc rejects --
+# `fips-tls` cannot turn a default feature off. Please do not "fix" this by
+# dropping `sso`; that is the trade, not an oversight.
+aws-sdk-s3 = { version = "1.106.0", default-features = false, features = ["rt-tokio"] }
 aws-sdk-sts = { version = "1.96.0", default-features = false, features = ["rt-tokio"] }
-aws-config = { version = "1.8.12", default-features = false, features = ["rt-tokio", "credentials-process"] }
+aws-config = { version = "1.8.12", default-features = false, features = ["rt-tokio", "credentials-process", "sso"] }
```
sfc-gh-boler: Drop `sso`. Do not put `ring` back into the FIPS artifact for IAM Identity Center.

WIF only needs the default chain for IMDS / IRSA / env keys / instance role. Stage transfers never use this chain. Platform detection is opt-in telemetry. UD Gherkins (`workload_identity.feature`) cover GetCallerIdentity and GetWebIdentityToken against a cloud identity, not `aws sso login`.

JDBC and .NET do not ship an SSO/Identity Center module. Python/Go/Node only get it as a boto/SDK default-chain side effect, and even Python's copy says "IAM role attached."

#1345 already had `default-features = false` without `sso`. Restoring it is a compatibility extra, not a WIF contract, and it is the remaining ring source this PR is supposed to be shrinking. Omit the feature; record Identity Center as unsupported on this core (same as JDBC).

sf_core/Cargo.toml
```diff
@@ -17,16 +17,24 @@ sonic-json = ["sonic-rs"]
...
-# once the non-TLS crypto is ported off OpenSSL.
+# Still named `fips-tls` rather than `fips`, but no longer for the original
+# reason. JWT signing, stage file encryption, DPoP and private-key parsing have
+# all been ported off OpenSSL to AWS-LC, so the feature's reach is no longer
+# TLS-only in practice.
+#
+# The name stays because the *claim* is not yet complete:
+#   - SigV4 request signing computes HMAC-SHA256 outside the module:
+#     `aws-sigv4` uses RustCrypto unconditionally, and `aws-sdk-s3` enables
+#     `sigv4a` by default, which links ring. That is every stage transfer, not
```
sfc-gh-boler: This commit turns `sigv4a` off. The `fips-tls` comment still says `aws-sdk-s3` enables `sigv4a` and that is why ring is linked. After dropping `sso`, ring from this graph should be gone; the F7 leftover is SigV4 HMAC via RustCrypto on every stage transfer. Update that bullet.

PR #1349, opened 2026-09-18
NO-SNOW: FIPS F8: draw OAuth PKCE and CSRF randomness from the module

zeroshade:
> **Stacked on #1348** (→ #1347 → #1346 → #1345). The diff against `main`
> carries those PRs' 16 commits — **review only the last commit here**,
> `NO-SNOW: FIPS F8: draw OAuth PKCE and CSRF randomness from the module`.
> GitHub cannot use a cross-repo base branch, so this targets `main` until the
> PRs below it merge.

## What

Closes **F8**. The PKCE code verifier and the CSRF `state` are the two values in
the authorization-code flow whose security rests on being unpredictable, and
both came from `oauth2`'s generators, which draw on `rand`'s thread RNG. That is
a fine default for a general-purpose client and the wrong source here: under
`fips-tls` the claim is that security-relevant randomness comes from the
validated module.

Both now come from AWS-LC's DRBG through a new `oauth::random::token_b64url`,
which reproduces the shape `oauth2` produced — N random bytes, base64url without
padding — so only the *source* changes. Values on the wire stay
indistinguishable, which is what keeps RFC 7636 compliance and IdP
interoperability from depending on this swap.

A DRBG failure fails the login via a new `OAuthError::RandomGeneration` rather
than falling back to another source, which would void the claim exactly when it
mattered.

## The part worth reviewing carefully: `set_pkce_challenge` is bypassed

The challenge digest moves to AWS-LC too. `oauth2` will only build a
`PkceCodeChallenge` around a digest it computed itself with RustCrypto
(`types.rs:486`), so the flow now appends `code_challenge` and
`code_challenge_method` itself — the same two parameters, with the same values,
that `set_pkce_challenge` emits (`code.rs:163-165`).

That bypass is the one part of this that could break silently, so the existing
end-to-end authorize-URL test now asserts the binding it never checked: that
`code_challenge` is the S256 digest of the `code_verifier` presented at the
token endpoint. Its previous assertions only checked the parameter was
non-empty, and would still have passed if the two were unrelated, or if the
digest were taken over the raw DRBG bytes rather than the encoded verifier —
both of which look right and fail at the IdP.

**I confirmed the new assertion is not vacuous** by deliberately hashing the
wrong input and watching it fail.

## `pkce.rs` stops being dead scaffolding

It was `#[allow(dead_code)]`, duplicating the inline implementation. Rather than
fix the RNG twice, or leave a `rand`-based generator sitting next to a
DRBG-based one for someone to wire up later, it becomes the single
implementation and the allow comes off.

## What `rand` is still used for

It stays in the dependency graph, for retry and refresh jitter (`http/retry.rs`,
`crl/cache.rs`, `token_cache/file_cache.rs`) and a non-secret FFI handle tag
(`handle_manager.rs`). None of those are security-relevant randomness. Recording
that classification at the call sites, so a dependency audit does not read them
as a bypass, belongs with the cargo-deny/SBOM work in Phase 5.

## Verification

- 2202 tests pass by default, 2203 with `--features fips-tls` (the three new
  `random` tests)
- fmt clean; clippy unchanged from baseline; all four bindings `cargo check`


sf_core/src/rest/snowflake/oauth/random.rs
```diff
@@ -0,0 +1,72 @@
...
+/// `purpose` names the value for the error message and is a `&'static str`
+/// rather than the value itself: these are secrets (the PKCE verifier) or
+/// anti-forgery tokens (the CSRF state) and must not reach a log.
+///
+/// The raw buffer is zeroized before returning. The encoded string is the only
+/// copy that leaves, and its owner decides how to protect it -- the PKCE
+/// verifier goes straight into a `SensitiveString`.
+pub(crate) fn token_b64url(num_bytes: usize, purpose: &'static str) -> Result<String, OAuthError> {
+    let mut bytes = vec![0u8; num_bytes];
+    fill(&mut bytes).context(RandomGenerationSnafu { purpose })?;
```
sfc-gh-pfus: Right source and right failure mode — no fallback to `rand` if the DRBG fails.

One leftover of the same finding, not for this PR: DPoP `jti` in `oauth/dpop.rs` is still `Uuid::new_v4()`. Keygen and the proof signature are already on AWS-LC; the replay nonce is not. Fine as a follow-up.

sf_core/Cargo.toml
```diff
@@ -17,16 +17,24 @@ sonic-json = ["sonic-rs"]
...
 # happening to unify rustls's transitive activation into the same crate.
 #
-# Named `fips-tls` rather than `fips` because it covers TLS only. JWT signing,
-# stage file encryption, DPoP and key parsing still run on OpenSSL, which has no
-# FIPS-validated provider -- and under `vendored-openssl` that unvalidated copy
-# is statically linked where a customer cannot replace it. An artifact built
-# from this feature is therefore not a FIPS artifact, and the narrower name is
-# what keeps the two apart: `fips` is left unclaimed for the complete thing,
-# once the non-TLS crypto is ported off OpenSSL.
+# Still named `fips-tls` rather than `fips`, but no longer for the original
```
sfc-gh-boler: Rewrite this as the current inventory, not a changelog.

“No longer for the original reason” only makes sense against the deleted OpenSSL sentences. After merge nobody has that diff. Drop the F8 bullet (this commit closed it) and the `sigv4a` clause (gone on #1348).

Something like: this feature selects `aws-lc-fips-sys` for rustls and for the application crypto that already runs on AWS-LC (JWT, private keys, stage encryption, DPoP, OAuth PKCE/CSRF). It stays `fips-tls` because SigV4 HMAC-SHA256 still runs on RustCrypto (`aws-sigv4`) on every stage transfer (F7). Rename to `fips` once that — and any other remaining off-module security crypto — is gone.

There is no follow-up PR in this stack that finishes the claim. HMAC is AWS-SDK work, not another commit here.

sf_core/src/rest/snowflake/oauth/dpop.rs
```diff
@@ -193,23 +248,15 @@ pub(crate) fn proof_jwt(
...
-    let sig = EcdsaSig::sign(&digest, &key.key).context(DPoPProofGenerationSnafu)?;
-    let r_bytes = sig
-        .r()
-        .to_vec_padded(COORD_BYTES)
+    // `sign` hashes the message itself. Deliberately not `sign_digest` with a
+    // pre-computed hash: AWS-LC marks that path as not FIPS-approved, so a
+    // `fips-tls` build must hand over the message and let the module hash it.
+    let sig = key
+        .key
+        .sign(&SystemRandom::new(), signing_input.as_bytes())
```
sfc-gh-boler: Same F8 argument as PKCE/CSRF: this is OAuth-flow security randomness, and it still is not the module.

A few lines above, `jti` is minted with `Uuid::new_v4()` (`getrandom`). It is the replay id on every DPoP proof. Signing already uses `SystemRandom`; the nonce next to it does not.

Either draw 16 bytes with the new helper (any unique string is a legal `jti`) and fail the proof the same way as `RandomGeneration`, or put this call on the leftover list with jitter. Do not treat F8 as emptied OAuth of off-module RNG while proofs still mint `jti` off-module.

